Issue #14, opened 2020-12-11
Support aws_vpc


eliecharra:
**Description**
Add support for aws_vpc

PR #25, opened 2020-12-15
Add VPC support

eliecharra:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | yes
| 🚀 New feature?   | yes
| ⚠ Deprecations?   | no
| ❌ BC Break       | no
| 🔗 Related issues | #15 #14 
| ❓ Documentation  | yes

## Description

Add support for
- `aws_default_vpc`
- `aws_vpc`

⚠️ This PR will also add support for returning mixed resources types from a single supplier


pkg/remote/aws/default_vpc_supplier.go
```diff
@@ -0,0 +1,76 @@
...
+	}
+}
+
+func (s DefaultVPCSupplier) Resources() ([]resource.Resource, error) {
+	input := ec2.DescribeVpcsInput{}
+	var defaultVPCs []*ec2.Vpc
+	err := s.client.DescribeVpcsPages(&input,
+		func(resp *ec2.DescribeVpcsOutput, lastPage bool) bool {
+			for _, vpc := range resp.Vpcs {
+				if vpc.IsDefault != nil && *vpc.IsDefault {
```
eliecharra, line 40: @moadibfr WDYT of handling both `aws_default_vpc` and `aws_vpc` in a single supplier ? May I rename this supplier to `aws_vpc_supplier.go` and handle both default and non default VPCs ?

It will avoid nearly duplicated logic.

moadibfr: I think it's a good idea to avoir duplication as this is almost identical. And yeah `aws_vpc_supplier` might be clearer. We might want to add a little comment on the struct to warn that it supply the two resources also.

eliecharra: 👍🏻

pkg/remote/aws/vpc_supplier_test.go
```diff
@@ -0,0 +1,118 @@
...
+								},
+								{
+									VpcId: aws.String("vpc-0768e1fd0029e3fc3"),
+								},
+								{
+									VpcId:     aws.String("vpc-020b072316a95b97f"),
+									IsDefault: aws.Bool(false),
+								},
+							},
+						}, true)
```
wbeuil, line 67: Shouldn't it be `false` here ?

wbeuil: Even though this won't change the test itself and its results, we should make it so that it's logic for reading PR purposes.

pkg/resource/aws/deserializer/vpc_deserializer.go
```diff
@@ -0,0 +1,43 @@
...
+
+func NewVPCDeserializer() *VPCDeserializer {
+	return &VPCDeserializer{}
+}
+
+func (s *VPCDeserializer) HandledType() resource.ResourceType {
+	return resourceaws.AwsVpcResourceType
+}
+
+func (s VPCDeserializer) Deserialize(recordList []cty.Value) ([]resource.Resource, error) {
```
wbeuil, line 24: You kept the recordList from Route53 records, can you change it to something more inline with vpc, like `vpcList` and update all the lines below.

pkg/resource/aws/deserializer/default_vpc_deserializer.go
```diff
@@ -0,0 +1,43 @@
...
+
+func NewDefaultVPCDeserializer() *DefaultVPCDeserializer {
+	return &DefaultVPCDeserializer{}
+}
+
+func (s *DefaultVPCDeserializer) HandledType() resource.ResourceType {
+	return resourceaws.AwsDefaultVpcResourceType
+}
+
+func (s DefaultVPCDeserializer) Deserialize(recordList []cty.Value) ([]resource.Resource, error) {
```
wbeuil, line 24: Same here for `recordList`

wbeuil (changes requested):

pkg/remote/aws/vpc_supplier_test.go
eliecharra: Sure, I've inverted this bool meaning in my mind 👍🏻 nice catch

pkg/resource/aws/deserializer/vpc_deserializer.go
eliecharra: Nice catch, let's update it for something more generic like `rawList` as we use `rawResource` as iterator name in the loop.

wbeuil (dismissed):
LGTM

moadibfr (dismissed):

Issue #7, opened 2020-12-11
Create a thread safe mecanism to handle alerts during execution


eliecharra:
**Description**
Create a running routine with an exposed channel to allow us to send alerts from any moment of our lifecycle.

Use cases:
- We want to send alert during middleware runs (when exploding security group rules)
- We want to send alert during analysis for computed fields.

We want to be able to query alert collection when outputing results:
- Output inline for console output
- Output in a `alerts` key in json output

**Example**
Alerts should be in analysis struct.

```go
type Alerts map[string][]Alert
type Alert struct {
   Message string
   ShouldIgnoreResource bool
}

// Alert keys are indexed on resources fqdn
Alerts["aws_s3_bucket.my-bucket"]
// For a s3 403 read permission denied 
Alerts["aws_s3_bucket"]

// JSON Output
{
    "alerts": {
        "aws_s3_bucket.my-bucket": [
            {
                "message": "Unable to read bucket"
            }
        ],
        "aws_s3_bucket": [
            {
                "message": "Permissions denied while reading bucket list",
                "should_ignore_resource": true
            },
            {
                "message": "Another issue"
            }
        ]
    }
}
```

PR #29, opened 2020-12-16
Handle alert during execution

wbeuil:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | no
| 🚀 New feature?   | yes
| ⚠ Deprecations?   | no
| ❌ BC Break       | no
| 🔗 Related issues | #7
| ❓ Documentation  | yes

## Description

Simple way to start a goroutine and to send alerts to it so as to display useful information to users.

An example could be if there are computed attributes of a resource that drifted but wasn't refreshed inside the tfstates, we should inform the user that he/she needs to refresh its states since it could be a false positive drift.

image

### DIFF
--- a/pkg/alerter/alerter.go
+++ b/pkg/alerter/alerter.go
@@ -1,0 +1,71 @@
+package alerter
+
+import (
+	"fmt"
+
+	"github.com/cloudskiff/driftctl/pkg/resource"
+)
+
+type Alerter struct {
+	alerts   Alerts
+	alertsCh chan Alerts
+	doneCh   chan bool
+}
+
+func NewAlerter() *Alerter {
+	var alerter = &Alerter{
+		alerts:   make(Alerts),
+		alertsCh: make(chan Alerts),
+		doneCh:   make(chan bool),
+	}
+
+	go alerter.run()
+
+	return alerter
+}
+
+func (a *Alerter) run() {
+	defer func() { a.doneCh <- true }()
+	for alert := range a.alertsCh {
+		for k, v := range alert {
+			if val, ok := a.alerts[k]; ok {
+				a.alerts[k] = append(val, v...)
+			} else {
+				a.alerts[k] = v
+			}
+		}
+	}
+}
+
+func (a *Alerter) SetAlerts(alerts Alerts) {
+	a.alerts = alerts
+}
+
+func (a *Alerter) GetAlerts() Alerts {
+	close(a.alertsCh)
+	<-a.doneCh
+	return a.alerts
+}
+
+func (a *Alerter) SendAlert(key string, alert Alert) {
+	a.alertsCh <- Alerts{
+		key: []Alert{alert},
+	}
+}
+
+func (a *Alerter) IsResourceIgnored(res resource.Resource) bool {
+	alert, alertExists := a.alerts[fmt.Sprintf("%s.%s", res.TerraformType(), res.TerraformId())]
+	wildcardAlert, wildcardAlertExists := a.alerts[res.TerraformType()]
+	shouldIgnoreAlert := a.shouldBeIgnored(alert)
+	shouldIgnoreWildcardAlert := a.shouldBeIgnored(wildcardAlert)
+	return (alertExists && shouldIgnoreAlert) || (wildcardAlertExists && shouldIgnoreWildcardAlert)
+}
+
+func (a *Alerter) shouldBeIgnored(alert []Alert) bool {
+	for _, a := range alert {
+		if a.ShouldIgnoreResource {
+			return true
+		}
+	}
+	return false
+}
--- a/pkg/alerter/alerter.go
+++ b/pkg/alerter/alerter.go
@@ -41,7 +41,7 @@ func (a *Alerter) SetAlerts(alerts Alerts) {
 	a.alerts = alerts
 }
 
-func (a *Alerter) GetAlerts() Alerts {
+func (a *Alerter) Retrieve() Alerts {
 	close(a.alertsCh)
 	<-a.doneCh
 	return a.alerts
--- a/pkg/analyser/analysis.go
+++ b/pkg/analyser/analysis.go
@@ -8,9 +8,16 @@ import (
 	"github.com/r3labs/diff/v2"
 )
 
+type Change struct {
+	diff.Change
+	Computed bool `json:"computed"`
+}
+
+type Changelog []Change
+
 type Difference struct {
 	Res       resource.Resource
-	Changelog diff.Changelog
+	Changelog Changelog
 }
 
 type Summary struct {
@@ -32,7 +39,7 @@ type Analysis struct {
 
 type serializableDifference struct {
 	Res       resource.SerializableResource `json:"res"`
-	Changelog diff.Changelog                `json:"changelog"`
+	Changelog Changelog                     `json:"changelog"`
 }
 
 type serializableAnalysis struct {
--- a/pkg/analyser/analysis.go
+++ b/pkg/analyser/analysis.go
@@ -3,6 +3,7 @@ package analyser
 import (
 	"encoding/json"
 
+	"github.com/cloudskiff/driftctl/pkg/alerter"
 	"github.com/cloudskiff/driftctl/pkg/resource"
 	"github.com/r3labs/diff/v2"
 )
@@ -26,6 +27,7 @@ type Analysis struct {
 	deleted     []resource.Resource
 	differences []Difference
 	summary     Summary
+	alerts      alerter.Alerts
 }
 
 type serializableDifference struct {
@@ -40,6 +42,7 @@ type serializableAnalysis struct {
 	Deleted     []resource.SerializableResource `json:"deleted"`
 	Differences []serializableDifference        `json:"differences"`
 	Coverage    int                             `json:"coverage"`
+	Alerts      alerter.Alerts                  `json:"alerts"`
 }
 
 func (a Analysis) MarshalJSON() ([]byte, error) {
@@ -61,6 +64,7 @@ func (a Analysis) MarshalJSON() ([]byte, error) {
 	}
 	bla.Summary = a.summary
 	bla.Coverage = a.Coverage()
+	bla.Alerts = a.alerts
 
 	return json.Marshal(bla)
 }
@@ -97,6 +101,7 @@ func (a *Analysis) UnmarshalJSON(bytes []byte) error {
 			Changelog: di.Changelog,
 		})
 	}
+	a.AddAlerts(bla.Alerts)
 	return nil
 }
 
@@ -127,6 +132,10 @@ func (a *Analysis) AddDifference(diffs ...Difference) {
 	a.summary.TotalDrifted += len(diffs)
 }
 
+func (a *Analysis) AddAlerts(alerts alerter.Alerts) {
+	a.alerts = alerts
+}
+
 func (a *Analysis) Coverage() int {
 	if a.summary.TotalResources > 0 {
 		return int((float32(a.summary.TotalManaged) / float32(a.summary.TotalResources)) * 100.0)
@@ -152,4 +161,8 @@ func (a *Analysis) Differences() []Difference {
 
 func (a *Analysis) Summary() Summary {
 	return a.summary
+}
+
+func (a *Analysis) Alerts() alerter.Alerts {
+	return a.alerts
 }
--- a/pkg/analyser/analysis.go
+++ b/pkg/analyser/analysis.go
@@ -108,7 +108,7 @@ func (a *Analysis) UnmarshalJSON(bytes []byte) error {
 			Changelog: di.Changelog,
 		})
 	}
-	a.AddAlerts(bla.Alerts)
+	a.SetAlerts(bla.Alerts)
 	return nil
 }
 
@@ -139,7 +139,7 @@ func (a *Analysis) AddDifference(diffs ...Difference) {
 	a.summary.TotalDrifted += len(diffs)
 }
 
-func (a *Analysis) AddAlerts(alerts alerter.Alerts) {
+func (a *Analysis) SetAlerts(alerts alerter.Alerts) {
 	a.alerts = alerts
 }
 
--- a/pkg/analyser/analyzer.go
+++ b/pkg/analyser/analyzer.go
@@ -1,30 +1,36 @@
 package analyser
 
 import (
+	"fmt"
+	"reflect"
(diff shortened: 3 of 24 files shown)


pkg/alerter/alerter.go
```diff
@@ -0,0 +1,79 @@
...
+		defer close(a.AlertsCh)
+		for {
+			alert := <-a.AlertsCh
+			a.mergeAlerts(alert)
+		}
+	}()
+}
+
+func (a *Alerter) Alert(key string, alert Alert) {
+	a.AlertsCh <- Alerts{
```
eliecharra, line 35: Why not doing a `AlertsCh chan Alert` instead of `AlertsCh chan Alerts`

pkg/alerter/alerter.go
```diff
@@ -0,0 +1,79 @@
...
+func NewAlerter() *Alerter {
+	return &Alerter{
+		Alerts:   make(Alerts),
+		AlertsCh: make(chan Alerts),
+	}
+}
+
+func (a *Alerter) Start() {
+	go func() {
+		defer close(a.AlertsCh)
```
eliecharra, line 26: I think you should do something like this instead

```go
    for alert := range a.AlertsCh {
        for k, v := range alert {
		if val, ok := a.Alerts[k]; ok {
			a.Alerts[k] = append(val, v...)
		} else {
			a.Alerts[k] = v
		}
	}
    }
```

Since it is the only place we mutate the alert list we don't need a mutex.
When you'll close the AlertCh, the loop will break.

pkg/alerter/alerter.go
```diff
@@ -0,0 +1,79 @@
...
+				"type": res.TerraformType(),
+				"id":   res.TerraformId(),
+			}).Debug("Found resource to ignore because of an alert")
+		}
+	}
+
+	return results
+}
+
+func (a *Alerter) shouldBeIgnored(alert []Alert) bool {
```
eliecharra, line 72: It could be a method of `Alerts` struct directly

```go
func (a *Alerts) shouldBeIgnored(resource resource.Resource) bool {
  for _, v := range a {
      // Check if the resource match and if it should be ignored
  }
} 
```

pkg/analyser/analyzer.go
```diff
@@ -34,10 +40,22 @@ func (a Analyzer) Analyze(remoteResources []resource.Resource, resourcesFromStat
 				Res:       stateRes,
 				Changelog: delta,
 			})
+			for _, d := range delta {
```
eliecharra, line 43: Maybe we could move this in a separate unexported method to avoid a too complex `Analyze` method

pkg/alerter/alerter.go
```diff
@@ -0,0 +1,79 @@
...
+	go func() {
+		defer close(a.AlertsCh)
+		for {
+			alert := <-a.AlertsCh
+			a.mergeAlerts(alert)
+		}
+	}()
+}
+
+func (a *Alerter) Alert(key string, alert Alert) {
```
eliecharra, line 34: In term of naming, I found `Add` or `NewAlert` / `AddAlert` more clear (the second option is maybe a little redundant `alerter.AddAlert`)

pkg/analyser/analyzer.go
```diff
@@ -49,3 +67,18 @@ func findCorrespondingRes(resources []resource.Resource, res resource.Resource)
 	}
 	return -1, nil, false
 }
+
+func getField(t reflect.Type, path []string) (reflect.StructField, bool) {
```
eliecharra, line 71: Scope it to analyzer struct 

`func (a *Analyzer) getField(t reflect.Type, path []string) (reflect.StructField, bool)`

It does not makes sense to expose this method in the whole package

pkg/alerter/alerter.go
```diff
@@ -0,0 +1,79 @@
...
+}
+
+func NewAlerter() *Alerter {
+	return &Alerter{
+		Alerts:   make(Alerts),
+		AlertsCh: make(chan Alerts),
+	}
+}
+
+func (a *Alerter) Start() {
```
eliecharra, line 24: Do we really need a `Start` method ? Could we move this in the `NewAlerter` ?

eliecharra (changes requested):
Just a partial review on code architecture, I will review details in a next pass

pkg/alerter/alerter.go
```diff
@@ -0,0 +1,79 @@
...
+	for k, v := range alert {
+		if val, ok := a.Alerts[k]; ok {
+			a.Alerts[k] = append(val, v...)
+		} else {
+			a.Alerts[k] = v
+		}
+	}
+}
+
+func (a *Alerter) IgnoreResources(resources []resource.Resource) []resource.Resource {
```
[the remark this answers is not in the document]
wbeuil: I remembered our last talk where we described the "ignoring part" being somewhat similar to the one we did for the driftignore. Since we did it right before the analysis starts, I put the IgnoreResources right before the analysis too in the `driftctl.go` file.

Would love @moadibfr's feedback on this one.

pkg/alerter/alerter.go
wbeuil: I did an `AlertsCh chan Alerts` so that I don't need to send into one function channel or two differents channels and the key and the alert (this is just a message and a bool) to create the map of alerts.

pkg/alerter/alerter.go
wbeuil: Do you mean that I can get rid of the Mutex composition struct and to this instead ?

wbeuil: I just tried it on indeed it works. I was afraid that with the goroutines it would have messed up the alerts map.

pkg/alerter/alerter.go
wbeuil: We could do a method of Alerts but I don't get it why we may need a loop here.

Since it's a map of string []Alert, we could simply check if both the single `key.id` and the wildcard `key` exist, and retrieve if either one those two has an Alert inside it that has the bool ShouldIgnoreResource.

cf. what i did on the `IgnoreResources` or the `Run` of the drigtignore.

pkg/analyser/analyzer.go
wbeuil: I agree 👌🏻

pkg/alerter/alerter.go
wbeuil: I find `Add` very minimalist and don't incorporate the idea of sending alert. What about `SendAlert` or `Send` ?

pkg/alerter/alerter.go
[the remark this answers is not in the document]
moadibfr: I agree with Elie on the name and on the place.
In my PR on ignore I had the same problem added argument to Analyzer to pass a Filterer. I'm not sure this is ideal either

pkg/alerter/alerter.go
moadibfr: :+1:

pkg/alerter/alerter.go
moadibfr: I like SendAlert too

pkg/alerter/alerter.go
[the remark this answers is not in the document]
eliecharra: Lets close this one and continue discussion there : https://github.com/cloudskiff/driftctl/pull/28/files#discussion_r545750961

pkg/alerter/alerter.go
eliecharra: The goroutine is only an async task which read from the chan and append to a list, this is thread safe unless you launch multiples instances of this routine to mutate the map.

This can be viewed as a single thead queue consumer inside your struct. So no concurrency issues there

pkg/alerter/alerter.go
eliecharra: 👍🏻 for SendAlert

pkg/alerter/alerter.go
wbeuil: Let's remove it. I assumed it would have confused you to have both the initialization and the goroutine inside it. I was wrong :)

wbeuil: Removed 7f5211c

pkg/analyser/analyzer.go
wbeuil: Done 7f5211c

pkg/alerter/alerter.go
wbeuil: Done 7f5211c

pkg/analyser/analyzer.go
wbeuil: Done 7f5211c

pkg/alerter/alerter.go
wbeuil: Done 7f5211c

pkg/alerter/alerter.go
```diff
@@ -0,0 +1,71 @@
...
+type Alerter struct {
+	Alerts   Alerts
+	AlertsCh chan Alerts
+}
+
+func NewAlerter() *Alerter {
+	alerts := make(Alerts)
+	alertsCh := make(chan Alerts)
+
+	go func() {
```
moadibfr, line 19: for readability and isolation I would prefer to see the goroutine as a Alerter method i think

pkg/alerter/alerter.go
```diff
@@ -0,0 +1,71 @@
...
+	}
+}
+
+func (a *Alerter) SendAlert(key string, alert Alert) {
+	a.AlertsCh <- Alerts{
+		key: []Alert{alert},
+	}
+}
+
+func (a *Alerter) IgnoreResources(resources []resource.Resource) []resource.Resource {
```
moadibfr, line 44: thought we decided to move this logic in analyzer ?

moadibfr (changes requested):

pkg/alerter/alerter.go
wbeuil: Indeed, done in 442d3ed

pkg/alerter/alerter.go
wbeuil: Done 442d3ed

pkg/alerter/alerter.go
eliecharra: What about something like this ?

image

moadibfr (changes requested):

pkg/alerter/alerter.go
```diff
@@ -0,0 +1,71 @@
...
+			if val, ok := a.alerts[k]; ok {
+				a.alerts[k] = append(val, v...)
+			} else {
+				a.alerts[k] = v
+			}
+		}
+	}
+}
+
+func (a *Alerter) SetAlerts(alerts Alerts) {
```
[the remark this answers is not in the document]
wbeuil: It's for test purposes. I wanted a way to add alerts to the alerter without sending them through the channel. Maybe we can do it in an easier way.

pkg/alerter/alerter.go
```diff
@@ -0,0 +1,71 @@
...
+	a.alerts = alerts
+}
+
+func (a *Alerter) GetAlerts() Alerts {
```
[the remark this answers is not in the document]
wbeuil: Indeed we can easily change the name. Something like `ReadAlerts`, or `FetchAlerts`, or `DisplayAlerts`, what do you think? If you have other ideas, let's write them here.

pkg/analyser/analysis.go
```diff
@@ -127,6 +139,10 @@ func (a *Analysis) AddDifference(diffs ...Difference) {
 	a.summary.TotalDrifted += len(diffs)
 }
 
+func (a *Analysis) AddAlerts(alerts alerter.Alerts) {
```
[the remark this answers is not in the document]
wbeuil: I put it like that to be consistent with what we have in the other setters for the`Analysis`: `AddDeleted`, `AddUnmanaged`, `AddManaged`, `AddDifference`.

pkg/alerter/alerter.go
wbeuil: It would mean that if by any chance we have this kind of setup for alerts, it won't be ignored:

```
"alerts": {
  "aws_resource": [
    { "Message": "something", "ShouldIgnoreResource": true },
  ],
  "aws_resource.foo": [
    { "Message": "updated.field is a computed field" },
  ]
}
```

eliecharra (changes requested):

pkg/alerter/alerter.go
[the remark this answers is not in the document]
eliecharra: I'd rather prefer some wording without "alerts", something like `Retrieve`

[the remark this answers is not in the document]
moadibfr: `Retrieve` or `Fetch` would be nice I think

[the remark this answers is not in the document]
wbeuil: Ok let's go for simply `Retrieve`

pkg/analyser/analysis.go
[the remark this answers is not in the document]
moadibfr: yeah but the others append to their list, it's not what you did. Either make it a setter like `SetAlerts` or append to alerts

pkg/alerter/alerter.go
eliecharra: Right nice catch, we need to invert the two alerts assignations to ensure correct precedence 

image

pkg/analyser/analyzer.go
```diff
@@ -51,25 +57,32 @@ func (a Analyzer) Analyze(remoteResources []resource.Resource, resourcesFromStat
...
 			})
-			changelog := make([]diff.Change, 0, len(delta))
+			changelog := make([]Change, 0, len(delta))
 			for _, change := range delta {
 				if filter.IsFieldIgnored(stateRes, change.Path) {
 					continue
 				}
-				changelog = append(changelog, change)
+				c := Change{Change: change}
+				if a.isComputedField(stateRes, c) {
```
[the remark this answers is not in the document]
wbeuil: Done 011b380

pkg/analyser/analysis.go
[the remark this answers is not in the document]
wbeuil: `SetAlerts` on commit 011b380

pkg/alerter/alerter.go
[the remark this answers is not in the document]
wbeuil: Done 011b380

eliecharra (approved):

moadibfr (approved):

PR #51, opened 2020-12-22
add fmt check in ci

moadibfr:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | no
| 🚀 New feature?   | no
| ⚠ Deprecations?   | no
| ❌ BC Break       | no
| 🔗 Related issues |
| ❓ Documentation  | no

## Description

We had badly formated (not passed to fmt) file in main. `make` will create diff that add noise to PR.
I added a check on ci during tests phase to prevent PR to be merge when fmt is not ok.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -17,6 +17,17 @@ jobs:
       - image: golang:1.15
     steps:
       - checkout
+      - run:
+          name: Enforce Go Formatted Code
+          command: |
+            go fmt ./...
+            if [[ -z $(git status --porcelain) ]]; then
+              echo "Git directory is clean."
+            else
+              echo "Git directory is dirty. Run make fmt locally and commit any formatting fixes or generated code."
+              git status --porcelain
+              exit 1
+            fi
       - run: make install-tools
       - run: make test
       - run:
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -17,6 +17,9 @@ jobs:
       - image: golang:1.15
     steps:
       - checkout
+      - run:
+          name: Enforce Go Formatted Code
+          command: test -z "$(go fmt ./...)"
```
wbeuil, line 22: Didn't know that you could use `test -z` to return false or true and that CircleCI would exit. Thanks :)

What I envisioned is more verbose. We could display what are the files we need to re-format just like terraform does.

```yaml
- run:
      name: Enforce Go Formatted Code
      command: go fmt ./...
- run:
      name: Verify No Code Was Generated
      command: |
            if [[ -z $(git status --porcelain) ]]; then
              echo "Git directory is clean."
            else
              echo "Git directory is dirty. Run `make fmt` locally and commit any formatting fixes or generated code."
              git status --porcelain
              exit 1
            fi
```
This is what they do in their ci/cd pipeline. But we can use `circleci-agent step halt` instead of `exit 1` as per CircleCI support [documentation](https://support.circleci.com/hc/en-us/articles/360015562253-Conditionally-end-a-running-job-gracefully) to gracefully end a running job.

wbeuil (approved):
LGTM

PR #67, opened 2021-01-05
Several improvements on acceptance tests

eliecharra:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | no
| 🚀 New feature?   | no
| ⚠ Deprecations?   | no
| ❌ BC Break       | no
| 🔗 Related issues | #...
| ❓ Documentation  | no

## Description

- Add filter on aws_instance test
- Add terraform locks
- Do not run terraform init if already done previously
- Do not use chdir in tests (it could cause race if we want to switch to parallel tests, uses --from arg instead)
- Handle terraform destroy errors

### DIFF
--- a/test/acceptance/testing.go
+++ b/test/acceptance/testing.go
@@ -8,6 +8,7 @@ import (
 	"io/ioutil"
 	"os"
 	"os/exec"
+	"path"
 	"testing"
 
 	"github.com/cloudskiff/driftctl/pkg/analyser"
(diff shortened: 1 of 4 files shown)


test/acceptance/testing.go
```diff
@@ -147,18 +158,33 @@ func Run(t *testing.T, c AccTestCase) {
...
+
+	err = c.terraformApply()
 	if err != nil {
 		t.Fatal(err)
 	}
-	defer terraformDestroy()
+
+	defer func() {
+		os.Setenv("CHECKPOINT_DISABLE", checkpoint)
+		err := c.terraformDestroy()
```
wbeuil, line 183: Terraform doesn't do any checks in a `terraform destroy` ?
Should we instead reset the `CHECKPOINT_DISABLE` after the destroy ?

eliecharra: > Terraform doesn't do any checks in a terraform destroy ?

I don't know

> Should we instead reset the CHECKPOINT_DISABLE after the destroy ?

Sure this is a nice catch thanks !

eliecharra: Fixed

wbeuil (approved):
LGTM

Issue #59, opened 2021-01-04
ERRO[0009] unsupported attribute "enclave_options"


sjourdan:
**Description**
I got this error: `ERRO[0009] unsupported attribute "enclave_options"` 
This is related to a simple EC2 instance resource: 

```
# Create AWS EC2 Instance
resource "aws_instance" "main" {
  ami           = data.aws_ami.ubuntu.id
  instance_type = "t2.nano"

  tags = {
    Name  = var.name
    TTL   = var.ttl
    owner = "${var.name}-guide"
  }
}

```

This is new from 3.22.0 https://github.com/hashicorp/terraform-provider-aws/blob/master/CHANGELOG.md#3220-december-18-2020

**Example**
I was following the hashicorp tutorial [here](https://learn.hashicorp.com/tutorials/terraform/secrets-vault).

See the tfstate attached

[terraform.zip](https://github.com/cloudskiff/driftctl/files/5764961/terraform.zip)

** Context ** 

https://docs.aws.amazon.com/enclaves/latest/user/nitro-enclave.html

_AWS Nitro Enclaves is an Amazon EC2 feature that allows you to create isolated execution environments, called enclaves, from Amazon EC2 instances. Enclaves are separate, hardened, and highly constrained virtual machines. They provide only secure local socket connectivity with their parent instance._

It's set to false by default

eliecharra:
@sjourdan We only support aws provider version 3.19, should we plan an update to 3.22 in the next version ?

sjourdan:
Good catch! Yes I think we should try to keep our tool as synchronized as possible. Let's add this to a future release 👍🏻

eliecharra:
We could go further and ignore incompatible fields to allow us to handle state generated with a newer version.
We also should stop execution and do not display unmanaged resources when a resource cannot be read from state, this output inconsistent result to the user.

eliecharra:
@sjourdan We fixed this one by ignoring fields not known in our current provider schema. Note that provider version was not upgraded.

PR #75, opened 2021-01-06
Ignore unknown attributes error in state

eliecharra:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | yes
| 🚀 New feature?   | no
| ⚠ Deprecations?   | no
| ❌ BC Break       | no
| 🔗 Related issues | #59 
| ❓ Documentation  | no

## Description

This PR allow us to ignore new fields from a state generated from a newer version of terraform provider.
It allow us to read state coming from aws provider version 3.22 if we currently only support version 3.19.
This PR also changes error mgmt during state reading, if we cannot decode a part of the state, we stop execution. It prevent us to return inconsistent drift result as some resources from the state are missing.

## References 

- https://github.com/zclconf/go-cty/issues/60

### DIFF
--- a/pkg/iac/terraform/state/terraform_state_reader.go
+++ b/pkg/iac/terraform/state/terraform_state_reader.go
@@ -1,20 +1,19 @@
 package state
 
 import (
-	"github.com/cloudskiff/driftctl/pkg/remote/deserializer"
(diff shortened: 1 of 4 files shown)


wbeuil (approved):
Huge feature and works perfectly as intended !

Issue #94, opened 2021-01-11
Support aws_route


eliecharra:
**Description**
Add support for aws_route

Issue #19, opened 2020-12-11
Support aws_default_route_table


eliecharra:
**Description**
Add support for aws_default_route_table

sjourdan:
works well! (1f66e51) 

```
Found drifted resources:
  - rtb-07bc7fefffb520e13 (aws_default_route_table):
    ~ Tags.Terraform: "true" => "false"
```

Issue #18, opened 2020-12-11
Support aws_route_table


eliecharra:
**Description**
Add support for aws_route_table

PR #116, opened 2021-01-14
Add aws_route, aws_route_table, aws_default_route_table support

eliecharra:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | no
| 🚀 New feature?   | yes
| ⚠ Deprecations?   | no
| ❌ BC Break       | no
| 🔗 Related issues | #94 #18 #19 
| ❓ Documentation  | yes


moadibfr (approved):

pkg/remote/aws/route_table_supplier.go
```diff
@@ -0,0 +1,133 @@
...
+		}
+		if isMain {
+			s.defaultRouteTableRunner.Run(func() (cty.Value, error) {
+				return s.readRouteTable(res, true)
+			})
+			continue
+		}
+		s.routeTableRunner.Run(func() (cty.Value, error) {
+			return s.readRouteTable(res, false)
+		})
```
wbeuil, line 64: We should first run one of the runner and wait for it to end before running and waiting for the other one. Otherwise we could have race conditions.

wbeuil (changes requested):
Need to change the possible race condition between both runners. Otherwise LGTM

pkg/remote/aws/route_table_supplier.go
eliecharra: It does not seem to be a race condition as we are using two separated runners.

Issue #93, opened 2021-01-11
Support aws_default_security_group


eliecharra:
**Description**
Add support for aws_default_security_group

sjourdan:
I could detect changes from a default SG!

PR #122, opened 2021-01-15
Add aws_default_security_group

wbeuil:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | no
| 🚀 New feature?   | yes
| ⚠ Deprecations?   | no
| ❌ BC Break       | no
| 🔗 Related issues | #93 
| ❓ Documentation  | yes


pkg/remote/aws/init.go
```diff
@@ -46,7 +46,7 @@ func Init(alerter *alerter.Alerter) error {
 	resource.AddSupplier(NewLambdaFunctionSupplier(provider.Runner().SubRunner(), lambda.New(provider.session)))
 	resource.AddSupplier(NewDBSubnetGroupSupplier(provider.Runner().SubRunner(), rds.New(provider.session)))
 	resource.AddSupplier(NewDBInstanceSupplier(provider.Runner().SubRunner(), rds.New(provider.session)))
-	resource.AddSupplier(NewVPCSecurityGroupSupplier(provider.Runner().SubRunner(), ec2.New(provider.session)))
+	resource.AddSupplier(NewVPCSecurityGroupSupplier(provider.Runner(), ec2.New(provider.session)))
```
eliecharra, line 49: We need to use a new sub runner to parallelize processing like other suppliers

pkg/iac/deserializers.go
```diff
@@ -26,6 +26,7 @@ func Deserializers() []deserializer.CTYDeserializer {
 		awsdeserializer.NewLambdaFunctionDeserializer(),
 		awsdeserializer.NewDBSubnetGroupDeserializer(),
 		awsdeserializer.NewVPCSecurityGroupDeserializer(),
+		awsdeserializer.NewDefaultSecurityGroupDeserializer(),
```
eliecharra, line 29: Maybe we should name it `DefaultVPCSecurityGroupDeserializer` to be more consistent in term of naming with `VPCSecurityGroupDeserializer`

pkg/remote/aws/vpc_security_group_supplier.go
```diff
@@ -16,59 +16,102 @@ import (
...
+	return resources, nil
 }
 
 func (s VPCSecurityGroupSupplier) readSecurityGroup(securityGroup ec2.SecurityGroup) (cty.Value, error) {
-	id := aws.StringValue(securityGroup.GroupId)
-	resSg, err := s.reader.ReadResource(terraform.ReadResourceArgs{
-		Ty: resourceaws.AwsSecurityGroupResourceType,
-		ID: id,
+	var Ty resource.ResourceType = resourceaws.AwsSecurityGroupResourceType
+	if securityGroup.GroupName != nil && *securityGroup.GroupName == "default" {
```
eliecharra, line 85: It could be better to have this logic once, could you make an unexported method in this supplier and reuse it there and in `listSecurityGroups`, or pass a `isDefault` bool to this method when looping on resources.

pkg/remote/aws/vpc_security_group_supplier_test.go
```diff
@@ -4,62 +4,66 @@ import (
...
 			test:    "with security groups",
 			dirName: "vpc_security_group_multiple",
-			securityGroupsPages: mocks.DescribeSecurityGroupsPagesOutput{
-				{
-					false,
-					&ec2.DescribeSecurityGroupsOutput{
-						SecurityGroups: []*ec2.SecurityGroup{
-							{
-								GroupId: aws.String("sg-0254c038e32f25530"),
+			mocks: func(client *mocks.FakeEC2) {
```
eliecharra, line 49: 👍🏻

pkg/resource/aws/aws_security_group_test.go
```diff
@@ -0,0 +1,28 @@
+package aws_test
+
+import (
+	"testing"
+
+	"github.com/cloudskiff/driftctl/test/acceptance"
+)
+
+func TestAcc_AwsSecurityGroup(t *testing.T) {
```
eliecharra, line 9: Awesome ! One more an acceptance test 👍🏻

eliecharra (changes requested):
Awesome work, clean and concise 👍🏻

pkg/iac/deserializers.go
wbeuil: It could be a good discussion to have it normalized indeed. We need to make it a way so that new contributor must not have to think about this kind of naming for a deserializer and a supplier. Let me explain.

On terraform the resource is aws_default_security_group and aws_security_group. At first, if I remembered well, I decided to name it so that the template would be `service_resource`, thus `vpc_security_group` or `ec2_ami` or `vpc_subnet`. But as far as we are now, it is more something like juste `resource`, thus it would be `security_group` or `subnet`.

We need to get it over with now so that we can change past resource with wrong naming.

pkg/remote/aws/init.go
wbeuil: I did it this way because for vpc and subnet we did it that way.

eliecharra: You are right my bad we create sub runner in constructor ...

pkg/remote/aws/vpc_security_group_supplier_test.go
wbeuil: What about past resource with wrong or different mocks ?

pkg/iac/deserializers.go
eliecharra: Hmm let's move the topic to another PR, we need to rethink resources / suppliers and deserializer naming to avoid collisions.

pkg/remote/aws/vpc_security_group_supplier_test.go
eliecharra: Not very important atm, we can do that later in dedicated PR

eliecharra (approved):

PR #124, opened 2021-01-15
rename "remotes" to "IaaS Providers"

sjourdan:
_"Supported remotes"_ makes me think it's about the terraform state remote storage or something, and not about all the supported resources in this IaaS.

| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | no
| 🚀 New feature?   | no
| ⚠ Deprecations?   | no
| ❌ BC Break       | no
| 🔗 Related issues | #...
| ❓ Documentation  | yes 

## Description

Rename for clarity

### DIFF
--- a/doc/README.md
+++ b/doc/README.md
@@ -18,6 +18,6 @@ Details of reported data can be found [here](./cmd/flags/error-reporting.md)
   - Scan
     - [Output format](cmd/scan/output.md)
     - [Filtering resources](cmd/scan/filter.md)
-    - [Supported remotes](cmd/scan/supported_resources/README.md)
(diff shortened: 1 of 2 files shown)


eliecharra:
A provider like Github is not really an IaaS provider, we should found a more generic naming IMHO.

wbeuil:
> A provider like Github is not really an IaaS provider, we should found a more generic naming IMHO.

I agree with this statement. We need to think like Terraform about that kind of naming.

eliecharra:
@sjourdan Can you open a discussion for this topic ? I'm closing this PR.

Issue #95, opened 2021-01-11
Support aws_route_table_association


eliecharra:
**Description**
Support for aws_route_table_association

PR #128, opened 2021-01-18
Add support for aws_route_table_association

eliecharra:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | no
| 🚀 New feature?   | yes
| ⚠ Deprecations?   | no
| ❌ BC Break       | no
| 🔗 Related issues | #95 
| ❓ Documentation  | yes

## Description

Add support for aws_route_table_association


pkg/remote/aws/route_table_association_supplier.go
```diff
@@ -0,0 +1,102 @@
...
+		client,
+		terraform.NewParallelResourceReader(runner),
+	}
+}
+
+func (s RouteTableAssociationSupplier) Resources() ([]resource.Resource, error) {
+
+	tables, err := listRouteTables(s.client)
+	if err != nil {
+		logrus.Error(err)
```
moadibfr, line 37: I saw this a lot on my pr. It's not a really good practice in go to log the error and return it.
You might want to wrap it to add this to the stacktrace but there is already a caller that log the same error.

pkg/resource/aws/testdata/acc/aws_route_table_association/.terraform.lock.hcl
```diff
@@ -0,0 +1,20 @@
+# This file is maintained automatically by "terraform init".
```
moadibfr, line 1: does this belong in the pr ? ^^

moadibfr (changes requested):

pkg/resource/aws/testdata/acc/aws_route_table_association/.terraform.lock.hcl
eliecharra: Yes it is, it't the dependency lock file for acceptance tests

pkg/remote/aws/route_table_association_supplier.go
eliecharra: Good catch, I fixed this one.
How do you handle context in errors ? If I want to return an error but with local context to be visible on caller log ?

image

moadibfr: You can wrap the error https://godoc.org/github.com/pkg/errors#Wrap
But this only allow to add to the stacktrace and error message.

pkg/resource/aws/testdata/acc/aws_route_table_association/.terraform.lock.hcl
moadibfr: Ok got it :+1:

pkg/remote/aws/route_table_association_supplier.go
eliecharra: A custom error type may be a solution too ?

wbeuil: Is this something that we could formalize in terms of kind of error ? As @eliecharra said, if we find a 403 error we could "Wrap" the err in a `forbidden` error ?

moadibfr: > A custom error type may be a solution too ?

yep we could start making our own error type and add a context map for our logger for example. But still wrapping is cool that way we could print the error stack when debugging

moadibfr: > Is this something that we could formalize in terms of kind of error ? As @eliecharra said, if we find a 403 error we could "Wrap" the err in a `forbidden` error ?

yep it's a common pattern to have type or singleton to identify error source.
But in what I did 403 are completely hidden to the outside.

PR #141, opened 2021-01-22
only send one alert for computed fieds

moadibfr:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | no
| 🚀 New feature?   | no
| ⚠ Deprecations?   | no
| ❌ BC Break       | yes
| 🔗 Related issues | .
| ❓ Documentation  | no

## Description

Change driftctl behavior when it find drift on computed field.
We use to send one alert per computed field and have a separate method to display only one alert on console output.
This pr will change that so only one alert is sent and all alerts are displayed on console output (like on the json output).

image


pkg/analyser/analyzer.go
```diff
@@ -64,15 +63,19 @@ func (a Analyzer) Analyze(remoteResources, resourcesFromState []resource.Resourc
...
 			}
 			if len(changelog) > 0 {
 				analysis.AddDifference(Difference{
 					Res:       stateRes,
 					Changelog: changelog,
 				})
-				a.sendAlertOnComputedField(stateRes, changelog)
 			}
+
+			a.sendAlertOnComputedField(haveComputedDiff)
```
eliecharra, line 78: Maybe we do not need a dedicated function for this anymore.

It will be more straigthforward imho

```go
if haveComputedDiff {
  a.alerter.SendAlert("",
		  alerter.Alert{
			  Message: "You have diffs on computed fields, check the documentation for potential false positive drifts",
		  })
}
```

eliecharra (changes requested):
Nice 👍🏻 Only a minor remark

pkg/analyser/analyzer.go
moadibfr: :+1:

eliecharra (approved):

PR #152, opened 2021-01-25
Handle Access denied errors when listing resources by triggering alerts 

moadibfr:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | no
| 🚀 New feature?   | yes
| ⚠ Deprecations?   | no
| ❌ BC Break       | yes
| 🔗 Related issues | #9
| ❓ Documentation  | no

## Description
Send alert and ignore resource when driftctl doesn't have the rights to list them

Screenshot from 2021-01-27 19-15-37


pkg/remote/aws/db_instance_supplier.go
```diff
@@ -43,8 +51,11 @@ func (s DBInstanceSupplier) Resources() ([]resource.Resource, error) {
 	resourceList, err := listAwsDBInstances(s.client)
 
 	if err != nil {
-		logrus.Error(err)
-		return nil, err
+		handled := handleListAwsError(err, resourceaws.AwsDbInstanceResourceType, s.alerter)
+		if !handled {
+			return nil, err
+		}
+		return []resource.Resource{}, nil
```
eliecharra, line 58: Can we return err in any case there and handle error in the scanner more globally ? 
We can use a custom error to add local scope context like the resource type.

I know it changes original architecture of this PR but there will be a single point for error handling.

We could do something like this in supplier 

```go
if err != nil {
  return errorWithContext(err, resourceaws.AwsDbInstanceResourceType)
}
```

and in the parent call on scanner side

```go
	res, err := supplier.Resources()
	if err != nil {
		if handleListAwsError(err, s.alerter) {
			return nil, nil // or maybe return []resource.Resource{}, nil
		}
		return nil, err
	}
```

pkg/remote/aws/db_instance_supplier_test.go
```diff
@@ -72,10 +82,41 @@ func TestDBInstanceSupplier_Resources(t *testing.T) {
...
+					},
+				},
+			},
+			wantAlert: alerter.Alerts{},
+			err:       nil,
+		},
+		{
+			test:                "Cannot list db instances",
+			dirName:             "db_instance_empty",
+			instancesPagesError: awserr.NewRequestFailure(nil, 403, ""),
```
eliecharra, line 112: Maybe it's a good option to start using mockery generated mock to return error instead of our own mocks for this case ?

eliecharra (commented):
First pass and got and architectural question

pkg/remote/aws/aws_error_handler.go
```diff
@@ -0,0 +1,33 @@
...
+		if reqerr.StatusCode() == 403 {
+
+			message := fmt.Sprintf("Ignoring %s from drift calculation: Listing %s is forbidden.", typ, typ)
+			if forbiddenTyp != "" {
+				message = fmt.Sprintf("Ignoring %s from drift calculation. Listing %s is forbidden.", typ, forbiddenTyp)
+			}
+			logrus.Debugf(message)
+			alertr.SendAlert(typ, alerter.Alert{
+				Message:              message,
+				ShouldIgnoreResource: true,
```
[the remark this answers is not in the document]
moadibfr: isn't the goal to ignore the false drift caused by the forbidden listing error ?

pkg/remote/aws/db_instance_supplier.go
moadibfr: hmm That sounds nice I'll look into this !

pkg/remote/aws/db_instance_supplier_test.go
moadibfr: yeah it would have been easier for me but I didn'want to change all our test for this pr

pkg/remote/aws/aws_error_handler.go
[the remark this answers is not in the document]
eliecharra: You are right, In my mind as listing do not return resources we do not need to ignore anything, but I forgot resources coming from the state 😅 So nervermind about this comment my bad

pkg/remote/aws/db_instance_supplier.go
moadibfr: last push is something with that idea please tell me what you think

pkg/remote/aws/db_instance_supplier.go
```diff
@@ -43,8 +48,7 @@ func (s DBInstanceSupplier) Resources() ([]resource.Resource, error) {
 	resourceList, err := listAwsDBInstances(s.client)
 
 	if err != nil {
-		logrus.Error(err)
-		return nil, err
+		return nil, NewBaseListError(err, resourceaws.AwsDbInstanceResourceType, resourceaws.AwsDbInstanceResourceType)
```
eliecharra, line 51: Hmm it's sad to have to send two types, 90% of the time this is useless as it only concern resources that require multiples levels of listing. Any idea how to handle this in a better way ?

eliecharra: Maybe using a struct to pass context

```go
NewBaseListError(err, ErrorListContext{supplierType: resourceaws.AwsDbInstanceResourceType})
NewBaseListError(err, ErrorListContext{supplierType: resourceaws.AwsDbInstanceResourceType, erroredListType: resourceaws.AwsDbInstanceResourceType })

```

pkg/remote/aws/error.go
```diff
@@ -0,0 +1,61 @@
+package aws
+
+import "fmt"
+
+type ListError interface {
```
eliecharra, line 5: `ListError` sound too much generic imho, what about something like `ResourceEnumerationError` ?

pkg/remote/aws/error.go
```diff
@@ -0,0 +1,61 @@
...
+	error
```
eliecharra, line 6: As `SupplierError` already wrap `error` is it required to compose it there too ?

pkg/remote/aws/iam_access_key_supplier_test.go
```diff
@@ -107,6 +113,37 @@ func TestIamAccessKeySupplier_Resources(t *testing.T) {
...
 		},
+		{
+			test:    "Cannot list iam user",
+			dirName: "iam_access_key_empty",
+			mocks: func(client *mocks.FakeIAM) {
+				client.On("ListUsersPages",
+					&iam.ListUsersInput{},
+					mock.MatchedBy(func(callback func(res *iam.ListUsersOutput, lastPage bool) bool) bool {
+						return true
+					})).Return(awserr.NewRequestFailure(nil, 403, ""))
```
eliecharra, line 124: 👍🏻

pkg/remote/aws/iam_role_supplier.go
```diff
@@ -40,7 +46,7 @@ func awsIamRoleShouldBeIgnored(roleName string) bool {
 }
 
 func (s IamRoleSupplier) Resources() ([]resource.Resource, error) {
-	roles, err := listIamRoles(s.client)
+	roles, err := listIamRoles(s.client, resourceaws.AwsIamRoleResourceType)
```
eliecharra, line 49: Maybe we could add a method to resource supplier interface `SupportedTypes() []string` ? It will be usefull in the future to avoid run filtered suppliers. Could you open a refactoring issues with this details if you agree ?

pkg/remote/aws/route_table_supplier.go
```diff
@@ -38,9 +38,8 @@ func NewRouteTableSupplier(runner *parallel.ParallelRunner, client ec2iface.EC2A
 
 func (s RouteTableSupplier) Resources() ([]resource.Resource, error) {
 
-	retrievedRouteTables, err := listRouteTables(s.client)
+	retrievedRouteTables, err := listRouteTables(s.client, aws.AwsRouteTableResourceType)
```
eliecharra, line 41: Why not handling error outside `listRouteTables` like in other suppliers ? It will avoid to pass the type everytime

eliecharra: image

eliecharra (changes requested):

pkg/error/aws_list_error_handler.go
```diff
@@ -0,0 +1,35 @@
+package error
```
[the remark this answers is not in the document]
moadibfr: Yep I didn't really know what to do with it. I could also just move this to scanner as for know it's the only user of this handler.

pkg/remote/aws/error.go
```diff
@@ -0,0 +1,61 @@
...
+	ListedTypeError() string
+}
+type SupplierError interface {
+	error
+	RootCause() error
+	SupplierType() string
+	Context() map[string]string
+}
+
+type BaseSupplierError struct {
```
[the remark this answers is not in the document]
moadibfr: Ho sorry I thought I renamed this my goal wah to have AWS specific struct and interface for generics error. So structs should be in aws package and interfaces in remote. But I kinda ask myself if there would be any needs for this.

pkg/remote/aws/error.go
```diff
@@ -0,0 +1,61 @@
...
+import "fmt"
+
+type ListError interface {
+	error
+	SupplierError
+	ListedTypeError() string
+}
+type SupplierError interface {
+	error
+	RootCause() error
```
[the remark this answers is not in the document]
moadibfr: We go go with rooterror (it's not necessarly the parrent if you wrap multiple time for example).
I was copying errors pkg (https://github.com/pkg/errors/issues/18)

go.sum
```diff
@@ -148,6 +148,7 @@ github.com/codegangsta/inject v0.0.0-20150114235600-33e0aa1cb7c0/go.mod h1:4Zcju
 github.com/coreos/bbolt v1.3.0/go.mod h1:iRUV2dpdMOn7Bo10OQBFzIJO9kkE559Wcmn+qkEiiKk=
 github.com/coreos/bbolt v1.3.2/go.mod h1:iRUV2dpdMOn7Bo10OQBFzIJO9kkE559Wcmn+qkEiiKk=
 github.com/coreos/etcd v3.3.10+incompatible/go.mod h1:uF7uidLiAD3TWHmW31ZFd/JWoc32PjwdhPthX9715RE=
+github.com/coreos/etcd v3.3.13+incompatible h1:8F3hqu9fGYLBifCmRCJsicFqDx/D68Rt3q1JMazcgBQ=
```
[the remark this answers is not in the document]
moadibfr: hmm probably goland auto import side effect I'll remote this

pkg/error/aws_list_error_handler.go
```diff
@@ -0,0 +1,35 @@
...
+
+func HandleListAwsError(err error, alertr *alerter.Alerter) (handled bool) {
+	listError, ok := err.(aws.ListError)
+	if !ok {
+		return false
+	}
+
+	reqerr, ok := listError.RootCause().(awserr.RequestFailure)
+	if !ok {
+		return false
```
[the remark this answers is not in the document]
moadibfr: :+1:

pkg/remote/aws/db_instance_supplier.go
moadibfr: I could also make a special constructor if you prefer.

pkg/remote/aws/error.go
moadibfr: :+1:

pkg/remote/aws/error.go
moadibfr: Good question I think it's not required indeed

pkg/remote/aws/iam_role_supplier.go
moadibfr: yeah I was thinking about that but the pr was already too large :sweat_smile:

pkg/remote/aws/route_table_supplier.go
moadibfr: this allows to factorize error trhowing for all the user of listRouteTables. Listbucket is a method of the client so it's different I don't think it should throw supplier errors.

pkg/remote/aws/iam_role_supplier.go
moadibfr: https://github.com/cloudskiff/driftctl/issues/160

pkg/remote/aws/s3_bucket_notification_supplier.go
```diff
@@ -74,6 +83,8 @@ func (s *S3BucketNotificationSupplier) listBucketNotificationConfiguration(name,
 			},
 		})
 		if err != nil {
+			logrus.Errorf("ERROOORRR %s", reflect.TypeOf(err))
```
wbeuil, line 86: I think you forgot to remove this one ^^

pkg/error/aws_list_error_handler.go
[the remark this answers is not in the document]
eliecharra: Yeah I think this could be a good move. We could image have a collection of error handlers of multiples kind in the future. It can make sense to create a dedicated pkg under `remote/` WDYT ?

pkg/error/aws_list_error_handler.go
```diff
@@ -0,0 +1,35 @@
...
+	"fmt"
+
+	"github.com/cloudskiff/driftctl/pkg/remote/aws"
+
+	"github.com/aws/aws-sdk-go/aws/awserr"
+	"github.com/cloudskiff/driftctl/pkg/alerter"
+	"github.com/sirupsen/logrus"
+)
+
+func HandleListAwsError(err error, alertr *alerter.Alerter) (handled bool) {
```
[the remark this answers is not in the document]
eliecharra: I was wondering myself if we should rather return the original err or nil if error is correctly handled.

pkg/remote/aws/error.go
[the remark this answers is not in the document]
eliecharra: It's ok for me if it's coming from the standard error library too !

pkg/remote/aws/db_instance_supplier_test.go
moadibfr: https://github.com/cloudskiff/driftctl/issues/161

pkg/remote/aws/error.go
eliecharra: I'd rather like the most verbose one

pkg/remote/aws/s3_bucket_notification_supplier.go
moadibfr: :disappointed:

pkg/error/aws_list_error_handler.go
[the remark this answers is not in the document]
eliecharra: Actually we do not need any AWS specific error handler imho. Enumeration error can occur from any supplier, I think we should have one error handler to handle every enum error coming from any provider.

pkg/remote/aws/error.go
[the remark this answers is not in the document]
eliecharra: Answered you there https://github.com/cloudskiff/driftctl/pull/152#discussion_r566085767

pkg/error/aws_list_error_handler.go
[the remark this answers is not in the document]
wbeuil: What about if we need them in our middlewares ?

pkg/remote/aws/error.go
```diff
@@ -0,0 +1,61 @@
...
+}
+
+func (b *BaseSupplierError) Context() map[string]string {
+	return b.context
+}
+
+type BaseListError struct {
+	BaseSupplierError
+	listedTypeError string
+}
```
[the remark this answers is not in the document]
moadibfr: Yes it happens a lot (https://github.com/aws/aws-sdk-go/blob/master/aws/awserr/error.go for ex)

pkg/remote/aws/iam_role_supplier.go
eliecharra: Thanks !

pkg/error/aws_list_error_handler.go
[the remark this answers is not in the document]
eliecharra: That's another topic for me, we should make theses error handler only related to remote scan operations. 
Our actual middleware are not meant to talk with remote, their goal is only to work on resourceFromState and resourceFromRemote slices.

pkg/remote/aws/route_table_supplier.go
eliecharra: Ooops sorry wrong screenshot example my bad 🙏🏻 

image

I mean it does not makes sense for me to factorize this error wraping as its bounded to the caller context. Thus you should updated `listRouteTables` method signature to add the type only for this error creation., we could `return nil, err` in `listRouteTables` and handle listing error in `Resources()` like in `EC2EbsSnapshotSupplier`.

Factorizing stuff just to have less line of codes is not the right thing to do for me in this context.

eliecharra: ok lets close this as sometime we have two level of list method nesting, we need to make our suppliers more independent each others. So let's open a refactoring issue and close resolve this thread.

pkg/remote/aws/iam_user_policy_supplier.go
```diff
@@ -3,7 +3,10 @@ package aws
 import (
 	"fmt"
 
+	error2 "github.com/cloudskiff/driftctl/pkg/remote/error"
```
eliecharra, line 6: Could you do a pass to use same import name in all places ? `remoteerror` is better than `error2`

pkg/remote/resource_list_error_handler.go
```diff
@@ -0,0 +1,35 @@
...
+	"fmt"
+
+	remoteerror "github.com/cloudskiff/driftctl/pkg/remote/error"
+
+	"github.com/aws/aws-sdk-go/aws/awserr"
+	"github.com/cloudskiff/driftctl/pkg/alerter"
+	"github.com/sirupsen/logrus"
+)
+
+func HandleResourceListError(err error, alertr *alerter.Alerter) error {
```
eliecharra, line 13: `HandleResourceEnumerationError`

wbeuil: Should we also change the name of the file ?

eliecharra: Indeed

pkg/error/aws_list_error_handler.go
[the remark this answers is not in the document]
eliecharra: Still not OK, you should have a test with `NewResourceEnumerationErrorWithType`

```go
		{
			name:       "Not Handled root error type",
			err:        remoteerror.NewResourceEnumerationErrorWithType(errors.New("error"), resourceaws.AwsVpcResourceType, resourceaws.AwsVpcResourceType),
			wantAlerts: map[string][]alerter.Alert{},
			wantErr:    true,
		},
```

eliecharra (changes requested):
One test case to fix, rename file to reflect changes on `HandleResourceEnumerationError` naming and we are good to go !

pkg/error/aws_list_error_handler.go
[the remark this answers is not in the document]
eliecharra: @moadibfr up

eliecharra (approved):

PR #158, opened 2021-01-28
Middleware to explode inline policy

wbeuil:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | yes
| 🚀 New feature?   | no
| ⚠ Deprecations?   | no
| ❌ BC Break       | no
| 🔗 Related issues | none
| ❓ Documentation  | no

## Description

Bug found yesterday where it was possible to create an inline policy that was viewed as unmanaged.
To be coherent with what we already have for the routes and security group rules, we should explode inline poilicy from state s3_bucket resource and create s3_bucket_policy from it.


eliecharra:
Very clear thanks for this catch 👍🏻

eliecharra (approved):

Issue #157, opened 2021-01-27
Disable warning in driftignore parsing


eliecharra:
**Description**
I would like to add comment in my driftignore, but when running driftctl it output multiples warnings.
We could skip lines starting with a `#` and empty lines

**Example**
```ignore
# intern laptop admin credentials
# TODO : remove tf-admin user and provision a key for intern user
aws_iam_user.tf-admin
aws_iam_access_key.XXXXXXXXXXXXXXX
aws_iam_policy_attachment.tf-admin-arn:aws:iam::aws:policy/AdministratorAccess

# Terraform bucket for state storage
aws_s3_bucket.remote-states
```

![image](https://user-images.githubusercontent.com/6154987/106042793-cee3b300-60dd-11eb-8c34-72317e8a7ef9.png)

sfenman:
Hi @eliecharra , I would like to give it a try if it's possible!

eliecharra:
Sure go on it would be awesome to merge your contribution !

PR #178, opened 2021-02-01
skip warnings on commented out and empty lines

sfenman:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | no
| 🚀 New feature?   | no
| ⚠ Deprecations?   | no
| ❌ BC Break       | no
| 🔗 Related issues | #157
| ❓ Documentation  | no

## Description

Created a condition to skip empty or commented out (starting with # lines) from checking.


pkg/filter/driftignore.go
```diff
@@ -39,36 +39,40 @@ func (r *DriftIgnore) readIgnoreFile() error {
...
 		typeVal := readDriftIgnoreLine(line)
 		nbArgs := len(typeVal)
-		if nbArgs < 2 {
-			logrus.WithFields(logrus.Fields{
-				"line": line,
-			}).Warnf("unable to parse line, invalid length, got %d expected >= 2", nbArgs)
-			continue
-		}
-		if nbArgs == 2 { // We want to ignore a resource (type.id)
+		if nbArgs == 0 || strings.HasPrefix(line, "#") {
```
eliecharra, line 42: `nbArgs == 0` is not exactly the check we want to ignore, we want to check if the line is empty

Your if directive should be moved just after `line := scanner.Text()`.
We do not want to call `readDriftIgnoreLine` for ignored lines.

You can also remove the `else` directive as you put a continue.

Something like below should do the job :

```diff
line := scanner.Text()
+if line == "" || strings.HasPrefix(line, "#") {
+  logrus.WithFields(logrus.Fields{
+    "line": line,
+  }).Debug("Skipped comment or empty line")
+  continue
+}
typeVal := readDriftIgnoreLine(line)
```

eliecharra (changes requested):

eliecharra:
Hi @sfenman thanks a lot for this PR 😃 , could you update the test suite to reflect your changes too ? 🙏🏻

sfenman:
Hi @eliecharra, thank you for the comments, I will update it shortly. I guess a new testing function is needed to test about empty and commented out lines explicitly right?

eliecharra:
> Hi @eliecharra, thank you for the comments, I will update it shortly. I guess a new testing function is needed to test about empty and commented out lines explicitly right?

I think you just have to update `drift_ignore_valid/.driftignore` by adding empty lines and comment.

We do not want to test that no warning are send at the moment at it will imply to use and inject a custom logger in driftignore logic to be able to mock it. So don't spend too much time on this 😃

sfenman:
I've made the changes based on your comments, but I couldn't omit the `else` directive because without it, the comments and the empty lines are still being passed to `readDriftIgnoreLine` .

eliecharra:
> I've made the changes based on your comments, but I couldn't omit the `else` directive because without it, the comments and the empty lines are still being passed to `readDriftIgnoreLine` .

You are missing the `continue` directive below in your `if` after the debug log, it will skip the line without the need for an else statement.

image

It will make the code more readable and avoid unnecessary else nesting.

eliecharra (approved):
Awesome, thanks a lot for this fix @sfenman 🙏🏻

wbeuil:
@all-contributors please add @sfenman for code

PR #182, opened 2021-02-02
add LIMITATIONS.md

sjourdan:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | no
| 🚀 New feature?   | no
| ⚠ Deprecations?   | no
| ❌ BC Break       | no
| 🔗 Related issues | #...
| ❓ Documentation  | yes 

## Description

We need a place to point our users to for known limitations., starting with the first we know about. 
It will keep getting better.

I placed it under `./doc` so it doesn't clutter the root of the repository and everything related to documentation is located at the same place.


doc/LIMITATIONS.md
```diff
@@ -0,0 +1,6 @@
+# Known Issues and Limitations
+
+## AWS Regions Limits
+
+- The user needs to use the same AWS region for both the scanned infrastructure and the S3 bucket where the Terraform state is stored (for example, a Terraform state stored on S3 on us-east-1 for an infrastructure to be scanned on us-west-1 won't work.).
+  - See the related [GitHub Discussion](https://github.com/cloudskiff/driftctl/discussions/130).
```
wbeuil, line 6: Can you remove the `.` inside the parentheses and aline the second `-`.

Thanks !

doc/LIMITATIONS.md
```diff
@@ -0,0 +1,6 @@
+# Known Issues and Limitations
+
+## AWS Regions Limits
+
+- The user needs to use the same AWS region for both the scanned infrastructure and the S3 bucket where the Terraform state is stored (for example, a Terraform state stored on S3 on us-east-1 for an infrastructure to be scanned on us-west-1 won't work.).
```
eliecharra, line 5: `the same AWS region` and the same credentials too !

eliecharra (changes requested):
Maybe we should add a precision, actually we can only scan one provider, so for a state using multiples aws provider with aliases we cannot scan every provider.

We can handle multiples states but only one `--to` flag so only one remote.

eliecharra (approved):
LGTM

wbeuil (approved):

PR #193, opened 2021-02-03
Handle invalid route crash in middlewares

eliecharra:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | yes
| 🚀 New feature?   | no
| ⚠ Deprecations?   | no
| ❌ BC Break       | no
| 🔗 Related issues | #191 
| ❓ Documentation  | no

## Description

image


pkg/middlewares/aws_route_table_expander_test.go
```diff
@@ -208,3 +211,85 @@ func TestAwsRouteTableExpander_Execute(t *testing.T) {
...
+
+func TestAwsRouteTableExpander_ExecuteWithInvalidRoutes(t *testing.T) {
+
+	mockedAlerter := &mocks.AlerterInterface{}
+	mockedAlerter.On("SendAlert", aws.AwsRouteTableResourceType, alerter.Alert{
+		Message: "Skipped invalid route found in state for aws_route_table.table_from_state",
+	})
+	mockedAlerter.On("SendAlert", aws.AwsDefaultRouteTableResourceType, alerter.Alert{
+		Message: "Skipped invalid route found in state for aws_default_route_table.default_table_from_state",
+	})
```
wbeuil, line 223: Why do you mock here the alerter instead of calling `alerter := alerter.NewAlerter()` and assert that at the end we have indeed the two alerts in its map of alerts ? 

image

eliecharra: The scope of this test is not to test the alerter itself but only the behaviour of the middleware handling this case.

Unit test should stay minimalist as possible IMHO

wbeuil: Make sense

wbeuil (approved):
LGTM

PR #203, opened 2021-02-03
Fix unexpected deleted resources in acceptance tests

eliecharra:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | yes
| 🚀 New feature?   | no
| ⚠ Deprecations?   | no
| ❌ BC Break       | no
| 🔗 Related issues | #...
| ❓ Documentation  | no

## Description

Cobra cmd seems to return flag twice when executed multiples times.
I moved cmd creation in the loop to start with a fresh cobra cmd each time.

First check

image

Second check

image


wbeuil (approved):
Nice catch, it makes more sense inside the loop

PR #207, opened 2021-02-04
Add retry mechanism for polling and use it to poll sqs queues during acc tests

wbeuil:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | yes
| 🚀 New feature?   | no
| ⚠ Deprecations?   | no
| ❌ BC Break       | no
| 🔗 Related issues | none
| ❓ Documentation  | no

## Description

Found error in CI while running acceptance tests. The problem in short is that SQS queues has a weird behaviour where creation/deletion takes time to propagate. TLDR, we need to poll to list sqs queues in a pre check.

### DIFF
--- a/pkg/resource/aws/aws_sqs_queue_policy_test.go
+++ b/pkg/resource/aws/aws_sqs_queue_policy_test.go
@@ -2,6 +2,11 @@ package aws_test
 
 import (
 	"testing"
+	"time"
+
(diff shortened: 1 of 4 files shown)


eliecharra (approved):
Totally awesome 🙏🏻 🎉

Issue #213, opened 2021-02-05
Add support for github_repository


eliecharra:

sjourdan:
Small typo methinks

```shell 
github[main ?]$ driftctl-0.5 scan --to github+tf
DEBU[0000] New provider library created
DEBU[0000] provider not found, downloading ... path=/home/sjourdan/.driftctl/plugins/linux_amd64/terraform-provider-github_v4.4.0
Downloading terraform provider: github
DEBU[0000] Downloading provider path=/home/sjourdan/.driftctl/plugins/linux_amd64 url=https://releases.hashicorp.com/terraform-provider-github/4.4.0/terraform-provider-github_4.4.0_linux_linux.zip
unsuccessful request to https://releases.hashicorp.com/terraform-provider-github/4.4.0/terraform-provider-github_4.4.0_linux_linux.zip: 403 Forbidden
```

`s/terraform-provider-github_4.4.0_linux_linux.zip/terraform-provider-github_4.4.0_linux_amd64.zip/g` ?

sjourdan:
job done! thanks!

sjourdan:
LGTM!! 🎉

PR #235, opened 2021-02-09
Make terraform provider generic

eliecharra:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | no
| 🚀 New feature?   | no
| ⚠ Deprecations?   | no
| ❌ BC Break       | no
| 🔗 Related issues | #213 
| ❓ Documentation  | no

## Description

Remove all related aws stuff from `TerraformProvider` and create an `AWSTerraformProvider`

⚠️ Architecture review please @moadibfr @wbeuil. I splitted commits in small reviewable parts.

### DIFF
--- a/pkg/remote/aws/provider.go
+++ b/pkg/remote/aws/provider.go
@@ -1,0 +1,70 @@
+package aws
+
+import (
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/cloudskiff/driftctl/pkg/remote/terraform"
+	tf "github.com/cloudskiff/driftctl/pkg/terraform"
+)
+
+type awsConfig struct {
+	AccessKey     string
+	SecretKey     string
+	CredsFilename string
+	Profile       string
+	Token         string
+	Region        string `cty:"region"`
+	MaxRetries    int
+
+	AssumeRoleARN         string
+	AssumeRoleExternalID  string
+	AssumeRoleSessionName string
+	AssumeRolePolicy      string
+
+	AllowedAccountIds   []string
+	ForbiddenAccountIds []string
+
+	Endpoints        map[string]string
+	IgnoreTagsConfig map[string]string
+	Insecure         bool
+
+	SkipCredsValidation     bool
+	SkipGetEC2Platforms     bool
+	SkipRegionValidation    bool
+	SkipRequestingAccountId bool
+	SkipMetadataApiCheck    bool
+	S3ForcePathStyle        bool
+}
+
+type AWSTerraformProvider struct {
+	*terraform.TerraformProvider
+	session *session.Session
+}
+
+func NewAWSTerraformProvider() (*AWSTerraformProvider, error) {
+	p := &AWSTerraformProvider{}
+	installer, err := tf.NewProviderInstaller(tf.ProviderConfig{
+		Key:     "aws",
+		Version: "3.19.0",
+		Postfix: "x5",
+	})
+	if err != nil {
+		return nil, err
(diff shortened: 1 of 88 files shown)


moadibfr (approved):

wbeuil (changes requested):
Nice rewrite of the architecture to make it more generic to create terraform provider.

My only comment, and I'm really against it, is the renaming of the `region` with `availability_zone`. In AWS, availability zones are really something that is completely different than regions as per the image below.

image

Availability Zones are multiple, isolated locations within each Region. It will be confusing to name it that way.

eliecharra:
> Availability Zones are multiple, isolated locations within each Region. It will be confusing to name it that way.

Agree, I haven't that in mind when looking for a new naming. We should find something generic enough between each provider, maybe we could use `alias` instead ?

wbeuil:
`alias` could do the trick. What do you think @moadibfr ?

eliecharra:
FYI alias came from this : https://www.terraform.io/docs/language/providers/configuration.html#alias-multiple-provider-configurations

pkg/remote/aws/provider.go
```diff
@@ -0,0 +1,70 @@
...
+	})
+	if err != nil {
+		return nil, err
+	}
+	p.session = session.Must(session.NewSessionWithOptions(session.Options{
+		SharedConfigState: session.SharedConfigEnable,
+	}))
+	tfProvider, err := terraform.NewTerraformProvider(installer, terraform.TerraformProviderConfig{
+		DefaultAlias: *p.session.Config.Region,
+		GetProviderConfig: func(availabilityZone string) interface{} {
```
wbeuil, line 58: Should be `alias (string)` here

pkg/remote/terraform/provider.go
```diff
@@ -10,76 +10,43 @@ import (
...
-
-	SkipCredsValidation     bool
-	SkipGetEC2Platforms     bool
-	SkipRegionValidation    bool
-	SkipRequestingAccountId bool
-	SkipMetadataApiCheck    bool
-	S3ForcePathStyle        bool
+type TerraformProviderConfig struct {
+	DefaultAlias      string
+	GetProviderConfig func(alias string) interface{}
```
wbeuil, line 27: Maybe just above the struct an explanation of what is the alias attribute.

Something like: `Alias refers to the 'meta-region' where the terraform provider will scan all resources`

eliecharra: WDYT of 

```
// "alias" in these struct is a way to namespaces gRPC client
// for example, if we need to read S3 bucket from AWS from multiples region, we'll have
// an alias per region, and the alias IS the region itself
// So we can query resources using a specific custom provider configuration
```

wbeuil: Nice !

Just quick typo here `"alias" in this ... to namespace ...` `from multiple AWS region, ...`

pkg/remote/aws/s3_bucket_inventory_supplier.go
```diff
@@ -94,7 +94,7 @@ func (s *S3BucketInventorySupplier) listBucketInventoryConfiguration(name, regio
 					Ty: aws.AwsS3BucketInventoryResourceType,
 					ID: id,
 					Attributes: map[string]string{
-						"aws_region": region,
+						"alias: region,
```
wbeuil, line 97: `"alias": region,`

wbeuil (dismissed):

Issue #210, opened 2021-02-05
Add support for aws_dynamodb_table


eliecharra:

sjourdan:
lots of detection, LGTM

![image](https://user-images.githubusercontent.com/419121/108397507-56719d00-7218-11eb-8c54-441df4a7c6b5.png)

PR #244, opened 2021-02-11
add support for dynamodb_table

moadibfr:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | no
| 🚀 New feature?   | yes
| ⚠ Deprecations?   | no
| ❌ BC Break       | no
| 🔗 Related issues | #210 
| ❓ Documentation  | yes

## Description

add support for dynamodb_table


pkg/resource/aws/aws_dynamodb_table_ext.go
```diff
@@ -0,0 +1,19 @@
+package aws
+
+import "github.com/cloudskiff/driftctl/pkg/resource"
+
+func (r *AwsDynamodbTable) NormalizeForState() (resource.Resource, error) {
+	// set default to empty struct for timeouts
+	if r.Timeouts == nil {
```
eliecharra, line 7: It's a Terraform only related field, for consistency purpose it's more straightforward to handle it like in other resource by adding a `diff:"-"` annotation.

pkg/remote/resource_enumeration_error_handler.go
```diff
@@ -21,9 +21,9 @@ func HandleResourceEnumerationError(err error, alertr *alerter.Alerter) error {
 		return err
 	}
 
-	if reqerr.StatusCode() == 403 {
+	if reqerr.StatusCode() == 403 || strings.Contains(reqerr.Code(), "AccessDenied") {
 		message := fmt.Sprintf("Ignoring %s from drift calculation: Listing %s is forbidden.", listError.SupplierType(), listError.ListedTypeError())
-		logrus.Debugf(message)
+		//logrus.Debugf(message)
```
eliecharra, line 26: You can remove this line 😉

pkg/remote/resource_enumeration_error_handler.go
```diff
@@ -21,9 +21,9 @@ func HandleResourceEnumerationError(err error, alertr *alerter.Alerter) error {
 		return err
 	}
 
-	if reqerr.StatusCode() == 403 {
+	if reqerr.StatusCode() == 403 || strings.Contains(reqerr.Code(), "AccessDenied") {
```
eliecharra, line 24: To be sure to catch only this case maybe we could check the status code too ?

```go
if reqerr.StatusCode() == 403 || (reqerr.StatusCode() == 400 && reqerr.Code() == "AccessDeniedException") {
```

pkg/remote/aws/test/dynamodb_table_list/results.golden.json
```diff
@@ -0,0 +1 @@
+[]
```
eliecharra, line 1: We expect two tables there ⚠️

eliecharra (changes requested):

pkg/remote/resource_enumeration_error_handler.go
moadibfr: From what we saw AccessDeniedException is always access denied it does not depends on the http code so I remove te 400 check but no matter what we do we won't catch everything, 403 on dynamo is still another error.

pkg/remote/aws/test/dynamodb_table_list/results.golden.json
moadibfr: no we don't it's the test for enumeration (list) error ^^

pkg/remote/resource_enumeration_error_handler.go
moadibfr: :facepalm:

pkg/resource/aws/aws_dynamodb_table_ext.go
moadibfr: ho you're right I thought these were real timeout not only used by tf...

pkg/remote/aws/test/dynamodb_table_list/results.golden.json
moadibfr: I could rename the test directory but I used the same naming for all the listing test...

eliecharra: Yeah the naming is pretty confusing but we'll refacto this in another PR

wbeuil (approved):
Seems good !

eliecharra (approved):

PR #258, opened 2021-02-15
Add support for github_repository

eliecharra:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | no
| 🚀 New feature?   | yes
| ⚠ Deprecations?   | no
| ❌ BC Break       | no
| 🔗 Related issues | #213 
| ❓ Documentation  | yes

## Description

Add support for github_repository

@cloudskiff/engineering ⚠️ architectural review required.


pkg/remote/terraform/provider.go
```diff
@@ -125,6 +125,12 @@ func (p *TerraformProvider) configure(alias string) error {
 		"alias": alias,
 	}).Debug("New gRPC client started")
 
+	fmt.Printf("Terraform provider initialized (name=%s", p.Config.Name)
```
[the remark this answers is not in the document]
eliecharra: Json output does not pipe the whole stdout, it will break if we use `--output json:///dev/stdout` but this is already broken actually. This is just a rework of the existing output by adding the name

pkg/remote/github/github_repository_supplier_test.go
```diff
@@ -0,0 +1,91 @@
...
+					},
+					{
+						Name: "driftctl-demos",
+					},
+				}, nil)
+			},
+			err: nil,
+		},
+		/*		{
+				test:    "cannot list github repository",
```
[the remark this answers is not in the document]
eliecharra: Sure good catch, I forgot to remove this one

pkg/remote/github/github_repository_supplier_test.go
```diff
@@ -0,0 +1,91 @@
...
+	"context"
+	"testing"
+
+	"github.com/cloudskiff/driftctl/pkg/parallel"
+	"github.com/cloudskiff/driftctl/pkg/resource"
+	ghdeserializer "github.com/cloudskiff/driftctl/pkg/resource/github/deserializer"
+	"github.com/cloudskiff/driftctl/pkg/terraform"
+	"github.com/cloudskiff/driftctl/test"
+	"github.com/cloudskiff/driftctl/test/goldenfile"
+	mocks2 "github.com/cloudskiff/driftctl/test/mocks"
```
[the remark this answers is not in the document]
eliecharra: 👍🏻 missed this one I will fix it

pkg/resource/github/deserializer/github_repository_deserializer.go
```diff
@@ -0,0 +1,43 @@
...
+	return github.GithubRepositoryResourceType
+}
+
+func (s GithubRepositoryDeserializer) Deserialize(rawResourceList []cty.Value) ([]resource.Resource, error) {
+	resources := make([]resource.Resource, 0)
+	for _, raw := range rawResourceList {
+		raw := raw
+		res, err := decodeRepository(&raw)
+		if err != nil {
+			logrus.Warnf("error when deserializing github_repository %+v : %+v", raw, err)
```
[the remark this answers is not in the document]
eliecharra: I just copy / paste structure from an existing provider, we are doing this everywhere. Maybe we need to refacto this, or maybe not if you are working on a generic deserializer ?

`Wrapf` is more appropriate that `WithMessage` btw as it also add the caller stack.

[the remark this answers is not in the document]
moadibfr: Yep it depends on what we want to do for wrap/message. I removed some log when I refactored supplier but yeah there is still some place where we log and return the same error after. This will probably cause the same log to be shown more than one time.
But yeah you are right it will be fixed on deserializer when the refactoring is done

pkg/remote/terraform/provider.go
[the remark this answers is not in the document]
moadibfr: hmm ok so it was already broken. I think it's a little sad that we have an abstract output but allow to print on other places. But ok it's not this pr topic :+1:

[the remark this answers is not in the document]
eliecharra: There is an open discussion about it there : https://github.com/cloudskiff/driftctl/discussions/56#discussioncomment-284917

doc/cmd/scan/supported_resources/github.md
```diff
@@ -0,0 +1,23 @@
...
+
+To use driftctl, we need credentials to make authenticated requests to github. Just like the terraform provider, we retrieve config from [environment variables](https://registry.terraform.io/providers/integrations/github/latest/docs#argument-reference).
+
+```bash
+$ GITHUB_TOKEN=14758f1afd44c09b7992073ccf00b43d GITHUB_ORGANIZATION=my-org driftctl scan --to github+tf
+```
+
+## Least privileged policy
+
+Below you can find the minimal scope required for driftctl to be able to scan every github supported resource
```
wbeuil, line 13: `every github supported resources`

doc/cmd/scan/supported_resources/github.md
```diff
@@ -0,0 +1,23 @@
...
+
+```shell
+repo # Required to enumerate public and private repos
+```
+
+**⚠️ Beware that if you don't set correct permissions for your token, no error will be send and every resource will appear as deleted from remote**
```
wbeuil, line 19: `, you won't see any errors and all resources will appear as ....`

pkg/iac/deserializers.go
```diff
@@ -53,5 +54,7 @@ func Deserializers() []deserializer.CTYDeserializer {
 		awsdeserializer.NewSNSTopicPolicyDeserializer(),
 		awsdeserializer.NewSNSTopicSubscriptionDeserializer(),
 		awsdeserializer.NewDynamoDBTableDeserializer(),
+
+		ghdeserializer.NewGithubRepositoryDeserializer(),
```
wbeuil, line 58: Maybe for another PR, we could create 2 different functions to list for the first AWS deserializers and for the second Github ones. And call them here.

I fear that it's gonna be messy in the near future in this function.

moadibfr (dismissed):

pkg/remote/github/provider.go
```diff
@@ -0,0 +1,60 @@
...
+		Key:     providerKey,
+		Version: "4.4.0",
+	})
+	if err != nil {
+		return nil, err
+	}
+	tfProvider, err := terraform.NewTerraformProvider(installer, terraform.TerraformProviderConfig{
+		Name:         providerKey,
+		DefaultAlias: p.getDefaultOwner(),
+		GetProviderConfig: func(owner string) interface{} {
```
wbeuil, line 33: Shouldn't it be `func(alias string) interface{} {`

pkg/remote/github/mock_GithubRepository_test.go
```diff
@@ -0,0 +1,33 @@
+// Code generated by mockery v2.3.0. DO NOT EDIT.
+
+package github
+
+import mock "github.com/stretchr/testify/mock"
+
+// MockGithubRepository is an autogenerated mock type for the GithubRepository type
+type MockGithubRepository struct {
+	mock.Mock
+}
```
wbeuil, line 10: Let's have all our mocks in the same folder as to not be lost when we need to find existed mocks. For AWS we put them in `mocks` root folder. Either we continue like that and even create a subfolder `github`, or like you did we can put them inside the `remote/<provider>` folder.

pkg/remote/github/repository.go
```diff
@@ -0,0 +1,120 @@
...
+		for _, repo := range query.Viewer.Repositories.Nodes {
+			results = append(results, repo.Name)
+		}
+		if !query.Viewer.Repositories.PageInfo.HasNextPage {
+			break
+		}
+		variables["cursor"] = githubv4.NewString(query.Viewer.Repositories.PageInfo.EndCursor)
+	}
+	return results, nil
+}
```
wbeuil, line 120: If I understand well here, we're going to have all our query types and thus all the listing functions in this repository file ? We don't follow what we've done for AWS where for each services, we had a `repository` file.

doc/cmd/scan/supported_resources/github.md
eliecharra: 🤦🏻

pkg/remote/github/provider.go
eliecharra: I've put owner there to make this parameter maybe more clear in github context.

* AWS alias = region
* Github alias = owner

I can totally put alias everywhere if you find it more clear

pkg/remote/github/repository.go
eliecharra: We don"t have the notion of "service" for Github. We got a single graphql endpoint.

I think it will be a lot of boilerplate to create a repository per resource.

pkg/iac/deserializers.go
eliecharra: Maybe, more generally we could think about how we want to handle tf states.

Do we really want to try to decode aws related resources when scanning github ?
But we may want to handle states with mixed providers in the future too ... I think it's too early to take a direction.

pkg/remote/github/provider.go
wbeuil: I'm fine with that

pkg/remote/github/repository.go
wbeuil: Ok so one repository for this single graphql endpoint. Got it

wbeuil (approved):

Issue #211, opened 2021-02-05
Add support for aws_route53_health_check


eliecharra:

sjourdan:
LGTM!

![image](https://user-images.githubusercontent.com/419121/108394919-8b302500-7215-11eb-9e9e-c58042f201f0.png)

PR #259, opened 2021-02-15
add support for route53 health check

moadibfr:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | no
| 🚀 New feature?   | yes
| ⚠ Deprecations?   | no
| ❌ BC Break       | no
| 🔗 Related issues | #211 
| ❓ Documentation  | yes 

## Description

add support for route53 health check


doc/cmd/scan/supported_resources/aws.md
```diff
@@ -142,7 +142,9 @@ As AWS documentation recommends, the below policy is granting only the permissio
                 "dynamodb:ListTagsOfResource",
                 "dynamodb:DescribeTimeToLive",
                 "dynamodb:DescribeTableReplicaAutoScaling",
-                "dynamodb:DescribeContinuousBackups"
+                "dynamodb:DescribeContinuousBackups",
+                "route53:ListHealthChecks",
+                "route53:GetHealthCheck"
```
wbeuil, line 147: Can we try to keep these permissions in order and put the two new route53 policies just next to the others. Thanks :)

go.sum
```diff
@@ -126,6 +126,7 @@ github.com/aws/aws-sdk-go v1.30.12/go.mod h1:5zCpMtNQVjRREroY7sYe8lOMRSxkhG6MZve
 github.com/aws/aws-sdk-go v1.31.9/go.mod h1:5zCpMtNQVjRREroY7sYe8lOMRSxkhG6MZveU8YkpAk0=
 github.com/aws/aws-sdk-go v1.34.2 h1:9vCknCdTAmmV4ht7lPuda7aJXzllXwEQyCMZKJHjBrM=
 github.com/aws/aws-sdk-go v1.34.2/go.mod h1:5zCpMtNQVjRREroY7sYe8lOMRSxkhG6MZveU8YkpAk0=
+github.com/aws/aws-sdk-go v1.37.8 h1:9kywcbuz6vQuTf+FD+U7FshafrHzmqUCjgAEiLuIJ8U=
```
wbeuil, line 129: Is this mandatory ?

pkg/remote/aws/route53_health_check_supplier_test.go
```diff
@@ -0,0 +1,93 @@
...
+				client.On("ListAllHealthChecks").Return([]*route53.HealthCheck{
+					{Id: aws.String("43c6ded9-1978-46cb-aa1f-e38a899076da")},
+					{Id: aws.String("8b679eaa-b93b-41b2-b406-dac956f2d20d")},
+				}, nil)
+			},
+			err: nil,
+		},
+		{
+			test:    "cannot list health check",
+			dirName: "route53_health_check_list_error",
```
wbeuil, line 55: Do we want to keep 2 empty results since the first test will have the same result ?

wbeuil (changes requested):

pkg/resource/aws/aws_route53_health_check_ext_test.go
```diff
@@ -0,0 +1,50 @@
...
+		},
+		{
+			name: "Tag name with fqdn and respath",
+			healthCheck: &AwsRoute53HealthCheck{
+				Id:           "id",
+				Tags:         map[string]string{"Name": "name"},
+				Fqdn:         aws.String("fq.dn"),
+				ResourcePath: aws.String("/toto"),
+			},
+			want: "name (fqdn: fq.dn path: /toto)",
```
eliecharra, line 28: I think it will be more consistent with other resources (and more readable) if we separate fields with coma here, like `name (fqdn: fq.dn, path: /toto)`

eliecharra (changes requested):
Maybe this is not the right place to rename this file

image

pkg/remote/aws/route53_health_check_supplier_test.go
eliecharra: We are testing error behaviour there it's not the same think that testing no results are returned from AWS side

wbeuil: From the product point of view it is indeed not the same, but as of the mocked golden file reader we return a slice of `[]resource.Resource` which here is an empty array.

I can easily see a future where we could want to test other errors and where with this logic the number of useless results can only increases exponentially.

eliecharra: I don't get your point, we need this test to test theses lines of code :

image

wbeuil: You could test this line like this:

```go
{
  test:    "cannot list health check",
  dirName: "route53_health_check_empty",
  mocks: func(client *mocks.Route53Repository) {
    client.On("ListAllHealthChecks").Return(nil, awserr.NewRequestFailure(nil, 403, ""))
  },
  err: remoteerror.NewResourceEnumerationError(awserr.NewRequestFailure(nil, 403, ""), resourceaws.AwsRoute53HealthCheckResourceType),
}
```

Look at the `dirName` here. I reused the one with empty results so that we don't need to create another empty results but with different naming. I did it like that in SQS.

eliecharra: Oh ok I got your point ! We can totally do this good idea 👍🏻

go.sum
moadibfr: hmm not sure I'll try without that

pkg/remote/aws/route53_health_check_supplier_test.go
moadibfr: yeah the only matter is that when we do that we cannot use -update to generate the folder so I used different folders in all the enumeration error test on all the other resources to keep -update behavior.

pkg/resource/aws/aws_route53_health_check_test.go
```diff
@@ -0,0 +1,93 @@
...
+				},
+				Check: func(result *acceptance.ScanResult, stdout string, err error) {
+					if err != nil {
+						t.Fatal(err)
+					}
+					result.AssertDriftCountTotal(0)
+					result.AssertDeletedCount(0)
+					result.AssertManagedCount(2)
+
+					for _, resource := range result.Analysis.Managed() {
```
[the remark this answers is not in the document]
moadibfr: to prevent crash if there is no managed resource. I was not sure the assertions above stopped the run. this is not so costly as there is 2 resourcea anyway...

pkg/remote/aws/route53_health_check_supplier_test.go
moadibfr: May I suggest that we stop at the error testing if it fail so we use no folder at all for error testing ? it's not really useful to test the list values and it's not mandatory that this return an empty list.

doc/cmd/scan/supported_resources/aws.md
moadibfr: Ok I'll group them by service

pkg/remote/aws/route53_health_check_supplier_test.go
moadibfr: ok that cannot work as we need the schema... I can use the same folder but it's not what I've done in other resource and I'm not a big fan a generating the golden file for a test and using them for another. We should find a way to have only one schema if we want to save space on golden files

moadibfr: ok it cannot work as we need the schema. I think we should find a way to use only one schema if we want to save space I'm not a really big fan of generating golden file for a test and using them for another...

pkg/remote/aws/test/route53_health_check_multiple/results.golden.json
```diff
@@ -0,0 +1 @@
+[]
```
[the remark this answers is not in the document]
moadibfr: :fearful:

pkg/resource/aws/aws_route53_health_check_test.go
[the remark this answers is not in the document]
eliecharra: not costly at all but difficult to understand why 😅 , you can safely the test will fail immediately if the assertion failed

pkg/remote/aws/route53_health_check_supplier_test.go
eliecharra: I'm good with that, we should find a way to reuse schema

pkg/resource/aws/aws_route53_health_check_test.go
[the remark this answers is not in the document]
eliecharra: image

pkg/remote/aws/route53_health_check_supplier_test.go
```diff
@@ -0,0 +1,93 @@
...
+
+	"github.com/aws/aws-sdk-go/service/route53"
+
+	"github.com/aws/aws-sdk-go/aws/awserr"
+	"github.com/cloudskiff/driftctl/mocks"
+	"github.com/cloudskiff/driftctl/pkg/parallel"
+	remoteerror "github.com/cloudskiff/driftctl/pkg/remote/error"
+	"github.com/cloudskiff/driftctl/test"
+	"github.com/cloudskiff/driftctl/test/goldenfile"
+	mocks2 "github.com/cloudskiff/driftctl/test/mocks"
```
eliecharra, line 17: `mocks2` 😉

eliecharra (changes requested):
:warning: You have removed `pkg/resource/aws/aws_security_group_rule_test.go`

pkg/remote/aws/route53_health_check_supplier_test.go
wbeuil: Ok then let's break it into 2 PRs, this one and mine on cloudfront will still have this kind of different folder with its own results and schema. Another PR will deal with a new architecture where we use only one schema per resource. Are you ok with that ?

pkg/remote/aws/route53_health_check_supplier_test.go
moadibfr: haha I don't understand why goland keep doing that -_-

pkg/remote/aws/route53_health_check_supplier_test.go
```diff
@@ -0,0 +1,93 @@
...
+				t.Fatal(err)
+			}
+			supplierLibrary.AddSupplier(NewSNSTopicSupplier(provider))
+		}
+
+		t.Run(c.test, func(tt *testing.T) {
+			fakeClient := mocks.Route53Repository{}
+			c.mocks(&fakeClient)
+			provider := testmocks.NewMockedGoldenTFProvider(c.dirName, providerLibrary.Provider(terraform.AWS), shouldUpdate)
+			SNSTopicDeserializer := awsdeserializer.NewRoute53HealthCheckDeserializer()
```
wbeuil, line 80: Can you just change here the naming of the variable

eliecharra (approved):

pkg/remote/aws/route53_health_check_supplier_test.go
moadibfr: for now i reused empty test folder like you asked, But yeah we should do a dedicated pr to refactor this

pkg/remote/aws/route53_health_check_supplier_test.go
moadibfr: :facepalm:

eliecharra (approved):

wbeuil (approved):

Issue #161, opened 2021-01-28
Some test use custom mock and should be rewritten with mockery


moadibfr:
**Description**
Some test use custom mock and should be rewritten with mockery

- [x] MockAwsClientFactory #266 
- [x] MockAwsS3Client #266 
- [x] MockAWSEC2Client #289
- [x] MockAWSRoute53Client #273
- [x] MockAWSRDSClient #296
- [x] MockAWSLambdaClient #294

PR #266, opened 2021-02-16
Remove custom mock and add repository for S3

eliecharra:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | no
| 🚀 New feature?   | no
| ⚠ Deprecations?   | no
| ❌ BC Break       | no
| 🔗 Related issues | #161 
| ❓ Documentation  | no

## Description

Refactor of S3 suppliers to use repository, also remove custom mock for S3


wbeuil (commented):
LGTM, just one question: why not having the 2 newly mocks in the root `mocks` folder ?

If we could avoid having all our repository mocks in different folders, that could be good.

Issue #249, opened 2021-02-12
Add support for github_branch_protection


eliecharra:

sjourdan:
:warning: 

it could detect a gh branch protection manually created on a tf managed repo 

```
  github_branch_protection:
    - Branch: require_pr_review (Id: MDIwOkJyYW5jaFByb3RlY3Rpb25SdWxlMTk2MjAwNjE=)
```

but a branch protection created from tf is detected as a drift while it's not

```
$ LOG_LEVEL=info driftctl-0.6 scan --from tfstate+s3://driftctl-tfstates/ --to github+tf
Terraform provider initialized (name=github, alias=sj-qa-1-org)
INFO[0000] Start reading IaC
INFO[0001] Start scanning cloud provider
Found unmanaged resources:
  github_membership:
    - sj-qa-1-org:sj-qa-1
  github_branch_protection:
    - Branch: main (Id: MDIwOkJyYW5jaFByb3RlY3Rpb25SdWxlMTk2MjAxMTg=)
Found 6 resource(s)
 - 66% coverage
 - 4 covered by IaC
 - 2 not covered by IaC
 - 0 deleted on cloud provider
 - 0/4 drifted from IaC
```
I confirm the branch protection is in the state 

```
$ terraform state show github_branch_protection_v3.yadda
# github_branch_protection_v3.yadda:
resource "github_branch_protection_v3" "yadda" {
    branch                 = "main"
    enforce_admins         = false
    etag                   = "W/\"548d28041960e5f347b50b83435109aafa53bb422ccc93bfc729ed95cc908e65\""
    id                     = "driftctl-demo-src-6qaidu:main"
    repository             = "driftctl-demo-src-6qaidu"
    require_signed_commits = false

    restrictions {
        users = [
            "sj-qa-1",
        ]
    }
}
```

sjourdan:
To reproduce, simplified:

A repo and a branch protection allowing a single user:

```hcl

resource "github_repository" "test-repo-1" {
  name        = "test-repo-1"
  description = "A GitHub Repo"
  visibility  = "public"
}

resource "github_branch_protection_v3" "yadda" {
  repository     = github_repository.test-repo-1.name
  branch         = "main"
  restrictions {
    users = ["your-username"]
  }
}
```

sjourdan:
okay @eliecharra showd me the docs and I mismatched v3 github_branch_protection :innocent: 

sooo ... LGTM! :tada: 

[...]

PR #301, opened 2021-02-26
Add github_branch_protection

eliecharra:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | no
| 🚀 New feature?   | yes
| ⚠ Deprecations?   | no
| ❌ BC Break       | no
| 🔗 Related issues | #249 
| ❓ Documentation  | yes

## Description

⚠️ To be able to alert when use user `github_branch_protection_v3`, we should create a resource, a deserializer and a finally a middleware to check if there is any `github_branch_protection_v3` resource in state just to send an alert and remove theses resources. Do we really want to do that ? Maybe a line in LIMITATIONS.md may be enough.

⚠️ It seems that the terraform provider is not returning the repository_id, so I've ignored it from the diff

Attentive review required please @moadibfr @wbeuil


pkg/remote/github/github_branch_protection_supplier.go
```diff
@@ -0,0 +1,59 @@
...
+		s.runner.Run(func() (cty.Value, error) {
+			completeResource, err := s.reader.ReadResource(terraform.ReadResourceArgs{
+				Ty: resourcegithub.GithubBranchProtectionResourceType,
+				ID: id,
+			})
+			if err != nil {
+				logrus.Warnf("Error reading %s[%s]: %+v", id, resourcegithub.GithubBranchProtectionResourceType, err)
+				return cty.NilVal, err
+			}
+			return *completeResource, nil
```
lotoussa, line 49: You could create a method for this like readProtection ?

pkg/resource/github/github _branch_protection_ext.go
```diff
@@ -0,0 +1,42 @@
...
+			decodedId, err := base64.StdEncoding.DecodeString(*r.RepositoryId)
+			if err == nil {
+				repoId = string(decodedId)
+			}
+		}
+
+		if repoId == "" {
+			return fmt.Sprintf("Branch: %s (Id: %s)", *r.Pattern, r.Id)
+		}
+		return fmt.Sprintf("Branch: %s, RepoId: %s", *r.Pattern, repoId)
```
lotoussa, line 23: Isn't missing parenthesis for RepoId ?

lotoussa (commented):
~LGTM

moadibfr:
> warning To be able to alert when use user github_branch_protection_v3, we should create a resource, a deserializer and a finally a middleware to check if there is any github_branch_protection_v3 resource in state just to send an alert and remove theses resources. Do we really want to do that ? Maybe a line in LIMITATIONS.md may be enough.

Maybe we could add a warning when we find unmanaged repository only ? Saying that this might be a false positive as we don't support v3 ?

wbeuil (changes requested):

doc/LIMITATIONS.md
```diff
@@ -18,3 +18,14 @@
...
+### Github
+- github_branch_protection_v3:
+  - This resource is not supported and will probably never be as it overlap with github_branch_protection.
+    `github_branch_protection` is more suitable for performance purpose.
+    We cannot support these two resources as we don't have any way to discriminate them by enumerating resources from
+    remote side. They represent the same notion but using REST vs GraphQL APIs. Driftctl team recommend you tu use the
+    newer `github_branch_protection`. Or at least ignore all your `github_branch_protection_v3` in driftignore.
+- github_branch_protection:
+  - Branch protection are not returned as unmanaged if the branch protection pattern does not match at least one branch
+  - We cannot show the repository for unmanaged resources as the terraform provider does not return the information
```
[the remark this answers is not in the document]
eliecharra: It wasn't clear when I read my own sentence so I changed to ` We cannot show the related repository name in driftctl output as the terraform provider does not retrieve this information.`

eliecharra:
> Maybe we could add a warning when we find unmanaged repository only ? Saying that this might be a false positive as we don't support v3 ?

We cannot make the difference between a false positive drift and a real drift with this approach. 
For me we the only way to handle this is to read v3 res from the state and try to match them to non-v3 unmanaged one and ignore them with a warning. So we'll end up with a v3 resource and deserializer without any supplier related.

WDYT @moadibfr @wbeuil ? Is it worth it ? I can totally do this in a couple of hours.

doc/LIMITATIONS.md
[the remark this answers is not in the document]
wbeuil: I'm fine with that !

wbeuil:
I'm not sure it's worth your time to do it. Are we sure that we can create a non-v3 resource with a v3 resource ? Can't we just warn the user if we find a v3 resource in its state ?

wbeuil (dismissed):

moadibfr (dismissed):

pkg/remote/github/github_branch_protection_supplier.go
eliecharra: This part of the code is never reused and it is readable enough, so for me there is no reason to create a new function.

pkg/resource/github/github _branch_protection_ext.go
eliecharra: This is a good question, WDYT @moadibfr @wbeuil ? Should we display `Branch: %s (RepoId: %s)` to be more consistent with `Branch: %s (Id: %s)` ?

wbeuil: Let's be consistent then. But I'm fine with one or the other

eliecharra: Fixed

wbeuil (approved):

Issue #250, opened 2021-02-12
Add support for aws_lambda_event_source_mapping


eliecharra:

sjourdan:
LGTM! 
Discovered an old bug with @moadibfr in the mean time: https://github.com/cloudskiff/driftctl/issues/311

```
$ LOG_LEVEL=info driftctl-0.6 scan --from tfstate+s3://driftctl-tfstates/
Terraform provider initialized (name=aws, alias=us-east-1)
INFO[0001] Start reading IaC
WARN[0002] error when reading iam policy attachment {{{{} map[groups:{{{} {{{} %!s(cty.primitiveTypeKind=83)}}}} id:{{{} %!s(cty.primitiveTypeKind=83)}} name:{{{} %!s(cty.primitiveTypeKind=83)}} policy_arn:{{{} %!s(cty.primitiveTypeKind=83)}} roles:{{{} {{{} %!s(cty.primitiveTypeKind=83)}}}} users:{{{} {{{} %!s(cty.primitiveTypeKind=83)}}}}] map[]}} map[groups:<nil> id:yaddattachment name:yaddattachment policy_arn:arn:aws:iam::141177182257:policy/policy roles:{map[%!s(int=3145829355):[driftctl-lambda-role]] {{{{} %!s(cty.primitiveTypeKind=83)}}}} users:<nil>]} : null value is not allowed
WARN[0002] Could not read from decoder for aws_iam_policy_attachment: null value is not allowed
INFO[0003] Start scanning cloud provider
Terraform provider initialized (name=aws, alias=eu-west-3)
Terraform provider initialized (name=aws, alias=eu-north-1)
Found unmanaged resources:
  aws_iam_policy_attachment:
    - driftctl-lambda-role-arn:aws:iam::141177182257:policy/policy
Found 7 resource(s)
 - 85% coverage
 - 6 covered by IaC
 - 1 not covered by IaC
 - 0 deleted on cloud provider
 - 0/6 drifted from IaC

```

PR #307, opened 2021-03-01
add support for aws_lambda_source_event_mapping

moadibfr:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | no
| 🚀 New feature?   | yes
| ⚠ Deprecations?   | no
| ❌ BC Break       | no
| 🔗 Related issues | #250 
| ❓ Documentation  | yes

## Description

add support for aws_lambda_source_event_mapping


pkg/iac/deserializers.go
```diff
@@ -58,8 +58,8 @@ func Deserializers() []deserializer.CTYDeserializer {
 		awsdeserializer.NewCloudfrontDistributionDeserializer(),
 		awsdeserializer.NewKMSKeyDeserializer(),
 		awsdeserializer.NewKMSAliasDeserializer(),
-
 		ghdeserializer.NewGithubRepositoryDeserializer(),
 		ghdeserializer.NewGithubTeamDeserializer(),
+		awsdeserializer.NewLambdaEventSourceMappingDeserializer(),
```
wbeuil, line 63: I would write it just below other aws deserializers

wbeuil (dismissed):
Just one small typo

pkg/remote/aws/lambda_event_source_mapping_supplier.go
```diff
@@ -0,0 +1,69 @@
...
+	}
+}
+
+func (s LambdaEventSourceMappingSupplier) Resources() ([]resource.Resource, error) {
+	functions, err := s.client.ListAllLambdaEventSourceMappings()
+	if err != nil {
+		return nil, remoteerror.NewResourceEnumerationError(err, resourceaws.AwsLambdaEventSourceMappingResourceType)
+	}
+	results := make([]cty.Value, 0)
+	if len(functions) > 0 {
```
eliecharra, line 41: useless if there

pkg/remote/aws/lambda_event_source_mapping_supplier.go
```diff
@@ -0,0 +1,69 @@
...
+		}
+	}
+	return s.deserializer.Deserialize(results)
+}
+
+func (s LambdaEventSourceMappingSupplier) readLambdaEventSourceMapping(sourceMappingConfig lambda.EventSourceMappingConfiguration) (cty.Value, error) {
+	resFunction, err := s.reader.ReadResource(
+		terraform.ReadResourceArgs{
+			Ty: resourceaws.AwsLambdaEventSourceMappingResourceType,
+			ID: aws.StringValue(sourceMappingConfig.UUID),
```
eliecharra, line 60: UUID can really be nil there ? I don't think `aws.StringValue` is very usefull there

pkg/remote/aws/lambda_event_source_mapping_supplier.go
```diff
@@ -0,0 +1,69 @@
+package aws
+
+import (
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/cloudskiff/driftctl/pkg/remote/aws/repository"
+	remoteerror "github.com/cloudskiff/driftctl/pkg/remote/error"
+
```
eliecharra, line 7: There is spaces between your imports (not only in this file)

pkg/resource/aws/aws_lambda_event_source_mapping.go
```diff
@@ -0,0 +1,38 @@
...
+	LastModified                   *string `cty:"last_modified" computed:"true" diff:"-"`
+	LastProcessingResult           *string `cty:"last_processing_result" computed:"true" diff:"-"`
+	MaximumBatchingWindowInSeconds *int    `cty:"maximum_batching_window_in_seconds"`
+	MaximumRecordAgeInSeconds      *int    `cty:"maximum_record_age_in_seconds" computed:"true"`
+	MaximumRetryAttempts           *int    `cty:"maximum_retry_attempts" computed:"true"`
+	ParallelizationFactor          *int    `cty:"parallelization_factor" computed:"true"`
+	StartingPosition               *string `cty:"starting_position" diff:"-"`
+	StartingPositionTimestamp      *string `cty:"starting_position_timestamp" diff:"-"`
+	State                          *string `cty:"state" computed:"true" diff:"-"`
+	StateTransitionReason          *string `cty:"state_transition_reason" computed:"true"`
```
eliecharra, line 23: I think that we could ignore this field too if we ignore the state ?

eliecharra (changes requested):

pkg/remote/aws/lambda_event_source_mapping_supplier.go
moadibfr: I shouldn't but we have seen strange errors so I think it's a good idea to not crash ?

pkg/remote/aws/lambda_event_source_mapping_supplier.go
moadibfr: yeah it's go import way or organizing import ?

pkg/resource/aws/aws_lambda_event_source_mapping.go
moadibfr: yeah you're right I forgot to add it there

pkg/remote/aws/lambda_event_source_mapping_supplier.go
eliecharra: If we want to handle this strange case we should not call `ReadResource` with a nil value IMHO, we should handle this case explicitly before in the loop, and maybe send a `log.Warning` as something went wrong

pkg/iac/deserializers.go
moadibfr: ok but the list is already unordered ^^

pkg/remote/aws/lambda_event_source_mapping_supplier.go
moadibfr: Hmm maybe it's goland I don't really know I removed them

eliecharra: I don't know what is the standard way to organize imports, but we should follow it and especially we should all use the same logic.

By default it seems that Goland split stdlib pkg (like `fmt`) from others and is using gofmt.

image

When using `go fmt -local` it groups imports by pkg and add whitespaces between them.

I also have goimport as file watcher in goland and the same gofmt logic seems to be applied.

I don't really care which option we take but we should avoid unnecessary conflicts in imports.

moadibfr: yeah I didn't see that go fmt had an other way of organizing, I removed the newline but I have no preference on what to use

eliecharra (approved):

Issue #232, opened 2021-02-09
.driftignore could allow a wildcard within a ResourceId after a prefix


psedge:
**Description**

Right now there are two cases covered by `.driftignore`:

1. To ignore an explicit `res.TerraformType`.`res.TerraformId` pair
2. To ignore any ID with type `res.TerraformType` using a wildcard '*'

I have a use-case where it would be useful to be able to ignore a set of `res.TerraformId`s which match a simple prefix pattern, for resources like: `aws_iam_policy_attachment.AWSServiceRoleFor{X}`. I believe that I can probably achieve the same with Filters, which I'll try next - but it would be really nice to keep rules like this within the `.driftignore` file for ease of use.

**Example**

The following `.driftignore` should ignore resources like ` AmazonSSMRoleForInstances` and. `aws_iam_role.AWSServiceRoleForAmazonSSM`:

```.driftignore
aws_iam_role.Amazon*
aws_iam_role.AWSServiceRole*
```

eliecharra:
> **Description**
> 
> Right now there are two cases covered by `.driftignore`:
> 
> 1. To ignore an explicit `res.TerraformType`.`res.TerraformId` pair
> 2. To ignore any ID with type `res.TerraformType` using a wildcard '*'

Juste to complete initial description 😉 

3. : To ignore a specific field of a resource `res.TerraformType.Tags` for example

Thanks for reporting this to us @psedge, this issue is pretty easy to fix, for some people interested in submitting a pull-request, you have to update how we match wildcards in `driftignore.go` `IsResourceIgnored`

```go
_, isExclusionRule := r.resExclusionList[fmt.Sprintf("%s.%s", res.TerraformType(), res.TerraformId())]
```

sjourdan:
LGTM! 

`aws_iam_policy_attachment.*azonSQ*`  in `.driftignore`  definitely ignored the following:

```
  aws_iam_policy_attachment:
    - labs-arn:aws:iam::aws:policy/AmazonSQSFullAccess
```

PR #314, opened 2021-03-04
DriftIgnore implement new wildcard match checker

lotoussa:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | no
| 🚀 New feature?   | yes
| ⚠ Deprecations?   | no
| ❌ BC Break       | no
| 🔗 Related issues | #232 
| ❓ Documentation  | yes -> wildcard examples for .driftignore

## Description

I decide to check the resources of the driftignore to be ignored by changing the way they are compared to the resources of the state or the provider.
To do this I add a comparison algorithm taking into account any wildcard.
To drastically reduce the complexity of this algorithm I have added a suppression of the wildcards that follow each other in a type or id of resource.

I modified the structure of DriftIgnore to avoid adding complexity to each resource check loop on the driftignore file.
It separates the exclusions with and without wildcard.

I'm not sure this pr is done.
Feel free to review.


eliecharra (changes requested):
Before I start a review, can you please add test to `TestDriftIgnore_IsResourceIgnored` with a new case dedicated to wildcard (add a new folder under testdata)

pkg/filter/driftignore.go
```diff
@@ -127,11 +141,33 @@ RuleCheck:
...
+	}
+	if str != "" && pattern != "" && str[0] == pattern[0] {
+		return wildcardMatchChecker(str[1:], pattern[1:])
+	}
+	return false
+}
+
 /**
  * Read a line of ignore
+ * Handle multiples asterisk escaping
```
wbeuil, line 163: `multiple`, this is invariable, it will never be plural.

pkg/filter/driftignore.go
```diff
@@ -11,14 +11,16 @@ import (
 )
 
 type DriftIgnore struct {
-	resExclusionList   map[string]struct{} // map[type.id] exists to ignore
-	driftExclusionList map[string][]string // map[type.id] contains path for drift to ignore
+	resExclusionList         map[string]struct{} // map[type.id] exists to ignore
+	resExclusionWildcardList map[string]struct{} // map[type.id] exists with wildcard to ignore
```
wbeuil, line 15: It works well but I wonder if we should have a list of all wildcards patterns. What if we want to handle another glob pattern like `{a,b,c}` ?

wbeuil (changes requested):
Well done, I tested it and wrote a small chunks of test and it worked perfectly as intended. My only fear is the way we keep a reference of all wildcard patterns. We can discuss it here but can probably amended it in another PR.

pkg/filter/driftignore.go
lotoussa: Great comment, i'll be glad to work on this in another pr.

My opinion on the subject would be to refactor the whole thing a bit to predict this behaviour, or else to assume a passage in strMatchChecker for each resource/exclusion (since the function also acts as a strcmp if there is no wildcard).

Simply, calls with string copy at each recursion (pointer arithmetic unavailable) seemed expensive and avoidable with a first check on a list containing only strings, which is also possible to continue to fill with a layer in addition to checking in readDriftIgnoreLine if a glob is present in the line buffer.

moadibfr: we could also use the same list and adapt the check when/if we add new patterns. When we discussed this it looked like a good idea to preprocess like the other list but it might need to change if the usage change later

wbeuil (approved):
LGTM

Issue #274, opened 2021-02-18
Show the user which .driftignore line is invalid


sjourdan:
**Description**

Simple UX in warnings

Today driftctl only displays which resource is concerned by the error, like: 

`WARN[0017] unable to parse line, invalid length, got 1 expected >= 2 line=github_repository.`

It would be easier for the user if it read sth like: 

`WARN[0017] unable to parse line, invalid length, got 1 expected >= 2 line=63`
**Example**

1. add an invalid line to your .driftignore
2. run driftctl

eliecharra:
> Today driftctl only displays which resource is concerned by the error, like:

We display the full line which can't be parsed, we can add the line number

PR #315, opened 2021-03-05
Add line number in log when invalid driftignore line encountered

lotoussa:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | no
| 🚀 New feature?   | yes
| ⚠ Deprecations?   | no
| ❌ BC Break       | no
| 🔗 Related issues | #274 
| ❓ Documentation  | no

## Description

Just added line index in logger for invalid field in .driftignore, but i don't understand why it print content first and line after, since i set line then content in logrus.Fields{}
Can avoid itoa conversion if index created outer loop scope.

### DIFF
--- a/pkg/filter/driftignore.go
+++ b/pkg/filter/driftignore.go
@@ -4,6 +4,7 @@ import (
 	"bufio"
 	"fmt"
 	"os"
+	"strconv"
 	"strings"
 
 	"github.com/cloudskiff/driftctl/pkg/resource"
@@ -35,7 +36,7 @@ func (r *DriftIgnore) readIgnoreFile() error {
 	defer file.Close()
 
 	scanner := bufio.NewScanner(file)
-	for scanner.Scan() {
+	for lineNumber := 1; scanner.Scan(); lineNumber++ {
 		line := scanner.Text()
 		if line == "" || strings.HasPrefix(line, "#") {
 			logrus.WithFields(logrus.Fields{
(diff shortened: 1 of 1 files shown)


pkg/filter/driftignore.go
```diff
@@ -47,7 +48,8 @@ func (r *DriftIgnore) readIgnoreFile() error {
 		nbArgs := len(typeVal)
 		if nbArgs < 2 {
 			logrus.WithFields(logrus.Fields{
-				"line": line,
+				"line":    strconv.Itoa(index + 1),
```
eliecharra, line 51: You can start your index at 1 instead of 0 `index := 1`, and thus there you can remove the index+1. And you can rename index to `lineNumber` it will add semantic sense to this variable.

Maybe it will work without the `Itoa` conversion, can you try ?

eliecharra (changes requested):

pkg/filter/driftignore.go
lotoussa: I doesn't work without the itoa cause the range give a loop object and print (&int=0)
The lineNumber = 1 seems cool, i was just thinking, if reusing the index variable in the loop so there is a need to start at 0.

eliecharra: > I doesn't work without the itoa cause the range give a loop object and print (&int=0)

I don't know what do you mean by a loop object, but ok if logrus print int values like this `Itoa` makes sense `&int=0`

eliecharra (approved):

Issue #257, opened 2021-02-15
Display a progress message when scanning resources


eliecharra:
**Description**
Scanning resources could be a long process, we want to display something to user to show that something is running and the process is still running.

**Example**
```
Found 25 resources ...
Found 30 resources ...
[...]
```

williamcodes:
Good lord, is this why it seems like it's just hanging indefinitely when you run a scan? I thought it was broken! I've been restarting it and tweaking the flags! Please, yes, absolutely, put some output on screen and save people all this wasted time.

mwarkentin:
This would be great, when running driftctl against one of our smaller sandbox environments it ran and returned the output very quickly. When I tried it against one of our larger accounts, it hung for an hour until I killed it and I couldn't tell if it was working or where it was getting hung up. The larger account also had dozens of terraform states, so I was also trying to figure out if I'd passed in the flags correctly which made it harder to tell if it was working correctly.

sjourdan:
LGTM! 

- the counter is incremented as resources are found 
- the ascii spinner is moving as it's running and discovering resources 
 
```shell 
$ driftctl scan --from tfstate+s3://driftctl-tfstates/
Scanning resources: ⢿ (48)
[...]
```

jacalata:
I just installed the most recent windows version and I don't get any response when I run driftctl. Presumably this change is in there - are there any debug logs being created as well?

eliecharra:
> I just installed the most recent windows version and I don't get any response when I run driftctl. Presumably this change is in there - are there any debug logs being created as well?

Yeah you can try to run driftctl with env var `LOG_LEVEL=debug`

jacalata:
Hm, I'm not seeing any output still, although I can see about 70 processes/threads in procexp so it's probably doing something. I've tried running with and without --from.

jacalata:
it does give this output when I kill it (here it ran for almost 20 minutes, then showed that WARN on ctrl-c)
12:34:13.10|D:\dev\webhooks\webhooks-delivery-lambda\terraform>driftctl scan --from tfstate+s3://webhooks-delivery-development.terraform-remote-state.tabint.net
[1;33mWARN[0m[1107][97m Detected interrupt during terraform provider configuration, cleanup ...[0m
12:53:33.24|D:\dev\webhooks\webhooks-delivery-lambda\terraform>

eliecharra:
> Hm, I'm not seeing any output still

[...]

PR #323, opened 2021-03-08
Show progress while reading resources from provider

moadibfr:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | no
| 🚀 New feature?   | yes
| ⚠ Deprecations?   | no
| ❌ BC Break       | no
| 🔗 Related issues | #257 
| ❓ Documentation  | no

## Description

Add progress service that display a spinner until stopped or timeouted


pkg/cmd/scan.go
```diff
@@ -85,6 +89,12 @@ func NewScanCmd() *cobra.Command {
 	}
 
 	fl := cmd.Flags()
+	fl.BoolP(
+		"quiet",
+		"",
+		false,
+		"Do not display anything outside the scan results",
```
eliecharra, line 96: Don't display anything but scan results

pkg/cmd/scan/output/output_test.go
```diff
@@ -304,7 +304,7 @@ func TestGetPrinter(t *testing.T) {
 				Options: map[string]string{
 					"path": tt.path,
 				},
-			}); !reflect.DeepEqual(got, tt.want) {
+			}, false); !reflect.DeepEqual(got, tt.want) {
```
eliecharra, line 307: I think it should be a part of the test case struct and we should add a case with `true`

pkg/output/progress.go
```diff
@@ -0,0 +1,98 @@
...
+
+	"go.uber.org/atomic"
+
+	"github.com/sirupsen/logrus"
+)
+
+var spinner = []string{"⣾", "⣽", "⣻", "⢿", "⡿", "⣟", "⣯", "⣷"}
+
+const (
+	timeout      = 10 * time.Second
```
eliecharra, line 14: When reading the test file I was wondering where does `timeout` come from, as it is exposed in the package maybe we can rename it to `progressTimeout` ?

`timeout => progressTimeout`
`displaySpeed => progressRefreshRate`

eliecharra (changes requested):
Looks awesome ! I only have some small comments

eliecharra:
@moadibfr We need a doc PR for the quiet flag too 😉

eliecharra (approved):

PR #339, opened 2021-03-16
fix race condition on progress that could cause deadlock

moadibfr:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | yes
| 🚀 New feature?   | no
| ⚠ Deprecations?   | no
| ❌ BC Break       | no
| 🔗 Related issues | 
| ❓ Documentation  | no 

## Description
fix race condition on progress that could cause deadlock

### DIFF
--- a/pkg/output/progress_test.go
+++ b/pkg/output/progress_test.go
@@ -7,13 +7,24 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
-func TestProgressTimeout(t *testing.T) {
+func TestProgressTimeoutDoesNotInc(t *testing.T) {
 	progress := NewProgress()
 	progress.Start()
-	time.Sleep(progressTimeout + 1)
-	progress.Inc()  // should not hang
+	progress.Stop() // should not hang
+	progress.Inc()  // should not hang or inc
+	assert.Equal(t, uint64(0), progress.Val())
+}
+
+func TestProgressTimeoutDoesNotHang(t *testing.T) {
+	progress := NewProgress()
+	progress.Start()
+	time.Sleep(progressTimeout)
+	for progress.started.Load() == true {
+	}
+	progress.Inc()  // should not hang or inc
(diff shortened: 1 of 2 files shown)


pkg/output/progress_test.go
```diff
@@ -7,13 +7,20 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
-func TestProgressTimeout(t *testing.T) {
+func TestProgressTimeoutDoesNotInc(t *testing.T) {
+	progress := NewProgress()
+	progress.Start()
+	progress.Stop() // should not hang
+	progress.Inc()  // should not hang
+	assert.Equal(t, progress.Val(), uint64(0))
```
sundowndev, line 15: You mixed expected and actual value here

```diff
-assert.Equal(t, progress.Val(), uint64(0))
+assert.Equal(t, uint64(0), progress.Val())
```

pkg/output/progress_test.go
```diff
@@ -7,13 +7,20 @@ import (
...
+}
+
+func TestProgressTimeoutDoesNotHang(t *testing.T) {
 	progress := NewProgress()
 	progress.Start()
 	time.Sleep(progressTimeout + 1)
 	progress.Inc()  // should not hang
 	progress.Stop() // should not hang
-	assert.Equal(t, uint64(0), progress.Val())
```
sundowndev, line 16: Isn't it relevant to assert this value?

pkg/output/progress_test.go
moadibfr: :+1: fixed

pkg/output/progress_test.go
moadibfr: not realy but I changed the test to be sure the timeout was reached so I could put this test back

sundowndev:
LGTM

sundowndev (approved):
LGTM

PR #344, opened 2021-03-17
Remove provider initialization message

### LINKED ISSUE #337: Remove provider initialization message
image

sundowndev:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | yes
| 🚀 New feature?   | no
| ⚠ Deprecations?   | no
| ❌ BC Break       | no
| 🔗 Related issues | #337
| ❓ Documentation  | no 

## Description

When launching a scan, the tool was logging the message "Terraform provider initialized" which means the tool has gained access to the provider's API and is scanning the account, looking for resources in the actual infrastructure. So it can be confusing to the user if you have resources in several regions.

### DIFF
--- a/pkg/remote/terraform/provider.go
+++ b/pkg/remote/terraform/provider.go
@@ -128,11 +128,11 @@ func (p *TerraformProvider) configure(alias string) error {
 		"alias": alias,
 	}).Debug("New gRPC client started")
 
-	output.Printf("Terraform provider initialized (name=%s", p.Config.Name)
+	logrus.Debugf("Terraform provider initialized (name=%s", p.Config.Name)
 	if alias != "" {
-		output.Printf(", alias=%s", alias)
+		logrus.Debugf(", alias=%s", alias)
 	}
-	output.Printf(")\n")
+	logrus.Debugf(")\n")
 
 	return nil
 }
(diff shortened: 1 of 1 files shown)


pkg/remote/terraform/provider.go
```diff
@@ -128,11 +128,11 @@ func (p *TerraformProvider) configure(alias string) error {
 		"alias": alias,
 	}).Debug("New gRPC client started")
 
-	output.Printf("Terraform provider initialized (name=%s", p.Config.Name)
+	logrus.Debugf("Terraform provider initialized (name=%s", p.Config.Name)
```
eliecharra, line 131: You use `logrus.WithFields` instead of adding context values directly in the string. It's a better way to make our logs more structured and readable.

eliecharra (changes requested):

pkg/remote/terraform/provider.go
sundowndev: You're right, the previous code was buggy anyway.

eliecharra (approved):
Great 👍🏻

wbeuil:
@all-contributors please add @sundowndev for code.

Issue #338, opened 2021-03-16
S3 bucket should be filtered on current region


eliecharra:
**Description**
Actually when retrieving S3 bucket, we retrieve bucket for every AWS region. To be consistent with other resources, we should only scan bucket for the current region.

In the screenshot below, I run driftctl on us-east-1 and I dont want to see the unmanaged bucket from `eu-north-1` in my output

![image](https://user-images.githubusercontent.com/6154987/111335031-56788780-8674-11eb-8d06-e0e009e7068b.png)
![image](https://user-images.githubusercontent.com/6154987/111335043-59737800-8674-11eb-9a79-83ee9498be25.png)

eliecharra:
@all-contributors please add @vainkop for bug.

sjourdan:
LGTM, I confirm an S3 bucket from `eu-north-1` didn't show during a scan with `AWS_REGION` set to `us-east-1`.

vainkop:
@sjourdan awesome, thanks!

PR #348, opened 2021-03-17
Filter out s3 buckets from different regions

sundowndev:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | yes
| 🚀 New feature?   | no
| ⚠ Deprecations?   | no
| ❌ BC Break       | no
| 🔗 Related issues | #338
| ❓ Documentation  | no 

## Description

>  Actually when retrieving S3 bucket, we retrieve bucket for every AWS region. To be consistent with other resources, we should only scan bucket for the current region.


pkg/remote/aws/s3_bucket_supplier.go
```diff
@@ -2,14 +2,15 @@ package aws
 
 import (
 	"github.com/aws/aws-sdk-go/service/s3"
+	"github.com/zclconf/go-cty/cty"
```
eliecharra, line 5: Can you revert theses ordering chunk ?

pkg/remote/terraform/provider.go
```diff
@@ -157,6 +157,10 @@ func (p *TerraformProvider) ReadResource(args tf.ReadResourceArgs) (*cty.Value,
 		delete(args.Attributes, "alias")
 	}
 
+	if alias != p.Config.DefaultAlias {
```
eliecharra, line 160: This is not the right place to do this, that's not what we want it will break the behavior of the grpcProvider context.

The work should be done in the S3 bucket supplier.

eliecharra (changes requested):

pkg/remote/terraform/provider.go
wbeuil: We didn't understand each other when I tried to explain it on Discord then.

pkg/remote/terraform/provider.go
```diff
@@ -217,3 +221,7 @@ func (p *TerraformProvider) Cleanup() {
 		client.Close()
 	}
 }
+
+func (p *TerraformProvider) GetConfig() TerraformProviderConfig {
+	return p.Config
```
lotoussa, line 226: Why don't you use this getter but created it ?

sundowndev: I created it to implement it in the ResourceReader interface but it's no longer needed

lotoussa (commented):
LGTM

pkg/remote/aws/s3_bucket_supplier.go
sundowndev: These changes are from gofmt, shouldn't we follow these rules? My IDE does that by default but I guess I can disable that

eliecharra (approved):
👍🏻

test/mocks/MockGoldenTerraformProvider.go
```diff
@@ -8,10 +8,11 @@ import (
 
 	"github.com/cloudskiff/driftctl/test/goldenfile"
 
-	"github.com/cloudskiff/driftctl/pkg/terraform"
 	"github.com/hashicorp/terraform/providers"
 	"github.com/zclconf/go-cty/cty"
 	ctyjson "github.com/zclconf/go-cty/cty/json"
+
+	"github.com/cloudskiff/driftctl/pkg/terraform"
```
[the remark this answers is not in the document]
sundowndev: I think you're viewing an outdated diff, I already reverted those changes

Issue #321, opened 2021-03-05
Sort resource names in unmanaged resources list


adamdonahue:
**Description**
Sort resource names in unmanaged resources list.  They appear to be unsorted at present, or sorted in search order?

**Example**
Here's what it looks like now, for example:

```
Found unmanaged resources:
  aws_ecr_repository:
    - s-dash/sample-3
    - s-dash/sample-data
    - s-dash/sample-1
    - s-dash/post-migration-app
    - plotly/reloader
    - s-dash/stag01
```
Perhaps this is only relevant to `aws_ecr_repository` objects.

eliecharra:
Hi @adamdonahue thanks for your report and your interest in driftctl 🙏🏻 

As we run scan in parallel, the resources ordering is unpredictable. 

We may introduce a default sorting method based on resources ID's, and in some case maybe this logic needs to be overridden to use another field than ID for sorting, for example the same as you said.

To be clear, here you only expect a reproductible order or a alpha-numeric sort ? What's the most relevant for you ?

adamdonahue:
I'm not talking about the order in which resources are processed.  Just the output.  Basic map/reduce pattern, essentially.  This is a really minor request though.

adamdonahue:
I didn't dig into the code too deeply but if this would impact performance (if, for example, you streaming and flushing output as you go), then I'd argue it's very low priority.

adamdonahue:
I noticed I didn't answer your question!  Just an alphanumeric sort -- makes it easier to scan and fine objects that need to be reconciled, but again, low priority.

sundowndev:
Closing as it has been fixed in v0.8.0.

PR #356, opened 2021-03-19
Sort unmanaged & deleted resources in a predictable order

sundowndev:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | yes
| 🚀 New feature?   | no
| ⚠ Deprecations?   | no
| ❌ BC Break       | no
| 🔗 Related issues | #321
| ❓ Documentation  | no 

## Description

For now, resources were fetched in parallel which would cause the resource list order to be unpredictable. With these changes, the resources are always sorted by their Terraform Id.


lotoussa (approved):
lgtm

pkg/analyser/analyzer.go
```diff
@@ -6,9 +6,10 @@ import (
 
 	resourceaws "github.com/cloudskiff/driftctl/pkg/resource/aws"
 
+	"github.com/r3labs/diff/v2"
```
moadibfr, line 9: pretty sure you have something different to handle import sorting we need to check that together

pkg/analyser/analyzer_test.go
```diff
@@ -940,23 +990,31 @@ func TestAnalyze(t *testing.T) {
...
 			}
 
-			unmanagedChanges, err := diff.Diff(result.Unmanaged(), c.expected.Unmanaged())
-			if err != nil {
-				t.Fatalf("Unable to compare %+v", err)
-			}
-			if len(unmanagedChanges) > 0 {
-				for _, change := range unmanagedChanges {
-					t.Errorf("%+v", change)
+			for i, expected := range c.expected.Unmanaged() {
```
moadibfr, line 993: diff have an ooption to check slice ordering: 
>SliceOrdering ensures that the ordering of items in a slice is taken into account
https://github.com/r3labs/diff

pkg/resource/resource.go
```diff
@@ -48,3 +49,8 @@ type NormalizedResource interface {
 func IsSameResource(rRs, lRs Resource) bool {
 	return rRs.TerraformType() == lRs.TerraformType() && rRs.TerraformId() == lRs.TerraformId()
 }
+
+func Sort(res []Resource) []Resource {
+	sort.SliceStable(res, func(i, j int) bool { return res[i].TerraformId() < res[j].TerraformId() })
```
moadibfr, line 54: might want to take type into account to have a more readable output

moadibfr (changes requested):

pkg/analyser/analyzer_test.go
sundowndev: Yes indeed! thanks

moadibfr (approved):
LGTM

wbeuil (approved):
I'm fine with that

eliecharra (approved):

Issue #354, opened 2021-03-19
use 'scan' command by default


sjourdan:
**Description**
`driftctl scan` is the basic and almost only command everybody uses.
Maybe we could use `scan` as a default command when `driftctl` is called?

**Example**

```bash
$ driftctl --from tfstate+s3://yaddi/yadda
[...]
```

adamdonahue:
I'm not sure I necessarily agree with this.  I think it's cleaner to specify `scan`, even though yes, it's one additional word.  It's also consistent with how a lot of other tools work.  There's no default action in `kubectl`, or `terraform`, etc.

sundowndev:
> I'm not sure I necessarily agree with this. I think it's cleaner to specify `scan`, even though yes, it's one additional word. It's also consistent with how a lot of other tools work. There's no default action in `kubectl`, or `terraform`, etc.

Just to clarify, the point of this is about having a short-cut for the scan command. The scan command will not be deleted anyway. So this shouldn't change much on the user experience. It's difficult to compare with the provided examples because kubectl and Terraform are two programs with a lot of features with a lot of commands. There's no way to tell what `terraform` without any argument would do; nor do kubectl. Driftctl was designed for one goal: scan the infrastructure for drifts. As we could say about curl: it was designed to simply request an HTTP endpoint; and curl run the request with no sub command. Hope it helps better understand this issue :-)

adamdonahue:
I understand your reasoning.  I think it's just interesting that scan is the default, which seems to break symmetry with other subcommands.  I'm trying to think of another subcommand-based option that has a default like that, which I'm sure there must be some.

If I run `driftctl` without options, will I get the help screen still?  I think that is better for usability.  

For example, if this is the default output now, not sure that's optimal:

```
(sre-driftctl-reporter) adonahue@devadmin:~/workspace/sre/management/infrastructure/groups/midas/infra$ driftctl scan
Terraform provider initialized (name=aws, alias=us-east-1)
Unable to scan resources: lstat terraform.tfstate: no such file or directory
```

Seems odd, especially if you're just seeing out for the first time (again, about usability).  I realize `-h` is the usual next step.  

Nevertheless, it's a compromise, as you said.   Just thought I'd provide some feedback as a devil's advocate.

adamdonahue:
[...]

PR #357, opened 2021-03-22
Use the "scan" command by default

sundowndev:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | no
| 🚀 New feature?   | yes
| ⚠ Deprecations?   | no
| ❌ BC Break       | yes
| 🔗 Related issues | #354
| ❓ Documentation  | yes 

## Description

Since running a scan is the most common behavior for driftctl, it should be the default behavior when running the tool without a specific command.

### DIFF
--- a/pkg/cmd/driftctl.go
+++ b/pkg/cmd/driftctl.go
@@ -5,6 +5,7 @@ import (
 	"strings"
 
 	"github.com/cloudskiff/driftctl/build"
+	"github.com/cloudskiff/driftctl/pkg/iac/terraform/state/backend"
 	"github.com/cloudskiff/driftctl/sentry"
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
@@ -35,6 +36,8 @@ type DriftctlCmd struct {
 }
 
 func NewDriftctlCmd(build build.BuildInterface) *DriftctlCmd {
+	scanCmd := NewScanCmd()
+
 	cmd := &DriftctlCmd{
 		cobra.Command{
 			Use:   "driftctl <command> [flags]",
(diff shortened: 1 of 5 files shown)


lotoussa (commented):
cool work, lgtm

pkg/cmd/driftctl.go
```diff
@@ -49,10 +52,14 @@ func NewDriftctlCmd(build build.BuildInterface) *DriftctlCmd {
 			Long:          "Detect, track and alert on infrastructure drift.",
 			SilenceErrors: true,
 			SilenceUsage:  true,
+			PreRunE:       scanCmd.PreRunE,
+			RunE:          scanCmd.RunE,
 		},
 		build,
 	}
 
+	RegisterScanFlags(&cmd.Command, &ScanOptions{BackendOptions: &backend.Options{}})
```
eliecharra, line 61: It will bind flags to the root cmd, I don't think we want to do that

image

It should stay like this for me

```
Detect, track and alert on infrastructure drift.

Usage: driftctl <command> [flags]

COMMANDS:
  completion               Generate completion script
  scan                     Scan
  version                  Display driftctl version

FLAGS:
      --error-reporting    Enable error reporting.
                           WARNING: may leak sensitive data
  -h, --help               Display help for command
      --no-version-check   Disable the version check

LEARN MORE:
  Use "driftctl <command> --help" for more information about a command

```

WDYT @cloudskiff/engineering

sundowndev: I think there's a misunderstanding of what this PR is about. The goal is to be able to run the scan from the root command, which implies to add the necessary flags to it. May be you just don't want to override the root command at all? I don't personally see this as a necessary change, most people will understand the help message and use the scan command since as it's stated in the docs. But may be it could improve a bit the user experience at no cost.

Issue #367, opened 2021-03-26
Add details to error reporting flag in help message


eliecharra:
**Description**
The error reporting flag seems to be confusing.
We need to make it clear that is sends some data to out Sentry server.

sjourdan:
Can we have a quick grammar fix? 

from 

```
      --send-crash-report   Enable error reporting. Crash data will be sent to us via Sentry.
                            WARNING: may leak sensitive data (please read documentation for more details)
                            This flag should be used only if an error occur during execution
```

to 

```
      --send-crash-report   Enable error reporting. Crash data will be sent to us via Sentry.
                            WARNING: may leak sensitive data (please read the documentation for more details)
                            This flag should be used only if an error occurred during execution
```

sjourdan:
LGTM (using present tense as discussed on discord)

```
      --send-crash-report   Enable error reporting. Crash data will be sent to us via Sentry.
                            WARNING: may leak sensitive data (please read the documentation for more details)
                            This flag should be used only if an error occurs during execution
```

PR #381, opened 2021-03-26
Add details to error reporting flag in help message

lotoussa:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | no
| 🚀 New feature?   | no
| ⚠ Deprecations?   | no
| ❌ BC Break       | **yes**
| 🔗 Related issues | #367 
| ❓ Documentation  | no

## Description

The error reporting flag seems to be confusing.
We need to make it clear that is sends some data to out Sentry server.

So i add an information i found on driftctl-docs error-reporting part to the helper message of the program.

### DIFF
--- a/pkg/cmd/driftctl.go
+++ b/pkg/cmd/driftctl.go
@@ -62,7 +62,7 @@ func NewDriftctlCmd(build build.BuildInterface) *DriftctlCmd {
 
 	cmd.PersistentFlags().BoolP("help", "h", false, "Display help for command")
 	cmd.PersistentFlags().BoolP("no-version-check", "", false, "Disable the version check")
-	cmd.PersistentFlags().BoolP("error-reporting", "", false, "Enable error reporting.\nWARNING: may leak sensitive data")
+	cmd.PersistentFlags().BoolP("send-crash-report", "", false, "Enable error reporting. Crash data will be sent to us via Sentry.\nWARNING: may leak sensitive data (please read documentation for more details)\nThis flag should be used only if an error occur during execution")
 
 	cmd.AddCommand(NewScanCmd())
 


pkg/cmd/driftctl.go
```diff
@@ -62,7 +62,7 @@ func NewDriftctlCmd(build build.BuildInterface) *DriftctlCmd {
 
 	cmd.PersistentFlags().BoolP("help", "h", false, "Display help for command")
 	cmd.PersistentFlags().BoolP("no-version-check", "", false, "Disable the version check")
-	cmd.PersistentFlags().BoolP("error-reporting", "", false, "Enable error reporting.\nWARNING: may leak sensitive data")
+	cmd.PersistentFlags().BoolP("error-reporting", "", false, "Enable error reporting. Crash data will be sent to us via Sentry.\nWARNING: may leak sensitive data")
```
sundowndev, line 65: 👍🏽 

May be we can tell the user to read the docs to find details about what is collected ?

Example:  `WARNING: may leak sensitive data (please read documentation for more details)`

WDYT @cloudskiff/engineering ?

moadibfr: Yeah I think it would be a good idea to tell users the docs give all information that will be sent to us.

I wonder if we should rename the flag also ? Some user seems to be expecting to see an error reporting but this is more like sending us more detail about a crash. (this only work if there is an error if I'm right)

Last point we talked about having a clear consent for this flag. Like a pause waiting for the user to type 'yes' or something like that ?

eliecharra: > Yeah I think it would be a good idea to tell users the docs give all information that will be sent to us.

Definitively a must have. We should also say that this flag should be used only if an error occur during execution.

> I wonder if we should rename the flag also ? Some user seems to be expecting to see an error reporting but this is more like sending us more detail about a crash. (this only work if there is an error if I'm right)

I agree, the flag should be renamed.

> Last point we talked about having a clear consent for this flag. Like a pause waiting for the user to type 'yes' or something like that ?

For now let's see if the documentation and flag rename is enough, maybe we can add this in another PR later.

eliecharra (changes requested):

pkg/cmd/driftctl.go
sundowndev: > I agree, the flag should be renamed.

How about `--send-crash-report` ?

eliecharra: Sounds great 👍🏻

sundowndev (approved):

eliecharra (approved):

Issue #374, opened 2021-03-26
Differences output is not predictable


eliecharra:
**Description**
We should sort the differences, actually we sort diff by `type` but we should sort by `Id` too.

https://github.com/cloudskiff/driftctl/blob/main/pkg/analyser/analyzer.go#L87-L89

sundowndev:
Not sure to understand the problem here, also differences does not have a `id` attribute. Can you double check @eliecharra ?

eliecharra:
> Not sure to understand the problem here, also differences does not have a id attribute. Can you double check @eliecharra ?

Differences have a `Res` attribute which can be sorted on `TerraformType` and `TerraformId`

sjourdan:
3 runs and 3 un-ordered outputs: 

```
Found resources not covered by IaC:
  aws_s3_bucket:
    - s3-bucket-1-6qaidu
    - test-resource-exposure-caxyxllsajfrzbwe
  aws_lambda_event_source_mapping:
    - Source: arn:aws:sqs:us-east-1:141177182257:queue-6qaidu.fifo Dest: arn:aws:lambda:us-east-1:141177182257:function:lk6n03-lambda-demo
[...]
```

2

```
Scanned resources:    (62)
Found resources not covered by IaC:
  aws_s3_bucket_policy:
    - test-resource-exposure-caxyxllsajfrzbwe
  aws_iam_user:
    - endgame-s3-demo
[...]
```

3 

```
Scanned resources:    (62)
Found resources not covered by IaC:
  aws_iam_user:
    - endgame-s3-demo
  aws_s3_bucket_policy:
    - test-resource-exposure-caxyxllsajfrzbwe
  aws_sqs_queue:
    - https://sqs.us-east-1.amazonaws.com/141177182257/queue-6qaidu.fifo
[...]
```

Changed resources were well ordered though!

```
Found changed resources:
  - aq5211-lambda-demo (aws_lambda_function):
    ~ QualifiedArn: "arn:aws:lambda:us-east-1:141177182257:function:aq5211-lambda-demo:$LATEST" => "arn:aws:lambda:us-east-1:141177182257:function:aq5211-lambda-demo:1" (computed)
    ~ Version: "$LATEST" => "1" (computed)
```

sjourdan:
only changed resources are concerned, 0.8.0 will deal with the rest. 

LGTM!

PR #385, opened 2021-03-26
Make differences output predictable

sundowndev:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | yes
| 🚀 New feature?   | no
| ⚠ Deprecations?   | no
| ❌ BC Break       | **yes**
| 🔗 Related issues | #374
| ❓ Documentation  | no 

## Description

Differences was expected to be sorted but it turns out the sort was placed at the wrong place in the code, making it completely useless. This PR, which follow the same implementation of #356, intend to fix the issue. We'll surely come across conflicts but they should be easy to resolve.

### DIFF
--- a/pkg/analyser/analysis.go
+++ b/pkg/analyser/analysis.go
@@ -3,6 +3,7 @@ package analyser
 import (
 	"encoding/json"
 	"sort"
+	"strings"
 
 	"github.com/r3labs/diff/v2"
 
@@ -203,5 +204,17 @@ func SortDifferences(diffs []Difference) []Difference {
 		}
 		return diffs[i].Res.TerraformId() < diffs[j].Res.TerraformId()
 	})
+
+	for _, d := range diffs {
+		SortChanges(d.Changelog)
+	}
+
 	return diffs
+}
+
+func SortChanges(changes []Change) []Change {
+	sort.SliceStable(changes, func(i, j int) bool {
+		return strings.Join(changes[i].Path, ".") < strings.Join(changes[j].Path, ".")
+	})
+	return changes
 }
--- a/pkg/analyser/analysis.go
+++ b/pkg/analyser/analysis.go
@@ -2,10 +2,12 @@ package analyser
 
 import (
 	"encoding/json"
+	"sort"
+
+	"github.com/r3labs/diff/v2"
 
 	"github.com/cloudskiff/driftctl/pkg/alerter"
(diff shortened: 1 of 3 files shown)


lotoussa (approved):

pkg/analyser/analysis.go
```diff
@@ -189,3 +191,17 @@ func (a *Analysis) Summary() Summary {
 func (a *Analysis) Alerts() alerter.Alerts {
 	return a.alerts
 }
+
+func (a *Analysis) SortResources() {
+	a.differences = SortDifferences(a.differences)
+}
+
+func SortDifferences(diffs []Difference) []Difference {
+	sort.SliceStable(diffs, func(i, j int) bool {
```
moadibfr, line 200: This does not sort the changed inside difference does it ?

pkg/analyser/analyzer.go
```diff
@@ -82,28 +82,28 @@ func (a Analyzer) Analyze(remoteResources, resourcesFromState []resource.Resourc
 		analysis.AddManaged(stateRes)
 
 		delta, _ := diff.Diff(stateRes, remoteRes)
-		if len(delta) > 0 {
-			sort.Slice(delta, func(i, j int) bool {
```
moadibfr, line 86: I think this is needed but it's missing sort on the path, only sorting on type is not sufficient

moadibfr (changes requested):

pkg/analyser/analysis.go
eliecharra: To make the output full predictable we should sort on `Change.Type` then `Change.Path` ? Or maybe only on `Change.Path` ?

pkg/analyser/analyzer.go
sundowndev: My tests showed that len(delta) is always equal to 1 so the sort has no effect at all. Would be nice if someone could confirm that

pkg/analyser/analysis.go
moadibfr: we could only sort on change.path but the type sort was added for readability at some point I think

pkg/analyser/analyzer.go
moadibfr: no you could have multiple diff on one resource so the len will be greater

eliecharra (approved):

Issue #371, opened 2021-03-26
The number of resources appear unreliable


sjourdan:
**Description**
After a demo today, it's been reported that the number of scanned resources after the spinner during scan is unreliable (the number is different after each run). 

**Environment**
 - OS:
 - driftctl version: 0.7.0

**How to reproduce**

https://youtu.be/-BS65owCCmQ?t=826

**Possible Solution**
We probably need to flush the number when scan is done.

sjourdan:
LGTM!
3 times discovered 62 resources 

```
 LOG_LEVEL=info driftctl-0.7.1 scan --from tfstate://./vpc/terraform.tfstate  --from tfstate+s3://driftctl-tfstates/
INFO[0001] Start reading IaC
INFO[0002] Start scanning cloud provider
Scanned resources:    (62)
```

PR #386, opened 2021-03-26
display last number of resources scanned and hide animation character when progress bar end

lotoussa:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | yes
| 🚀 New feature?   | no
| ⚠ Deprecations?   | no
| ❌ BC Break       | **yes**
| 🔗 Related issues | #371 
| ❓ Documentation  | no

## Description

Make the number of scanned resources after the spinner during scan reliable and hide progress bar animation when it end.

### DIFF
--- a/pkg/output/progress.go
+++ b/pkg/output/progress.go
@@ -47,8 +47,8 @@ func (p *progress) Start() {
 
 func (p *progress) Stop() {
 	if p.started.Swap(false) {
+		Printf("Scanned resources:    (%d)\n", p.count.Load())
 		close(p.endChan)
-		Printf("\n")
 	}
 }
 


moadibfr (approved):

pkg/output/progress.go
```diff
@@ -47,8 +47,8 @@ func (p *progress) Start() {
 
 func (p *progress) Stop() {
 	if p.started.Swap(false) {
+		Printf("Scanning resources:    (%d)\n", p.count.Load())
```
eliecharra, line 50: I think this should remain in the `render` method within `case <-p.endChan:` WDYT @moadibfr

lotoussa: I'm not able to reproduce a good comportment in render()

eliecharra: True, the render close can happen lately and will display inconsistent output.

You can remove useless white spaces and parenthesis there

sundowndev: May be we can write `Scanned resources: %d` instead

lotoussa: The reason why i keep the length of the string is because if i'm not it's display like this at the end:
`Scanned resources: 20 (18)`

eliecharra (approved):

eliecharra:
Edited description as this will not introduce BC break

PR #388, opened 2021-03-29
Implement strict mode

sundowndev:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | no
| 🚀 New feature?   | yes
| ⚠ Deprecations?   | no
| ❌ BC Break       | no
| 🔗 Related issues | #331
| ❓ Documentation  | yes 

## Description

When scanning a brand new AWS account, some users may see irrelevant results about AWS default resources. We can ignore these resources by default and add a `--strict` flag to disable that behavior.

### Ignored resources

- [x] aws_iam_role_policy where role has path matching `/aws-service-roles/*`
- [x] aws_iam_role where path matches `/aws-service-roles/*`
- [x] aws_iam_policy_attachment. where role has path matching `/aws-service-roles/*`


pkg/middlewares/aws_iam_role_defaults.go
```diff
@@ -0,0 +1,67 @@
...
+
+		if !isIgnored {
+			newRemoteResources = append(newRemoteResources, remoteResource)
+			continue
+		}
+
+		logrus.WithFields(logrus.Fields{
+			"id":   remoteResource.TerraformId(),
+			"type": remoteResource.TerraformType(),
+		}).Debug("Ignoring default iam policy attachment as it is not managed by IaC")
```
sundowndev, line 61: iam role*

pkg/middlewares/aws_security_group_defaults.go
```diff
@@ -0,0 +1,52 @@
...
+import (
+	"github.com/sirupsen/logrus"
+
+	"github.com/cloudskiff/driftctl/pkg/resource"
+	"github.com/cloudskiff/driftctl/pkg/resource/aws"
+)
+
+const defaultAwsSecurityGroupName = "default"
+
+// When scanning a brand new AWS account, some users may see irrelevant results about default AWS role policies.
```
lotoussa, line 12: aws security group defaults

pkg/middlewares/aws_security_group_defaults.go
```diff
@@ -0,0 +1,52 @@
...
+
+func NewAwsSecurityGroupDefaults() AwsSecurityGroupDefaults {
+	return AwsSecurityGroupDefaults{}
+}
+
+func (m AwsSecurityGroupDefaults) Execute(remoteResources, resourcesFromState *[]resource.Resource) error {
+	newRemoteResources := make([]resource.Resource, 0)
+
+	for _, remoteResource := range *remoteResources {
+		// Ignore all resources other than iam role
```
lotoussa, line 24: aws security group defaults

pkg/middlewares/aws_security_group_rules_defaults.go
```diff
@@ -0,0 +1,70 @@
+package middlewares
+
+import (
+	"github.com/sirupsen/logrus"
+
+	"github.com/cloudskiff/driftctl/pkg/resource"
+	"github.com/cloudskiff/driftctl/pkg/resource/aws"
+)
+
+// When scanning a brand new AWS account, some users may see irrelevant results about default AWS role policies.
```
lotoussa, line 10: aws security group rules defaults

pkg/cmd/scan.go
```diff
@@ -137,6 +138,11 @@ func NewScanCmd() *cobra.Command {
 		"Use those HTTP headers to query the provided URL.\n"+
 			"Only used with tfstate+http(s) backend for now.\n",
 	)
+	fl.BoolVar(&opts.Strict,
+		"strict",
+		false,
+		"Don't ignore default resources",
```
eliecharra, line 144: We nee to clarify what we call default resources. There are some default resources, for example a default VPC which should not be ignored regardless of the strict flag. Theses resources should be ignored if they are not managed by IaC, this is the actual logic.

The strict flag goal is to ignore resources created by the cloud provider to make some service work, a role for SSO for example. This is a different usecase.

For example logic around security groups with `default` name should not part of this PR their behavior should remain unchanged.

lotoussa (approved):

pkg/driftctl.go
```diff
@@ -53,6 +54,15 @@ func (d DriftCTL) Run() (*analyser.Analysis, error) {
 		middlewares.NewAwsSNSTopicPolicyExpander(),
 	)
 
+	if !d.strictMode {
+		middleware = append(middleware,
+			middlewares.NewAwsIamPolicyAttachmentDefaults(),
```
eliecharra, line 59: As we discussed in the issue, maybe a single middleware per cloud provider could be enough ? WDYT @wbeuil

pkg/middlewares/route53_records.go
```diff
@@ -30,7 +30,7 @@ func (m Route53DefaultZoneRecordSanitizer) Execute(remoteResources, resourcesFro
 
 		record, _ := remoteResource.(*aws.AwsRoute53Record)
 
-		if !isDefaultRecord(record) {
+		if !isDefaultRoute53Record(record) {
```
eliecharra, line 33: Is this related to this PR ?

pkg/middlewares/aws_iam_role_defaults.go
```diff
@@ -0,0 +1,67 @@
...
+		}
+
+		if existInState {
+			newRemoteResources = append(newRemoteResources, remoteResource)
+			continue
+		}
+
+		isIgnored := false
+		for _, id := range ignoredIamRoleIds {
+			if remoteResource.TerraformId() == id {
```
eliecharra, line 48: Actually they can be many of `AWSServiceRole*` roles.

image

I think for roles we should ignore them using the path attribute

image

Every role with a path starting with `/aws-service-role` should be skipped in strict mode **if they are not managed by IaC**

Let's ask our AWS certified guy @wbeuil about it, is it the right logic ?

pkg/middlewares/aws_iam_policy_attachment_defaults.go
```diff
@@ -0,0 +1,65 @@
...
+		}
+
+		if existInState {
+			newRemoteResources = append(newRemoteResources, remoteResource)
+			continue
+		}
+
+		isIgnored := false
+		for _, id := range ignoredIamPolicyAttachmentIds {
+			if remoteResource.TerraformId() == id {
```
eliecharra, line 46: Same remark as for the roles, but as we do not have a `path` attribute maybe we can match the string with wildcards ?

pkg/middlewares/aws_iam_role_policy_defaults.go
```diff
@@ -0,0 +1,66 @@
...
+		}
+
+		if existInState {
+			newRemoteResources = append(newRemoteResources, remoteResource)
+			continue
+		}
+
+		isIgnored := false
+		for _, id := range ignoredIamRolePolicyIds {
+			if remoteResource.TerraformId() == id {
```
eliecharra, line 47: Same remark as for the roles.

pkg/middlewares/aws_security_group_defaults.go
```diff
@@ -0,0 +1,52 @@
...
+
+	"github.com/cloudskiff/driftctl/pkg/resource"
+	"github.com/cloudskiff/driftctl/pkg/resource/aws"
+)
+
+const defaultAwsSecurityGroupName = "default"
+
+// When scanning a brand new AWS account, some users may see irrelevant results about default AWS security group.
+// We ignore these resources by default when strict mode is disabled.
+type AwsSecurityGroupDefaults struct{}
```
eliecharra, line 14: We already have handled these cases in `VPCDefaultSecurityGroupSanitizer`

pkg/middlewares/aws_security_group_rules_defaults.go
```diff
@@ -0,0 +1,70 @@
...
+import (
+	"github.com/sirupsen/logrus"
+
+	"github.com/cloudskiff/driftctl/pkg/resource"
+	"github.com/cloudskiff/driftctl/pkg/resource/aws"
+)
+
+// When scanning a brand new AWS account, some users may see irrelevant results about default AWS security group rules.
+// We ignore these resources by default when strict mode is disabled.
+type AwsSecurityGroupRuleDefaults struct{}
```
eliecharra, line 12: I think this is already handled too

eliecharra (changes requested):
We may need to redefine the exact scope of what we want to ignore when using the `--strict` flag

pkg/cmd/scan.go
wbeuil: This is not a simple one and it will grow as we expand AWS resources.

We probably need to do a call to redefine what we want to be ignored as default. One quick way for me is to see what I have in my `.driftignore`. I want all the Orga/SSO-related IAM resources inside it since it is created by default by AWS when I use SSO and I won't modify them. 

Indeed as @eliecharra explains above, security groups and security group rules are a special kind of resources with defaults but we must keep them under our radar, e.g. we should not ignore them by default if we decide to managed them.

pkg/driftctl.go
wbeuil: Yes I agree, once we redefine what we should put inside it, let's gather them inside one middleware.

pkg/middlewares/aws_iam_role_defaults.go
wbeuil: This is what a service-linked role is:

`A service-linked role is a unique type of IAM role that is linked directly to an AWS service. Service-linked roles are predefined by the service and include all the permissions that the service requires to call other AWS services on your behalf. The linked service also defines how you create, modify, and delete a service-linked role. A service might automatically create or delete the role. It might allow you to create, modify, or delete the role as part of a wizard or process in the service. Or it might require that you use IAM to create or delete the role. Regardless of the method, service-linked roles make setting up a service easier because you don't have to manually add the necessary permissions for the service to complete actions on your behalf.`

We can easily ignore them by default

pkg/middlewares/aws_security_group_defaults.go
wbeuil: Yep, it's already done

pkg/middlewares/aws_security_group_rules_defaults.go
wbeuil: It's handled implicitly since we do ignore the default security group.

pkg/middlewares/route53_records.go
sundowndev: Yes, `isDefaultRecord` was too generic and not specific to Route53. As I created a similar method for various resources, I needed to rename this method so it does not conflict with other methods in the same package.

pkg/cmd/scan.go
sundowndev: > we should not ignore them by default if we decide to managed them

Indeed, the goal of this PR is to ignore the default security groups if they're not managed by IaC. Security group rules are ignored if they're part of a default SG and have known default properties (protocol, type). Any resource managed by IaC will not be ignored regardless of the strict mode state.

pkg/middlewares/aws_security_group_rules_defaults.go
sundowndev: So should I delete this middleware? It's unclear to me what you mean by "this is already handled" since users see default security group rules as drifted resources in their scan result.

pkg/middlewares/route53_records.go
wbeuil: Why not, I'm not against it. But indeed, it is not related to this PR.

What is the other method that conflict with this one ?

pkg/middlewares/aws_security_group_rules_defaults.go
wbeuil: Can you reproduce this behaviour where default sg rules are seem as drifted ? I'm not sure of that use case

sundowndev: @wbeuil it has been seen in that video, referenced in the issue  https://github.com/cloudskiff/driftctl/issues/331#issuecomment-808262088

pkg/driftctl.go
sundowndev: The method is not used anymore, so I reverted this change

pkg/middlewares/route53_records.go
sundowndev: The method is not used anymore, so I reverted this change

pkg/middlewares/aws_iam_policy_attachment_defaults.go
sundowndev: > but as we do not have a path attribute

Not sure what you mean, iam policies have a `Roles` attribute that reference one or more iam roles, isn't it ?

pkg/middlewares/aws_iam_role_defaults_test.go
```diff
@@ -0,0 +1,133 @@
+package middlewares
```
sundowndev, line 1: May be I should put all tests in the same test file

pkg/cmd/scan.go
eliecharra: Maybe we could reword this as `Ignore resources managed by the cloud provider` ?

pkg/middlewares/aws_defaults.go
```diff
@@ -0,0 +1,204 @@
...
+				break
+			}
+		}
+
+		if existInState {
+			newRemoteResources = append(newRemoteResources, remoteResource)
+			continue
+		}
+
+		match, err := filepath.Match(defaultIamRolePathGlob, *remoteResource.(*aws.AwsIamRole).Path)
```
eliecharra, line 45: Why not doing a simple `strings.HasPrefix(*remoteResource.(*aws.AwsIamRole).Path, "/aws-service-role/")` I don't think we need a wildcard match there

pkg/middlewares/aws_defaults.go
```diff
@@ -0,0 +1,204 @@
...
+
+		if !match {
+			newRemoteResources = append(newRemoteResources, remoteResource)
+			continue
+		}
+
+		logrus.WithFields(logrus.Fields{
+			"id":   remoteResource.TerraformId(),
+			"type": remoteResource.TerraformType(),
+		}).Debug("Ignoring default iam role as it is not managed by IaC")
```
eliecharra, line 58: The wording should be changed to something like `Ignoring IAM role as it is a resource managed by the cloud provider`

pkg/middlewares/aws_defaults.go
```diff
@@ -0,0 +1,204 @@
...
+
+	for _, remoteResource := range *remoteResources {
+		// Ignore all resources other than iam role
+		if remoteResource.TerraformType() != aws.AwsIamRoleResourceType {
+			newRemoteResources = append(newRemoteResources, remoteResource)
+			continue
+		}
+
+		existInState := false
+		for _, stateResource := range *resourcesFromState {
```
eliecharra, line 33: I don't think we should check if the resource is managed ? @wbeuil Should we ignore theses cloud provider defaults only if they are not managed ? I'm not sure this makes sense to manage theses kind of resources using terraform (imported in terraform), some of them cannot be modified for example.

pkg/cmd/scan.go
wbeuil: In the AWS ecosystem, `managed resources` or in another extend`managed services` can be used interchangeably for specific use (e.g. RDS == fully managed relational database). But in the terraform ecosystem an RDS instance (aws_db_instance) is not what we want our user to see as `managed by the cloud provider`.

Maybe we can write `Ignore resources created by default by the cloud provider`

pkg/middlewares/aws_defaults.go
```diff
@@ -0,0 +1,204 @@
...
+		}
+
+		for _, roleId := range *remoteResource.(*aws.AwsIamPolicyAttachment).Roles {
+			var role *aws.AwsIamRole
+			for _, res := range *remoteResources {
+				if res.TerraformType() != aws.AwsIamRoleResourceType {
+					continue
+				}
+
+				if res.(*aws.AwsIamRole).Id == roleId {
```
eliecharra, line 96: You can do `res.TerraformId()` instead of `res.(*aws.AwsIamRole).Id`

pkg/cmd/scan.go
sundowndev: What do you think of this description for the strict flag?

`Resources managed by the cloud provider are ignored by default. Enabling the strict mode disables this behavior. Note that enabling this option can cause unexpected scan results.`

wbeuil: One more time, in AWS, Lambda is a resource (in terraform way) or service (in aws way) managed by the cloud provider. It is not what we want our user to understand.

pkg/middlewares/aws_defaults.go
```diff
@@ -0,0 +1,204 @@
...
+		}
+
+		if existInState {
+			newRemoteResources = append(newRemoteResources, remoteResource)
+			continue
+		}
+
+		for _, roleId := range *remoteResource.(*aws.AwsIamPolicyAttachment).Roles {
+			var role *aws.AwsIamRole
+			for _, res := range *remoteResources {
```
eliecharra, line 91: You can simplify this code 

```go
for _, res := range *remoteResources {
  if res.TerraformType() == aws.AwsIamRoleResourceType && res.TerraformId() == roleId {
    role = res.(*aws.AwsIamRole)
    break
  }
}
```

eliecharra (changes requested):

pkg/cmd/scan.go
eliecharra: I get your point about the managed wording👍🏻 But it will be wrong to say `Ignore resources created by default by the cloud provider` as we already ignore theses resources (for example default route53 records in zones). 

Strict mode is here to make driftctl not to ignore another kind of resource : resource required by AWS services to work.

Maybe we can use something more like `Do not ignore resources required by cloud provider service to operate` ?

pkg/middlewares/aws_defaults.go
wbeuil: I've never managed the default with terraform so maybe @sjourdan can help answering this question. But as for the "should we ignore the defaults only if they are not managed`, I answer let's ignore them all the time.

pkg/cmd/scan.go
wbeuil: > I get your point about the managed wording👍🏻 But it will be wrong to say Ignore resources created by default by the cloud provider as we already ignore theses resources (for example default route53 records in zones).

You're right :(

>Strict mode is here to make driftctl not to ignore another kind of resource : resource required by AWS services to work. Maybe we can use something more like Do not ignore resources required by cloud provider service to operate ?

I don't know why but I find it wrong when I say out loud ... One time @sjourdan to the rescue here ^^. Love to have his view on this one.

pkg/middlewares/aws_defaults.go
sundowndev: Shouldn't we consider users may want to manage those resources anyway ? If we ignore them all the time and the user manages them, the resources will appear as missing on cloud provider. Not sure how we should handle that behavior then.

eliecharra: We can exclude them from `resourcesFromState` too

sundowndev: Right, and the user can use the strict flag when they want to manage these resources. It makes sense

eliecharra: Exactly

wbeuil (approved):
I'm good with that one

pkg/middlewares/aws_defaults.go
eliecharra: Missing a `break` to avoid looping over the whole slice when we found the role

pkg/middlewares/aws_defaults.go
eliecharra: up

pkg/middlewares/aws_defaults.go
```diff
@@ -0,0 +1,181 @@
...
+			if match := strings.HasPrefix(*role.Path, defaultIamRolePathPrefix); !match {
+				newRemoteResources = append(newRemoteResources, remoteResource)
+				continue
+			}
+		}
+
+		logrus.WithFields(logrus.Fields{
+			"id":   remoteResource.TerraformId(),
+			"type": remoteResource.TerraformType(),
+		}).Debug("Ignoring default iam policy attachment as it is not managed by IaC")
```
eliecharra, line 107: Still not the good wording

pkg/middlewares/aws_defaults.go
```diff
@@ -0,0 +1,181 @@
...
+
+		if match := strings.HasPrefix(*role.Path, defaultIamRolePathPrefix); !match {
+			newRemoteResources = append(newRemoteResources, remoteResource)
+			continue
+		}
+
+		logrus.WithFields(logrus.Fields{
+			"id":   remoteResource.TerraformId(),
+			"type": remoteResource.TerraformType(),
+		}).Debug("Ignoring default iam role policy as it is not managed by IaC")
```
eliecharra, line 159: Not the good wording

pkg/middlewares/aws_defaults.go
```diff
@@ -0,0 +1,181 @@
...
+
+		for _, roleId := range *remoteResource.(*aws.AwsIamPolicyAttachment).Roles {
+			var role *aws.AwsIamRole
+			for _, res := range *remoteResources {
+				if res.TerraformType() == aws.AwsIamRoleResourceType && res.TerraformId() == roleId {
+					role = res.(*aws.AwsIamRole)
+				}
+			}
+
+			// If we couldn't find the linked role, include the resource
```
eliecharra, line 92: I'm not sure this is the expected behavior, in theory it should never happen as an attachment cannot exist without a role.

WDYT @wbeuil I'm not clear about this

eliecharra (changes requested):

pkg/middlewares/aws_defaults.go
wbeuil: Since we're looping across remote resources, indeed, it should never happen. Good catch.

pkg/middlewares/aws_defaults.go
```diff
@@ -0,0 +1,128 @@
...
+func (m AwsDefaults) Execute(remoteResources, resourcesFromState *[]resource.Resource) error {
+	newRemoteResources := append([]resource.Resource{}, *remoteResources...)
+	newResourcesFromState := append([]resource.Resource{}, *resourcesFromState...)
+	resourcesToIgnore := make([]resource.Resource, 0)
+
+	resourcesToIgnore = append(resourcesToIgnore, m.awsIamRoleDefaults(*remoteResources)...)
+	resourcesToIgnore = append(resourcesToIgnore, m.awsIamPolicyAttachmentDefaults(*remoteResources)...)
+	resourcesToIgnore = append(resourcesToIgnore, m.awsIamRolePolicyDefaults(*remoteResources)...)
+
+	for _, resourceToIgnore := range resourcesToIgnore {
```
eliecharra, line 107: Why not using the same logic than the others middlewares ? Rather than removing element from the slice, append them if they are not ignored to a new slice and replace the original one. WDYT @wbeuil @moadibfr 

If we choose to remove items why mutate the original slice directly ?

eliecharra (commented):
Almost good, only two comments

pkg/middlewares/aws_defaults.go
sundowndev: Note that we cannot perform indexing on a slice pointer, that's why a copied the original slice

sundowndev: The problem here is that we perform a loop for each resource to ignore. That means the loop can pass multiple times on the same element. If we used the same logic than the others middlewares, it would append resources more than once and break the scan result. But may be you had something better in mind?

eliecharra (approved):

Issue #347, opened 2021-03-17
Terraform AWS Provider v2 support


sjourdan:
**Description**

driftctl currently supports only v3 of the AWS Provider - many people still use v2.
Let's support it!

vainkop:
@sjourdan any updates on that? I see that you have `Past due by 5 days` with `37% complete` for `v0.8.0` release in https://github.com/cloudskiff/driftctl/milestone/11 so do you have a more precise estimate for support of aws provider v2 in particular?

sjourdan:
hey @vainkop! Thanks for your continued interest!

We added a much-needed bugfix release `v0.7.1` in the roadmap after a ton of feedback and some bugs that were more important to fix now than later. TBH it includes a lot of feedback and requests you already heard about during the Anton Babenko show!

So you´re right, it took much longer to release `v0.8.0` with v2/v3 AWS provider support. I think `v0.7.1` will be released by tomorrow (2021/4/9), so I also think `v0.8.0` will be available by the end of the month as the full team will work again on this, and hopefully, earlier.

Thank you for your patience :pray: !

vainkop:
@eliecharra too bad you've moved that from v0.8.0 to v0.9.0 😞

vainkop:
@sjourdan release v0.9.0 is here & no support for AWS provider v2 again? :(

sjourdan:
hey @vainkop you're right, we're still hard at work on the major rewrite needed to do this properly, so we decided to ship a feature useful to some users (the .driftignore generator) and a couple of bug fixes first. 

The good news is that the very last resource we needed to rewrite is about to be merged. Then the testing phase of the pre-release will begin. 

The whole thing much bigger than expected, but it was definitely highly needed to support any provider version. 

I'll let you know when a pre-release build is ready if you want to try it before it's released to the public, as you seemed to have very interesting use cases. 

Thank you for your patience and support :pray:

GeraldCrescione:
Hi @vainkop 
just doing a quick follow up on this one. With [the v0.10.0 release](https://github.com/cloudskiff/driftctl/releases/tag/v0.10.0) driftctl now supports any Terraform AWS provider version. Here's a [blog post](https://driftctl.com/how-to-use-driftctl-with-different-terraform-providers/) that relates to that.

PR #389, opened 2021-03-29
Add resource factory

eliecharra:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | no
| 🚀 New feature?   | no
| ⚠ Deprecations?   | no
| ❌ BC Break       | no
| 🔗 Related issues | #347
| ❓ Documentation  | no

## Description

* Add ResourceFactory
* Update middlewares to use resource factory and thus resources created from middlewares can now be filtered


pkg/middlewares/aws_instance_block_device_test.go
```diff
@@ -118,12 +123,21 @@ func TestAwsInstanceBlockDeviceResourceMapper_Execute(t *testing.T) {
 					},
 				},
 			},
+			func(factory *terraform.MockResourceFactory) {
+				factory.On("CreateResource", mock.Anything, "aws_ebs_volume").Times(2).Return(nil, nil)
```
wbeuil, line 127: Just FYI, you can use here `Twice()` instead of `Times(2)` but no need to change it.

eliecharra: Yeah I know but I prefer to always use `Times`

wbeuil: If i == 1 you use `Once` and if i > 1 `Times(i)` ?

eliecharra: Yeah and I was wondering myself if I should use `Times` everywhere

wbeuil (approved):
Nice one ! I will push really soon a PR that start testing those middlewares.

pkg/middlewares/aws_instance_block_device_test.go
wbeuil: Indeed, it is more explicit.

PR #396, opened 2021-03-30
Strict mode lotoussa

lotoussa:
Add aws_iam_role_policy_attachment to this branch.

### DIFF
--- a/pkg/middlewares/aws_iam_policy_attachment_defaults_test.go
+++ b/pkg/middlewares/aws_iam_policy_attachment_defaults_test.go
@@ -16,7 +16,7 @@ func TestAwsIamPolicyAttachmentDefaults_Execute(t *testing.T) {
 		name               string
 		remoteResources    []resource.Resource
 		resourcesFromState []resource.Resource
-		expected           []resource.Resource
+		expected           diff.Changelog
 	}{
 		{
 			"test that default iam policy attachment are excluded when not managed by IaC",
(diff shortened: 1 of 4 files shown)


pkg/middlewares/aws_iam_policy_attachment_defaults_test.go
```diff
@@ -0,0 +1,91 @@
...
+			[]resource.Resource{
+				&aws.AwsIamPolicyAttachment{
+					Id: "driftctl_test-arn:aws:iam::0123456789:policy/driftctl",
+				},
+			},
+			[]resource.Resource{
+				&aws.AwsIamPolicyAttachment{
+					Id: "driftctl_test-arn:aws:iam::0123456789:policy/driftctl",
+				},
+			},
```
sundowndev, line 40: Could you instead check for diffs in the changelog instead of checking the state ? Since the goal is to prevent adding noice in the changelog's diffs.

pkg/middlewares/aws_iam_policy_attachment_defaults_test.go
```diff
@@ -0,0 +1,91 @@
...
+
+			changelog, err := diff.Diff(tt.remoteResources, tt.resourcesFromState)
+			if err != nil {
+				t.Fatal(err)
+			}
+			if len(changelog) > 0 {
+				for _, change := range changelog {
+					t.Errorf("%s got = %v, want %v", strings.Join(change.Path, "."), awsutil.Prettify(change.From), awsutil.Prettify(change.To))
+				}
+			}
```
sundowndev, line 87: To compare with the expected the changelog: 

```go
			changelog, err := diff.Diff(tt.remoteResources, tt.resourcesFromState)
			if err != nil {
				t.Fatal(err)
			}

			diffs, err := diff.Diff(tt.expected, changelog)
			if err != nil {
				t.Fatal(err)
			}

			for _, change := range diffs {
				t.Errorf("%s got = %v, want %v", strings.Join(change.Path, "."), awsutil.Prettify(change.From), awsutil.Prettify(change.To))
			}
```

pkg/middlewares/aws_iam_policy_attachment_defaults_test.go
lotoussa: Yes

sundowndev (approved):
👍🏽

PR #402, opened 2021-04-01
Fix aws_iam_policy_attachment false-positives

lotoussa:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | yes
| 🚀 New feature?   | no
| ⚠ Deprecations?   | no
| ❌ BC Break       | no
| 🔗 Related issues | #366 
| ❓ Documentation  | no

## Description

Fix false positive for aws_iam_policy_attachment Users and Groups
Add Acc test.


sundowndev (approved):

pkg/resource/aws/testdata/acc/aws_iam_policy_attachment/iam_policy_attachment.tf
```diff
@@ -0,0 +1,61 @@
+resource "aws_lambda_function" "yadda" {
```
eliecharra, line 1: This is useless

pkg/resource/aws/aws_iam_policy_attachment_ext.go
```diff
@@ -0,0 +1,19 @@
...
+import (
+	"github.com/cloudskiff/driftctl/pkg/resource"
+)
+
+func (r *AwsIamPolicyAttachment) NormalizeForState() (resource.Resource, error) {
+	return r, nil
+}
+
+func (r *AwsIamPolicyAttachment) NormalizeForProvider() (resource.Resource, error) {
+	if r.Groups != nil && len(*r.Groups) == 0 {
```
eliecharra, line 12: This should be done in `NormalizeForState` too, if we have an empty slice in the state you'll introduced a new false positive

eliecharra (changes requested):

pkg/resource/aws/aws_iam_policy_attachment_ext.go
lotoussa: Oh right,
I was thinking that this modification should not be made because it introduces a lot of problems in the tests which seemed to me unsuitable for the ticket, but I'll take a look at it again and solve them

sundowndev: Should be ok now

eliecharra (approved):

Issue #392, opened 2021-03-30
aws_route false-positive (as deleted resource)


sjourdan:
**Description**

The following route is seen as deleted while it's not: 

```
$ driftctl scan --from tfstate://iam/terraform.tfstate --from tfstate://s3/terraform.tfstate --from tfstate://vpc/terraform.tfstate
Scanning resources: ⢿ (59)
Found deleted resources:
  aws_route:
    - Table: rtb-02d22056ec3a135ff, Destination: 10.0.1.0/24
Found 11 resource(s)
 - 90% coverage
 - 10 covered by IaC
 - 0 not covered by IaC
 - 1 deleted on cloud provider
 - 0/10 drifted from IaC
 ```
Here's the JSON output containing the route *twice* 

[output.json.gz](https://github.com/cloudskiff/driftctl/files/6229420/output.json.gz)

**Environment**
 - OS: Fedora 33
 - driftctl version: 0.7.0

**How to reproduce**

HCL as simple as that:

```
resource "aws_route_table" "table" {
  vpc_id = aws_default_vpc.default.id

  tags = {
    Name      = "Another Route Table"
    Terraform = "true"
  }
}

resource "aws_route" "route_1" {
  route_table_id         = aws_route_table.table.id
  destination_cidr_block = "10.0.1.0/24"
  gateway_id = aws_internet_gateway.staging_gw.id
}
```

**Possible Solution**

**Additional context**

eliecharra:
It happen when it exist a route both as a dedicated resource and embedded under `route` attribute in the route table resource.

When expanding the route in the middleware, we should first ensure that this route does not already exist.

sundowndev:
While trying reproducing that bug, I noticed I had another false positive on the `state` attribute of the `aws_route` resource.

My state (redacted):

```json
{
      "mode": "managed",
      "type": "aws_route",
      "name": "route_1",
      "provider": "provider[\"registry.terraform.io/hashicorp/aws\"]",
      "instances": [
        {
          "schema_version": 0,
          "attributes": {
            "destination_cidr_block": "10.0.1.0/24",
            "destination_ipv6_cidr_block": "",
            "destination_prefix_list_id": "",
            "egress_only_gateway_id": "",
            "gateway_id": "",
            "id": "<redacted>",
            "instance_id": "",
            "instance_owner_id": "",
            "local_gateway_id": "",
            "nat_gateway_id": "",
            "network_interface_id": "",
            "origin": "CreateRoute",
[...]

PR #404, opened 2021-04-02
Fix false positive on aws_route resources

sundowndev:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | yes
| 🚀 New feature?   | no
| ⚠ Deprecations?   | no
| ❌ BC Break       | no
| 🔗 Related issues | #392
| ❓ Documentation  | yes 

## Description

The AWS route table expander middleware was supposed to turn routes embeded in a route table into dedicated resources. But when a route is already a dedicated resource, the scan result shows it as a deleted resource.

### DIFF
--- a/pkg/middlewares/aws_route_table_expander.go
+++ b/pkg/middlewares/aws_route_table_expander.go
@@ -56,9 +56,9 @@ func (m AwsRouteTableExpander) Execute(remoteResources, resourcesFromState *[]re
 
 		var err error
 		if isDefault {
-			err = m.handleDefaultTable(defaultTable, &newList)
+			err = m.handleDefaultTable(defaultTable, &newList, *resourcesFromState)
 		} else {
-			err = m.handleTable(table, &newList)
+			err = m.handleTable(table, &newList, *resourcesFromState)
 		}
 
 		if err != nil {
@@ -69,7 +69,7 @@ func (m AwsRouteTableExpander) Execute(remoteResources, resourcesFromState *[]re
 	return nil
 }
 
-func (m *AwsRouteTableExpander) handleTable(table *aws.AwsRouteTable, results *[]resource.Resource) error {
+func (m *AwsRouteTableExpander) handleTable(table *aws.AwsRouteTable, results *[]resource.Resource, resourcesFromState []resource.Resource) error {
 	if table.Route == nil ||
 		len(*table.Route) < 1 {
 		return nil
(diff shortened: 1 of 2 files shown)


lotoussa (approved):
LGTM

pkg/middlewares/aws_route_table_expander_test.go
```diff
@@ -63,11 +63,21 @@ func TestAwsRouteTableExpander_Execute(t *testing.T) {
 			},
 		},
 		{
-			"test route are expanded",
+			"test routes are expanded except when it already exist",
```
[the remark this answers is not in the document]
wbeuil: Can you, instead of changing the two current tests, create a new one. One test that won't expand already created routes in a default and a non-default route-table ?

wbeuil (changes requested):

pkg/middlewares/aws_route_table_expander.go
```diff
@@ -80,6 +80,25 @@ func (m *AwsRouteTableExpander) handleTable(table *aws.AwsRouteTable, results *[
 			m.alerter.SendAlert(aws.AwsRouteTableResourceType, newInvalidRouteAlert(aws.AwsRouteTableResourceType, table.Id))
 			continue
 		}
+
+		routeAlreadyExists := false
```
eliecharra, line 84: You can put this code (check if the route exist in a slice of resources) in a unexported struct `AwsRouteTableExpander` method to reuse logic for route and default tables.

eliecharra (changes requested):
Almost all good, I got a small on DRY

pkg/middlewares/aws_route_table_expander.go
```diff
@@ -157,3 +169,19 @@ func (m *AwsRouteTableExpander) handleDefaultTable(table *aws.AwsDefaultRouteTab
 
 	return nil
 }
+
+func (m *AwsRouteTableExpander) routeExists(routeId string, resourcesFromState []resource.Resource) bool {
+	for _, res := range resourcesFromState {
+		if res.TerraformType() != aws.AwsRouteResourceType {
```
eliecharra, line 175: It may be more readable to do 

```go
for _, res := range resourcesFromState {
  if res.TerraformType() == aws.AwsRouteResourceType && res.TerraformId() == routeId {
    return true
  }
}
return false
```

eliecharra (changes requested):
Still one last readability comment

eliecharra (approved):
👍🏻

wbeuil (approved):
GO !

Issue #382, opened 2021-03-26
Explicitly mention 404 error using the HTTPS backend 


sjourdan:
**Description**

Using the `--from tfstate+https://` new backend support to access terraform states, if I request the wrong file, I expect to know about it (like "`Sorry, no t3rr4f0rm.tfstate is available at this location`"), more than a json parsing issue (as there's no file to parse obviously) 

```shell 
$ driftctl scan --from tfstate+https://<redacted>.net/tfstates/wrong.tfstate
Scanning resources: ⣾ (0)
2 problems:

- Unsupported state file format: The state file could not be parsed as JSON: syntax error at byte offset 1.
- Unsupported state file format: The state file does not have a "version" attribute, which is required to identify the format version.

```

A HTTP/404 is returned: 

```
$ curl -IL https://<redacted>.net/tfstates/wrong.tfstate
HTTP/2 404 
server: nginx
date: Fri, 26 Mar 2021 16:45:54 GMT
content-type: text/html
content-length: 146

```

**Environment**
 - OS: linux
 - driftctl version: 0.7.0

**How to reproduce**

**Possible Solution**

**Additional context**

sundowndev:
I can take this since I worked on initial implementation. I think every status code <200 and >=400 should throw an error.

eliecharra:
> I can take this since I worked on initial implementation. I think every status code <200 and >=400 should throw an error.

We should ensure that we follow 301 & 302 redirections too

sundowndev:
> > I can take this since I worked on initial implementation. I think every status code <200 and >=400 should throw an error.
> 
> We should ensure that we follow 301 & 302 redirections too

I can confirm the backend already support 301 & 302 redirections.

sundowndev:
@eliecharra You changed milestone to v0.9 but the PR already got merged into v0.8, should we put the milestone back to v0.8 then?

eliecharra:
It was a mistake indeed @sundowndev

sjourdan:
LGTM! 

```
driftctl-0.8.0 scan --from tfstate+https://***.com/terraform.tfstate 
Scanned states (1)      
Failed to read state file: The state file could not be read: error requesting HTTP(s) backend state: status code: 404
```

To be cheeky, I created a folder named terraform.tfstate on the remote server ... 

```
❯  curl -IL https://***.com/terraform.tfstate
HTTP/2 301 
date: Fri, 07 May 2021 15:06:16 GMT
[...]

PR #406, opened 2021-04-06
Return explicit error on bad http status code

sundowndev:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | yes
| 🚀 New feature?   | no
| ⚠ Deprecations?   | no
| ❌ BC Break       | no
| 🔗 Related issues | #382
| ❓ Documentation  | no 

## Description

> Using the `--from tfstate+https://` new backend support to access terraform states, if I request the wrong file, I expect to know about it (like "`Sorry, no t3rr4f0rm.tfstate is available at this location`"), more than a json parsing issue (as there's no file to parse obviously) 

This patch returns an explicit error when the HTTP status code is below 200 or above 399.

**Example**

```shell
$  driftctl scan --from tfstate+https://gitlab.com/api/v4/projects/25255199/terraform/state/terraform/versions/1
Scanning resources: ⣻ (0)
error in backend HTTP(s): non-200 OK status code: 401 Unauthorized body: "{"message":"401 Unauthorized"}"
exit status 1
```

### DIFF
--- a/pkg/iac/terraform/state/backend/http_reader.go
+++ b/pkg/iac/terraform/state/backend/http_reader.go
@@ -1,6 +1,8 @@
 package backend
 
 import (
+	"bytes"
+	"fmt"
 	"io"
 	"net/http"
 
@@ -29,6 +31,16 @@ func NewHTTPReader(rawURL string, opts *Options) (*HTTPBackend, error) {
 	res, err := client.Do(req)
 	if err != nil {
 		return nil, err
+	}
+
+	buf := new(bytes.Buffer)
+	_, err = buf.ReadFrom(res.Body)
+	if err != nil {
+		return nil, err
+	}
+
+	if res.StatusCode < 200 || res.StatusCode >= 400 {
+		return nil, fmt.Errorf("error in backend HTTP(s): non-200 OK status code: %s body: \"%s\"", res.Status, buf.String())
 	}
 
 	return &HTTPBackend{rawURL, res.Body}, nil
--- a/pkg/iac/terraform/state/backend/http_reader.go
+++ b/pkg/iac/terraform/state/backend/http_reader.go
@@ -6,6 +6,7 @@ import (
 	"net/http"
 
 	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
 )
 
 const BackendKeyHTTP = "http"
@@ -38,8 +39,10 @@ func NewHTTPReader(rawURL string, opts *Options) (*HTTPBackend, error) {
 		return nil, err
 	}
 
+	logrus.WithFields(logrus.Fields{"body": buf.String()}).Trace("HTTP(s) backend response")
+
 	if res.StatusCode < 200 || res.StatusCode >= 400 {
-		return nil, errors.Errorf("error in backend HTTP(s): non-200 OK status code: %s body: \"%s\"", res.Status, buf.String())
+		return nil, errors.Errorf("error requesting HTTP(s) backend state: status code: %d", res.StatusCode)
 	}
 
 	return &HTTPBackend{rawURL, res.Body}, nil
--- a/pkg/iac/terraform/state/backend/http_reader.go
+++ b/pkg/iac/terraform/state/backend/http_reader.go
@@ -2,7 +2,6 @@ package backend
 
 import (
 	"bytes"
-	"fmt"
 	"io"
 	"net/http"
 
@@ -40,7 +39,7 @@ func NewHTTPReader(rawURL string, opts *Options) (*HTTPBackend, error) {
 	}
 
 	if res.StatusCode < 200 || res.StatusCode >= 400 {
-		return nil, fmt.Errorf("error in backend HTTP(s): non-200 OK status code: %s body: \"%s\"", res.Status, buf.String())
+		return nil, errors.Errorf("error in backend HTTP(s): non-200 OK status code: %s body: \"%s\"", res.Status, buf.String())
 	}
 
 	return &HTTPBackend{rawURL, res.Body}, nil
--- a/pkg/iac/terraform/state/backend/http_reader_test.go
+++ b/pkg/iac/terraform/state/backend/http_reader_test.go
@@ -18,7 +18,7 @@ func TestNewHTTPReader(t *testing.T) {
 		name    string
 		args    args
 		wantURL string
-		wantErr bool
+		wantErr error
 	}{
 		{
 			name: "Should fail with wrong URL",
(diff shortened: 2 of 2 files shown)


pkg/iac/terraform/state/backend/http_reader.go
```diff
@@ -31,6 +33,16 @@ func NewHTTPReader(rawURL string, opts *Options) (*HTTPBackend, error) {
...
 	}
 
+	buf := new(bytes.Buffer)
+	_, err = buf.ReadFrom(res.Body)
+	if err != nil {
+		return nil, err
+	}
+
+	if res.StatusCode < 200 || res.StatusCode >= 400 {
+		return nil, fmt.Errorf("error in backend HTTP(s): non-200 OK status code: %s body: \"%s\"", res.Status, buf.String())
```
wbeuil, line 43: I'm a little confused with this error message. 

```
DEBU[0001] Closing providers
DEBU[0001] Closing provider key=aws
DEBU[0001] Closing gRPC client alias=eu-west-3
error in backend HTTP(s): non-200 OK status code: 401 Unauthorized body: "{"message":"401 Unauthorized"}"
```

Maybe we can display something simpler:

```go
return nil, fmt.Errorf("Error requesting HTTP(s) backend state: \"%s\"", buf.String())

// Error requesting HTTP(s) backend state: "401 Unauthorized"
```

wbeuil (changes requested):

pkg/iac/terraform/state/backend/http_reader.go
eliecharra, line 43: You should use `github.com/pkg/errors` pkg instead of `fmt`. `fmt` is banished from the project since it does not wrap the stack and thus it is very complicated to understand from sentry.

You can use `errors.Errorf()` here

pkg/iac/terraform/state/backend/http_reader.go
```diff
@@ -31,6 +33,16 @@ func NewHTTPReader(rawURL string, opts *Options) (*HTTPBackend, error) {
 		return nil, err
 	}
 
+	buf := new(bytes.Buffer)
+	_, err = buf.ReadFrom(res.Body)
+	if err != nil {
```
eliecharra, line 38: Which kind of errors are covered there ?

pkg/iac/terraform/state/backend/http_reader_test.go
```diff
@@ -42,14 +42,25 @@ func TestNewHTTPReader(t *testing.T) {
...
+					Headers: map[string]string{},
+				},
+			},
+			wantURL: "https://raw.githubusercontent.com/cloudskiff/driftctl-badprojecturl",
+			wantErr: errors.New("error in backend HTTP(s): non-200 OK status code: 404 Not Found body: \"404: Not Found\""),
 		},
 	}
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
 			got, err := NewHTTPReader(tt.args.url, tt.args.options)
```
eliecharra, line 61: Is this test doing a real request on Github side ?

* We should never make an unit test dependent of a third party service.
* It's very unexpected that the query is done when initializing the struct within the `New` method. It makes this code hard to test. For me the request should be done in the `Read()` call. Thus we can inject a mocked http client that will return fake response. Any reason not to do that ? (It can be refactored in another PR)

eliecharra (changes requested):

pkg/iac/terraform/state/backend/http_reader_test.go
sundowndev: We already discussed about that with @moadibfr. I suggested to use https://github.com/h2non/gock to mock the HTTP request but we end up to the conclusion that would make the test become a integration test. Not sure about how we should test this then.

eliecharra: What about creating a `HttpClient` interface and mock the `Do` method ?

pkg/iac/terraform/state/backend/http_reader.go
sundowndev: I followed the code snippet from `pkg/remote/error/errors.go` : 

```
error in supplier github_branch_protection: non-200 OK status code: 401 Unauthorized body: "{\"message\":\"Bad credentials\",\"documentation_url\":\"https://docs.github.com/graphql\"}"
```

Isn't it more productive for the user to get to response body as well? Returning a 400 error without the body will prevent the user from understand the issue.

pkg/iac/terraform/state/backend/http_reader_test.go
sundowndev: You're right, shoud I create another PR for that?

eliecharra: Yeah I think it is not directly related to this one

pkg/iac/terraform/state/backend/http_reader.go
sundowndev: I think the only possible path is when `client.Do(req)` returns a invalid response body object. Should we ignore this error?

eliecharra: I think it may introduce a weakness if we do not handle this case but we should find a way to test this behavior.

Let's close this discussion there and we'll see in your other refacto PR if we can handle that.

pkg/iac/terraform/state/backend/http_reader.go
wbeuil: The part where it says `401 Unauthorized` is the body. It is the first part that I'm not okay with (e.g. `non-200 OK status code: 401), I don't understand a thing ... Either we output only `error ...: status code: 401, body: "<message>"` or we only output the body.

eliecharra (dismissed):

pkg/iac/terraform/state/backend/http_reader.go
wbeuil: @eliecharra WDYT ? What could be a good error message ?

eliecharra: I think it's enough here if we keep a simple message only with the http status code : `Error requesting HTTP(s) backend state: HTTP code 4xx`

For me the body is related to debug purpose, maybe we can add a TRACE log with the response body.

wbeuil: That's way better I agree with Elie.

pkg/iac/terraform/state/backend/http_reader.go
```diff
@@ -31,6 +33,18 @@ func NewHTTPReader(rawURL string, opts *Options) (*HTTPBackend, error) {
...
+		return nil, err
+	}
+
+	logrus.Tracef("HTTP(s) backend response body: %s", buf.String())
```
eliecharra, line 42: Use something like this instead `logrus.WithFields(logrus.Fields{"response": buf.String()}).Trace("backend response body")`. It makes a more readable log record

eliecharra (changes requested):

eliecharra (approved):

wbeuil (approved):

PR #407, opened 2021-04-06
Fix body reader of http backend

sundowndev:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | no
| 🚀 New feature?   | no
| ⚠ Deprecations?   | no
| ❌ BC Break       | no
| 🔗 Related issues | #...
| ❓ Documentation  | no 

## Description

As discussed in #406 , we should mock the http.Client instead of running real HTTP calls on the network.

This change also fixes an issue we recently discovered: the response body is always empty because a buffer was reading the body content already.


wbeuil (dismissed):
LGTM

wbeuil (dismissed):
LGTM, @eliecharra @moadibfr ping on this one.

pkg/iac/terraform/state/backend/http_reader.go
```diff
@@ -17,7 +17,11 @@ type HTTPBackend struct {
 	reader io.ReadCloser
 }
 
-func NewHTTPReader(rawURL string, opts *Options) (*HTTPBackend, error) {
+type HttpClient interface {
```
eliecharra, line 20: This should be put in `test/http.go` with the mock file alongside (mockery default config)

pkg/iac/terraform/state/backend/http_reader.go
```diff
@@ -27,25 +31,23 @@ func NewHTTPReader(rawURL string, opts *Options) (*HTTPBackend, error) {
...
 
-	client := &http.Client{}
 	res, err := client.Do(req)
 	if err != nil {
 		return nil, err
 	}
 
-	buf := new(bytes.Buffer)
-	_, err = buf.ReadFrom(res.Body)
+	body, err := io.ReadAll(res.Body)
```
eliecharra, line 39: `client.Do` and response read should be done in the `Read` func IMHO, it's kinda unexpected to have this kind of logic within a `New...` method.

pkg/iac/terraform/state/backend/http_reader_test.go
```diff
@@ -15,10 +19,11 @@ func TestNewHTTPReader(t *testing.T) {
...
 	}
 	tests := []struct {
-		name    string
-		args    args
-		wantURL string
-		wantErr error
+		name       string
+		args       args
+		wantErr    error
+		httpClient func() HttpClient
```
eliecharra, line 25: You can put `HttpClient` directly and in each case you can do `httpClient: func() HttpClient { return &http.Client{} }()`

This more homogenous with what we do in other tests

eliecharra (changes requested):

pkg/iac/terraform/state/backend/http_reader_test.go
sundowndev: Isn't it what's already done in this PR ?

pkg/iac/terraform/state/backend/http_reader.go
```diff
@@ -27,30 +31,23 @@ func NewHTTPReader(rawURL string, opts *Options) (*HTTPBackend, error) {
...
 }
 
 func (h *HTTPBackend) Read(p []byte) (n int, err error) {
 	if h.reader == nil {
-		return 0, errors.New("Reader not initialized")
+		res, err := h.client.Do(h.request)
+		if err != nil {
+			return 0, err
+		}
+		h.reader = res.Body
```
eliecharra, line 43: I don't think we need the reader to be a struct attribute anymore as we do all the job in the `Read` now

pkg/iac/terraform/state/backend/http_reader_test.go
```diff
@@ -28,52 +32,91 @@ func TestNewHTTPReader(t *testing.T) {
...
+
+				return m
+			}(),
+			expected: "{}",
 		},
 		{
 			name: "Should fail with bad status code",
 			args: args{
-				url: "https://raw.githubusercontent.com/cloudskiff/driftctl-test/main/.dockerignore",
+				url: "https://wrong.url/cloudskiff/driftctl/main/terraform.tfstate",
```
eliecharra, line 74: This is very minor, but you can use `example.com` for this purpose, it is what it has been [designed for](https://www.iana.org/domains/reserved).

pkg/iac/terraform/state/backend/http_reader_test.go
```diff
@@ -107,82 +150,12 @@ func TestHTTPBackend_Close(t *testing.T) {
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
 			h := &HTTPBackend{
-				url:    tt.fields.url,
-				reader: tt.fields.reader(),
+				request: tt.fields.req,
+				reader:  tt.fields.reader(),
```
eliecharra, line 154: Should be `reader:  tt.fields.reader` no ? (without func call `()`)

eliecharra (changes requested):

pkg/iac/terraform/state/backend/http_reader.go
sundowndev: What should be the purpose of Close() method then?

pkg/iac/terraform/state/backend/http_reader_test.go
eliecharra: > But I can turn this field into a io.ReadCloser type instead

Yeah it is the same remark as for the `HttpClient`

pkg/iac/terraform/state/backend/http_reader.go
sundowndev: That'd make the HttpClient become an interface for testing, but it's actually used in the backend package. I don't think we want to use a test interface in our code : `func NewHTTPReader(client test.HttpClient...`, do we?

pkg/iac/terraform/state/backend/http_reader.go
eliecharra: You are right I forgot the `ReadCloser` interface nevermind

pkg/iac/terraform/state/backend/http_reader.go
eliecharra: That's true, the fact is that this generic interface should not be in `pkg/iac/terraform/state/backend/`.

Maybe we can move it to a new `pkg/http` package ?

mocks/http.go
```diff
@@ -0,0 +1,19 @@
+package mocks
+
+import (
+	"net/http"
+
+	mock "github.com/stretchr/testify/mock"
+)
+
+// HttpClient is an autogenerated mock type for the Client type
+type HttpClient struct {
```
eliecharra, line 10: The mock should be in `pkg/http` according to our mockery configuration.

eliecharra (changes requested):

eliecharra (approved):

PR #410, opened 2021-04-07
Fix #409 do not list directory when using s3 as state backend

### LINKED ISSUE #409: directory are listed when we use s3 as a backend for states
**Description**
Sometimes directory are listed when we use s3 as a backend for states.
Reading them produce a "no state" error and stop the scan.

**How to reproduce**
create an empty inside a bucket using aws web console. You may add statefile in it.
Tun scan with this bucket as a backend.

lotoussa:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | yes
| 🚀 New feature?   | no
| ⚠ Deprecations?   | no
| ❌ BC Break       | no
| 🔗 Related issues | #409 
| ❓ Documentation  | no

## Description

- Ignore directory (size <= 0) when listing bucket object
- Add test for empty directory

### DIFF
--- a/pkg/iac/terraform/state/enumerator/s3.go
+++ b/pkg/iac/terraform/state/enumerator/s3.go
@@ -3,6 +3,7 @@ package enumerator
 import (
 	"strings"
 
+	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/s3"
 	"github.com/aws/aws-sdk-go/service/s3/s3iface"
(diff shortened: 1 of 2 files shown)


pkg/iac/terraform/state/enumerator/s3.go
```diff
@@ -45,7 +46,9 @@ func (s *S3Enumerator) Enumerate() ([]string, error) {
 	}
 	err := s.client.ListObjectsV2Pages(input, func(output *s3.ListObjectsV2Output, lastPage bool) bool {
 		for _, metadata := range output.Contents {
-			keys = append(keys, strings.Join([]string{bucket, *metadata.Key}, "/"))
+			if aws.Int64Value(metadata.Size) > 0 {
```
eliecharra, line 49: We should add a test case where this value is `nil`

pkg/iac/terraform/state/enumerator/s3_test.go
```diff
@@ -73,6 +79,48 @@ func TestS3Enumerator_Enumerate(t *testing.T) {
 				"bucket-name/a/nested/prefix/state6",
 			},
 		},
+        {
+            name: "test results are not returned",
```
eliecharra, line 83: `test that directories objects are filtered`

eliecharra (changes requested):

eliecharra (approved):

Issue #219, opened 2021-02-08
ACC test should be independent from the local environment.


sshota0809:
This is related to #119.

Currently ACC test is dependent on the local envirnment. it's executed with local version of terraform. That means local version of terrafrom should be supported by driftctl. That might bother developers.

Once ACC test is launched, terraform should be installed automatically as temporary for ACC test.
 
In #119, I mentioned about this as following.

> 2. Install terraform for ACC test through terraform-exec and be able to execute ACC test being independent from the developer's local.

eliecharra:
@sjourdan Nothing to test here for you, closing

PR #416, opened 2021-04-09
Add dynamic Terraform versions to acceptance tests

sundowndev:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | no
| 🚀 New feature?   | no
| ⚠ Deprecations?   | no
| ❌ BC Break       | no
| 🔗 Related issues | #219
| ❓ Documentation  | no 

## Description

> Currently ACC test is dependent on the local envirnment. it's executed with local version of terraform. That means local version of terrafrom should be supported by driftctl. That might bother developers. Once ACC test is launched, terraform should be installed automatically as temporary for ACC test.

This patch assume every acceptance test have their own Terraform version. By specifying the desired TF version, the acceptance system will install the needed TF version in a temporary directory. That directory will be accessible to all tests so we avoid installing same versions twice.


lotoussa (dismissed):
LGTM, 
the ticket was talking about terraform-exec, you didn't use it ?

sundowndev:
> LGTM,
> the ticket was talking about terraform-exec, you didn't use it ?

We were already using it, but with the local version of Terraform. This patch assume we now have a dedicated Terraform version for each test.

wbeuil (approved):
LGTM

eliecharra (approved):

Issue #415, opened 2021-04-09
.driftignore generator


sjourdan:
**Description**

As a new driftctl user
I want to generate a .driftignore file containing all my current drifts
So I can start fresh and get information about new drifts quickly, and fix my exiting ones later_

Goal: reformat the JSON output to a `type.name` schema that fits the `.driftignore`. 

## Mockup 2021/5/18

**User run 1:** 

This is the first driftctl run for the user, likely on a laptop or an interactive terminal. The user likely also already spent time finding the right environment variables, or tfstate locations.

This user discovers many drifts through the human-readable output of the CLI. 

When drift is detected, driftctl tells the user how to generate a .driftignore file so he can go to run 2.

```shell
$ driftctl scan --from tfstate+s3://mybucket/ 
Scanned states (1) 
Found resources not covered by IaC:
  aws_security_group:
    - sg-04bc6391da4b53075
[...]
Found 16 resource(s)
 - 18% coverage
 - 3 covered by IaC
 - 13 not covered by IaC
 - 0 missing on cloud provider
 - 0/3 changed outside of IaC
You have unmanaged security group rules that could be false positives, find out more at https://docs.driftctl.com/limitations

Hint: use driftctl gen-driftignore to generate a .driftignore file 
exit status 1
```

**User run 2:** 

In this run, the user will launch the command run 1 mentioned - and driftctl will show how to use it:

```shell
$ driftctl gen-driftignore
Error: to generate a .driftignore file, use driftctl gen-driftignore --from <drifts.json>
Generate a JSON using driftctl scan --output json://path/to/drifts.json
```

**User run 3:** 

In this run, the user will run the same command as the first, but with the JSON output as indicated in run 2.

```shell
$ driftctl scan --output json://drifts.json
Scanned states (1) 
Scanned resources  
$ file drifts.json 
drifts.json: JSON data
```

**User run 4:** 

In this run, the user finally generates the .driftignore file from the JSON output generated in run 3.

```shell 
$ driftctl gen-driftignore --from drifts.json 
# Resources not covered by IaC resources
aws_s3_bucket.example-1
aws_iam_access_key.AKIASXXX
[...]

$ driftctl gen-driftignore --from drifts.json > .driftignore

$ cat .driftignore
# Resources not covered by IaC resources
aws_s3_bucket.example-1
aws_iam_access_key.AKIASXXX
[...]
```

**User run 5:**

Final run: the user runs driftctl with an initial .driftignore file so he can get started.

[...]

PR #433, opened 2021-04-13
Create driftignore generator command

sundowndev:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | no
| 🚀 New feature?   | yes
| ⚠ Deprecations?   | no
| ❌ BC Break       | no
| 🔗 Related issues | #415
| ❓ Documentation  | yes 

TL;DR: The purpose of this change is to be able to generate a .driftignore file containing current drifts.

## Context

New users can struggle when trying the tool because the output is too large and hard to read. They might want to ignore all the current drifts so they can start introduce the tool in their workflow quickly and fix the current issues later. This feature allows them to create a driftignore file from a JSON output.

## Flags

- `--exclude-unmanaged` allows to control whether unmanaged resources are included or not
- `--exclude-missing` allows to control whether missing resources are included or not
- `--exclude-changed` allows to control whether changed resources are included or not

## Examples

```shell
# Ignore all current drifts
$ driftctl scan -o json://stdout | driftctl gen-driftignore -f /dev/stdin > .driftignore

# Changed resources will not be included
$ driftctl scan --from tfstate://state1.tfstate -o json://stdout | driftctl gen-driftignore --exclude-changed

# Unmanaged resources will not be included
$ cat result.json | driftctl gen-driftignore --exclude-unmanaged
```


eliecharra:
To rebase on `v0.8` and change target branch

lotoussa (dismissed):
LGTM

pkg/cmd/gen_driftignore.go
```diff
@@ -0,0 +1,69 @@
...
+				return errors.New("An existing .driftignore file was found. Please rename or delete the existing one to generate a new one.")
+			}
+
+			input, err := io.ReadAll(bufio.NewReader(os.Stdin))
+			if err != nil {
+				return err
+			}
+
+			analysis := &analyser.Analysis{}
+			err = analysis.UnmarshalJSON(input)
```
eliecharra, line 37: It is better to use `json.Unmarshall(input, analysis)` than calling Unmarshaller interface method directly

pkg/cmd/gen_driftignore.go
```diff
@@ -0,0 +1,69 @@
...
+			if err != nil {
+				return err
+			}
+
+			// Sort resources in a predictable order
+			analysis.SortResources()
```
eliecharra, line 43: Is it required ? As we sort the analysis results in the scan the order is preserved during the serialization / deserialization process I think.

pkg/cmd/gen_driftignore.go
```diff
@@ -0,0 +1,69 @@
...
+
+			fmt.Printf("Added %v resources to .driftignore\n", n)
+
+			return nil
+		},
+	}
+
+	fl := cmd.Flags()
+
+	fl.BoolVar(&opts.IncludeUnmanaged, "unmanaged", true, "Include resources not managed by IaC")
```
eliecharra, line 64: `--without-changed` maye be better from the end user point of view WDYT ?

`driftctl gen-driftignore --without-changed --without-unmanaged`

Passing `--changed=false`  looks strange as true is the default you will never pass `--changed=true` 

Same remark for other flags

pkg/cmd/gen_driftignore.go
```diff
@@ -0,0 +1,69 @@
...
+			analysis := &analyser.Analysis{}
+			err = analysis.UnmarshalJSON(input)
+			if err != nil {
+				return err
+			}
+
+			// Sort resources in a predictable order
+			analysis.SortResources()
+
+			f, err := os.OpenFile(filter.DriftIgnoreFilename, os.O_CREATE|os.O_RDWR|os.O_TRUNC, 0600)
```
eliecharra, line 45: `defer f.Close()` after the error check

pkg/filter/driftignore.go
```diff
@@ -143,6 +152,51 @@ RuleCheck:
...
 }
 
+func AnalysisToList(analysis *analyser.Analysis, opts AnalysisListOptions) (int, string) {
+	var list []string
+
+	n := 0
+
+	addResources := func(res ...resource.Resource) {
+		for _, r := range res {
+			list = append(list, fmt.Sprintf("%s.%s", escapeKey(r.TerraformType()), escapeKey(r.TerraformId())))
```
eliecharra, line 162: Escaping the type is not necessary

pkg/filter/driftignore.go
```diff
@@ -143,6 +152,51 @@ RuleCheck:
...
+		list = append(list, "# Changed resources")
+		addDifferences(analysis.Differences()...)
+		n += analysis.Summary().TotalUnmanaged
+	}
+
+	return n, strings.Join(list, "\n")
+}
+
+func escapeKey(line string) string {
+	line = strings.ReplaceAll(line, "\\", `\\`)
```
eliecharra, line 194: You can use backticks everywhere there

pkg/filter/driftignore.go
```diff
@@ -143,6 +152,51 @@ RuleCheck:
 	return false
 }
 
+func AnalysisToList(analysis *analyser.Analysis, opts AnalysisListOptions) (int, string) {
```
eliecharra, line 155: Why putting this method there ? It's related to `Analysis` and maybe we can make it an Analysis struct method like `analysis.ToDriftIgnoreList()`

You can see it as another way to serialize an analysis struct. It should not remain in the filter pkg for me.

eliecharra (changes requested):
Maybe it could be usefull to introduce a `--force` flag to overwrite an existing `.driftignore` file too.

pkg/cmd/gen_driftignore.go
sundowndev: Why not, but may be the exclude keyword is more explicit : `--exclude-changed`

pkg/cmd/testdata/output_stdin_valid.txt
```diff
@@ -0,0 +1,18 @@
+# Ignoring 14 resources
+# Resources not covered by IaC
+aws_iam_user.driftctl
+aws_iam_user.sundowndev
+aws_iam_user.test_user
+aws_iam_role_policy.OrganizationAccountAccessRole:AdministratorAccess
+aws_iam_role_policy.driftctl_assume_role:driftctl_policy\.10
+aws_iam_role.OrganizationAccountAccessRole
+aws_iam_role.driftctl_assume\\_role
```
eliecharra, line 9: This is unexpected, the original id in the input is `driftctl_assume\\_role` and thus the escaped ignored string should be `driftctl_assume\\\\_role`

eliecharra (changes requested):

pkg/cmd/testdata/output_stdin_valid.txt
sundowndev: > This is unexpected, the original id in the input is driftctl_assume\\_role and thus the escaped ignored string should be driftctl_assume\\\\_role

The input is `driftctl_assume\\_role` but is serialized as `driftctl_assume\_role`, so the escaped ignored string is correct: `driftctl_assume\\_role`.

The JSON serializer doesn't accept `driftctl_assume\_role` because it means we try to escape `_` and produce the following error : `invalid character '_' in string escape code`

eliecharra: 🆗 👍🏻

eliecharra (approved):
Great job 👍🏻

PR #435, opened 2021-04-14
Add a extra test case for the strict mode

sundowndev:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | no
| 🚀 New feature?   | no
| ⚠ Deprecations?   | no
| ❌ BC Break       | no
| 🔗 Related issues | #...
| ❓ Documentation  | no 

## Description

As discussed in https://github.com/cloudskiff/driftctl/pull/429#discussion_r612542039, an extra test case was needed for the strict mode.


pkg/driftctl_test.go
```diff
@@ -49,24 +48,13 @@ func runTest(t *testing.T, cases TestCases) {
...
 			remoteSupplier.On("Resources").Return(c.remoteResources, nil)
 
-			var filter *jmespath.JMESPath
-			if c.filter != "" {
-				f, err := filter2.BuildExpression(c.filter)
-				if err != nil {
-					t.Fatalf("Unable to build filter expression: %s\n%s", c.filter, err)
-				}
-				filter = f
-			}
```
wbeuil, line 59: Given the verbosity to rewrite in each test case the filtering option, I wonder if it wouldn't be better to just add the boolean for the strict mode instead.

wbeuil (changes requested):
Can you add a test with strict mode AND filtering enabled

pkg/driftctl_test.go
sundowndev: Yes it would be less verbose but it adds complexity when it comes to use other ScanOption fields, isn't it? I think this approach gives more control on what we want to test without modifying the test case struct.

wbeuil (dismissed):

PR #449, opened 2021-04-19
Enforce one occurrence of filter flag

sundowndev:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | yes
| 🚀 New feature?   | no
| ⚠ Deprecations?   | no
| ❌ BC Break       | no
| 🔗 Related issues | https://github.com/cloudskiff/driftctl/issues/440
| ❓ Documentation  | no 

## Description

> As previously discussed there cloudskiff/driftctl-docs#10 (comment), we should not allow to pass multiples --filter flags, it should trigger an error.

We still have to decide if we really want to fix this issue on our side, or in the pflag project, which already has an issue opened about this (https://github.com/spf13/pflag/issues/72).


wbeuil (approved):
I'm good with that one

wbeuil (approved):

lotoussa (approved):

Issue #434, opened 2021-04-14
Support for Terraform Enterprise/Cloud


kmai:
**Description**
Currently, you can retrieve the `terraform.tfstate` file from S3, which is great. But what if we use Terraform Enterprise or Terraform Cloud? I haven't found any issues related to this feature, or documentation. Not even a mention in the `driftctl` help.

**Example**
For example, we're currently supporting:

* `tfstate://`: Local
* `tfstate+s3://`: S3
* `tfstate+http://`: HTTP Endpoint
* `tfstate+https://`: HTTPS Endpoint

This feature request would also implement something like `tfstate+tfcloud://` and/or `tfstate+enterprise://`.

wbeuil:
Hello @kmai, thanks for opening this issue.

I only tested with terraform cloud, thus I can't talk about terraform enterprise but for now you can use driftctl like this:

```
# First, use Terraform API to request the hosted-state-download-url
$ curl \
  --header "Authorization: Bearer $TOKEN" \
  --header "Content-Type: application/vnd.api+json" \
https://app.terraform.io/api/v2/workspaces/$WORKSPACE_ID/current-state-version

# Then, use driftctl with the https endpoint
$ driftctl scan --from tfstate+https://<hosted-state-download-url>
```

Would it fit your use case at this moment ?

kmai:
Hey! Thanks for the follow up. 

I tried that approach and it works, but it'd be amazing to have the support for it a bit more transparent. It is a bit more hacky than expected and it's not documented either.

I think that this and https://github.com/cloudskiff/driftctl/issues/53 are on my wishlist, but so far, this is a really cool tool.

I can't wait to implement it!

eliecharra:
Hey @ayshiff we are always open to contributions 🙏🏻 You can join us on discord our #dev channel if you want additional details and if you want to discuss about implementation details

nikolay:
How will this works if you have your infrastructure broken into multiple TFC workspaces?

sundowndev:
> How will this works if you have your infrastructure broken into multiple TFC workspaces?

@nikolay You can specify multiple state files for a single scan, as long as you can have a single TFC token for multiple workspaces. I'm not a TFC user so may be I miss something, but it should be possible like so: 

```shell
$ TFC_TOKEN=xxx \
  driftctl scan --from tfstate+tfcloud://<workspace_id> --from tfstate+tfcloud://<workspace_id>
```

sjourdan:
[...]

PR #458, opened 2021-04-25
Support for Terraform Cloud / Terraform Enterprise

ayshiff:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | no
| 🚀 New feature?   | yes
| ⚠ Deprecations?   | yes/no
| ❌ BC Break       | no
| 🔗 Related issues | #434
| ❓ Documentation  | yes https://github.com/cloudskiff/driftctl-docs/pull/47

## Description

This PR adds support for Terraform Cloud / Terraform Enterprise.

As described in the issue the logic is as follow:

1. Fetch `hosted-state-download-url` from the API with the provided `$WORKSPACE_ID` (`tfstate+tfcloud://$WORKSPACE_ID`) and the API token through the headers (`--headers 'Authorization=Bearer $API_TOKEN'`)
2. Use `HTTPReader` with  the retrieved `hosted-state-download-url`

Note that the logic is the same for Terraform Cloud and Terraform Enterprise as the API is the same for both.

Here is a complete command example:

```bash
driftctl scan --from tfstate+tfcloud://$WORKSPACE_ID --headers 'Authorization=Bearer $API_TOKEN'
```
I also added some test the different cases we can have:
- The provided workspace_id is wrong (`404`)
- The provided authentication api_token is wrong (`401`)
- The workspace_id and api_token are good

I have as well updated the documentation in https://github.com/cloudskiff/driftctl-docs/pull/47 to fit this new state reader.


pkg/iac/terraform/state/backend/cloud_reader.go
```diff
@@ -0,0 +1,65 @@
...
+	}
+
+	client := &http.Client{}
+	res, err := client.Do(req)
+
+	if res.StatusCode == 404 {
+		return nil, errors.Errorf("Error reading state from Terraform Cloud/Enterprise workspace: wrong workspace id")
+	}
+
+	if res.StatusCode == 401 {
```
sundowndev, line 44: I think we should catch every non-success HTTP status code and throw an generic error, like we do in http backend.

pkg/iac/terraform/state/backend/cloud_reader.go
```diff
@@ -0,0 +1,65 @@
...
+	}
+
+	bodyBytes, _ := ioutil.ReadAll(res.Body)
+
+	body := Body{}
+	err = json.Unmarshal(bodyBytes, &body)
+
+	if err != nil {
+		fmt.Println("error:", err)
+		panic(err.Error())
```
sundowndev, line 59: May be a mistake? We shouldn't panic, never.

pkg/iac/terraform/state/backend/cloud_reader.go
```diff
@@ -0,0 +1,65 @@
...
+
+type Attributes struct {
+	HostedStateDownloadUrl string `json:"hosted-state-download-url"`
+}
+
+type Data struct {
+	Attributes Attributes `json:"attributes"`
+}
+
+type Body struct {
```
sundowndev, line 23: Since we're in a common "backend" package that is not dedicated to tfcloud, I think we should give more explicit names to types like `TFCloudBody`

pkg/iac/terraform/state/backend/cloud_reader_test.go
```diff
@@ -0,0 +1,91 @@
...
+			wantErr:   errors.New("Error reading state from Terraform Cloud/Enterprise workspace: bad authentication token"),
+			responder: httpmock.NewBytesResponder(http.StatusUnauthorized, []byte{}),
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			httpmock.Reset()
+			httpmock.RegisterResponder("GET", tt.url, tt.responder)
+			if tt.name == "Should fetch URL with auth header" {
+				httpmock.RegisterResponder("GET", "https://archivist.terraform.io/v1/object/test", httpmock.NewBytesResponder(http.StatusOK, []byte(`{}`)))
```
sundowndev, line 78: I think you can group those httpmock directives to a method in the testcase, instead of filtering testcases like this.

example: `mock: func() { httpmock.Reset() ... }`

wbeuil:
Thanks @ayshiff for your contribution. You did great on a first step to support TF Cloud. But we should converge to what they do in Terraform for our user to use the support quite the same way. 

You can find all the information for the credentials part [here](https://www.terraform.io/docs/cli/commands/login.html). They rely on their `.terraformrc` file which can be created with `terraform login` which is simply:

```json
credentials "app.terraform.io" {
  token = "XXX"
}
```

We should, thus, support 2 ways:
- Get the token from `.terraformrc` (maybe in another filename as well)
- Inline token from CLI (what you did but would rather write simply --token or something else)

I didn't go further into terraform package but we could re-use one of their functions/methods to fetch the credentials from the file. Maybe they even have a function/method the fetch the state with the token and the `workspace_id`.

sundowndev (changes requested):
Great job! 👍🏽 

Left few comments about error handling and tests. We also might discuss a bit about authentication, as there are multiple ways to authenticate to TFC. We usually don't perform integration tests for backends, so we have to discuss about that too.

ayshiff:
I fixed the various issues that came up and added a new `--terraform-cloud-token` flag to provide the Terraform Cloud API token.

I still have to work on the way to retrieve the token from `.terraformrc` as I didn't have the time to look at it in detail.

eliecharra:
> I still have to work on the way to retrieve the token from .terraformrc as I didn't have the time to look at it in detail.

@wbeuil @sundowndev Maybe this could be splitted in another PR ? I'm fine with that

sundowndev:
> > I still have to work on the way to retrieve the token from .terraformrc as I didn't have the time to look at it in detail.
> 
> @wbeuil @sundowndev Maybe this could be splitted in another PR ? I'm fine with that

Well I don't really know much about how credentials management work in TFC, but yeah we could introduce this feature with the flag only for now.

wbeuil:
We could indeed split this PR into two.

Just for the record, there is this [method](https://github.com/hashicorp/terraform/blob/main/command/cliconfig/cliconfig.go#L87) which seems to read and load everything there is inside a config file (`.terraformrc` in linux, `terraform.rc` in windows).

pkg/iac/terraform/state/backend/cloud_reader.go
```diff
@@ -0,0 +1,63 @@
...
+
+	body := TFCloudBody{}
+	err = json.Unmarshal(bodyBytes, &body)
+
+	if err != nil {
+		return nil, err
+	}
+
+	rawURL := body.Data.Attributes.HostedStateDownloadUrl
+	logrus.WithFields(logrus.Fields{"hosted-state-download-url": rawURL}).Trace("Cloud backend response")
```
sundowndev, line 59: Terraform Cloud backend response

pkg/iac/terraform/state/backend/cloud_reader.go
```diff
@@ -0,0 +1,63 @@
...
+
+func NewCloudReader(workspaceId string, opts *Options) (*HTTPBackend, error) {
+	req, err := http.NewRequest(http.MethodGet, fmt.Sprintf("%s/workspaces/%s/current-state-version", TerraformCloudAPI, workspaceId), nil)
+
+	req.Header.Add("Content-Type", "application/vnd.api+json")
+	req.Header.Add("Authorization", fmt.Sprintf("Bearer %s", opts.TerraformCloudToken))
+
+	if err != nil {
+		return nil, err
+	}
```
sundowndev, line 36: Isn't `req` going to be `nil` if there's an error here? Error handling should be right next to the expression IMO

pkg/cmd/scan_test.go
```diff
@@ -39,6 +39,7 @@ func TestScanCmd_Valid(t *testing.T) {
 		{args: []string{"scan", "-t", "aws+tf", "-f", "tfstate://test"}},
 		{args: []string{"scan", "--to", "aws+tf", "--from", "tfstate://test"}},
 		{args: []string{"scan", "--to", "aws+tf", "--from", "tfstate+https://github.com/state.tfstate"}},
+		{args: []string{"scan", "--to", "aws+tf", "--from", "tfstate+tfcloud://workspace_id"}},
```
sundowndev, line 42: Can you add a test case for the token flag as well?

sundowndev (changes requested):
This seems great, I just left some comments about the code

pkg/cmd/scan.go
```diff
@@ -128,6 +128,13 @@ func NewScanCmd() *cobra.Command {
 		"Use those HTTP headers to query the provided URL.\n"+
 			"Only used with tfstate+http(s) backend for now.\n",
 	)
+	fl.StringVarP(&opts.BackendOptions.TerraformCloudToken,
+		"terraform-cloud-token",
```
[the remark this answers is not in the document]
nikolay: The standard abbreviation is `TFC`, so, `TFCLOUD` is coining a new one. I think it would be best to use what HashiCorp uses themselves, i.e. 'TF_API_TOKEN`.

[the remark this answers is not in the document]
sundowndev: Thanks for pointing that out @nikolay, you're right. Also [in the docs](https://docs.driftctl.com/next/usage/cmd/scan-usage#http--terraform-cloud) we were already using the `TFC_TOKEN` env var for Terraform Cloud.

wbeuil (changes requested):
Small comments to be fully consistent with the new wording. Can you also update your branch to have the latest changes from @sundowndev ? Thanks !

sundowndev (approved):

wbeuil (approved):
LGTM !

eliecharra:
Hey @all-contributors please add @ayshiff for code

Issue #408, opened 2021-04-06
Listing states from S3 can take time and make the progressbar timeout


moadibfr:
**Description**
Listing states from S3 can take time and make the progressbar timeout

**How to reproduce**
Fill a bucket with a lot a files and run scan with this bucket in --from

**Possible Solution**
We should introduce a new progressbar for state enumeration count and one for resource enumeration from states files

eliecharra:
@moadibfr I was wondering if the progressbar should be restarted automatically after a timeout if a new event occur. Sometime read actions may be slow but ends with a success after the timeout, for me we should show that something hang bit if it's ends with a success we should resume the spinning. WDYT @wbeuil we talked about it weeks ago in a pair session if I remember correctly 🤔 

For example :

```
State: Spinning
<-- No new events
State: Hanged after N seconds
<-- New event occur after the timeout
State: Spinning 
etc ...

```

wbeuil:
I had the problem when testing driftctl with my Cloudflare DNS (thanks Orange). And it is indeed misleading to see that I hit the timeout, my requests succeed afterwards and the progress was not restarted. I agree with @eliecharra

moadibfr:
If this happen we should set the timeout to a higher value. Timeout should happen if there is nothing happening.
I think the idea is to tell the user nothing is happening and he can quit so it should be an rare exception if this is not a real hang.

But yeah I think it could be a good idea to allow progressbar to get event even when timeouted and restart the render. So when we have this exception if the user wait enough it will still work.

eliecharra:
> If this happen we should set the timeout to a higher value.

It's gonna be a never ending battle. Timeout can be relative to user setup (xDSL with latency line, issues with DNS like @wbeuil , etc ...)

> I think the idea is to tell the user nothing is happening and he can quit so it should be an rare exception if this is not a real hang.

I found this is very weak as we cannot predict any network latency or other issues that can slow down the process but does not mean there is an hang. I think it should be the user responsibility to take the decision to interrupt, our actual timeout should be there for UI/UX purpose only.

moadibfr:
> It's gonna be a never ending battle. Timeout can be relative to user setup (xDSL with latency line, issues with DNS like @wbeuil , etc ...)
yeah but we should find a timeout value that make it as rare as possible

[...]

PR #460, opened 2021-04-26
Introduce a new progress bar for state enumeration

sundowndev:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | yes
| 🚀 New feature?   | no
| ⚠ Deprecations?   | no
| ❌ BC Break       | no
| 🔗 Related issues | #408
| ❓ Documentation  | no 

## Description

> Listing states from S3 can take time and make the progressbar timeout. We should introduce a new progressbar for state enumeration count and one for resource enumeration from states files.

This change updates the actual progress interface and introduce a new progress bar for state enumeration.


pkg/output/progress.go
```diff
@@ -98,3 +121,19 @@ Loop:
...
+
+func (p *progress) flush() {
+	for i := 0; i < p.highestLineLength; i++ {
+		Printf(" ")
+	}
+	Printf("\r")
+}
+
+func (p *progress) printf(format string, args ...interface{}) {
+	length := len(fmt.Sprintf(format, args...))
```
moadibfr, line 133: why don't you save the sprintf result to print it at the end instead of doing it 2 times ?

pkg/driftctl.go
```diff
@@ -137,6 +140,8 @@ func (d DriftCTL) scan() (remoteResources []resource.Resource, resourcesFromStat
 		return nil, nil, err
 	}
 
+	d.progress.Start()
+
 	logrus.Info("Start scanning cloud provider")
 	remoteResources, err = d.remoteSupplier.Resources()
 	if err != nil {
```
moadibfr, line 147: think y ou should stop here

pkg/iac/supplier/supplier.go
```diff
@@ -28,7 +29,9 @@ func IsSupplierSupported(supplierKey string) bool {
 	return false
 }
 
-func GetIACSupplier(configs []config.SupplierConfig, library *terraform.ProviderLibrary, backendOpts *backend.Options) (resource.Supplier, error) {
+func GetIACSupplier(configs []config.SupplierConfig, library *terraform.ProviderLibrary, backendOpts *backend.Options, progress output.Progress) (resource.Supplier, error) {
+	progress.Start()
```
moadibfr, line 33: shouldn't you stop in the same function ?

moadibfr (changes requested):

pkg/iac/supplier/supplier.go
sundowndev: I think it's the Resources() method that performs the state enumeration, so the progress stops there

moadibfr: It's really confusing to know when the progess is started/stopper if it's not around the listing call.
Maybe it would be easier to start both bars in driftctl.go scan ?

pkg/driftctl.go
```diff
@@ -132,12 +137,16 @@ func (d DriftCTL) Stop() {
 
 func (d DriftCTL) scan() (remoteResources []resource.Resource, resourcesFromState []resource.Resource, err error) {
 	logrus.Info("Start reading IaC")
+	d.iacProgress.Start()
 	resourcesFromState, err = d.iacSupplier.Resources()
 	if err != nil {
```
moadibfr, line 142: think you have to stop before the return here ?

moadibfr (approved):

PR #464, opened 2021-04-27
Fix version check lambda version bump

eliecharra:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | no
| 🚀 New feature?   | no
| ⚠ Deprecations?   | no
| ❌ BC Break       | no
| 🔗 Related issues | none
| ❓ Documentation  | no

## Description

Our lambda has now new env var handler outside driftctl scope.
This PR allow us to update only a single env variable in a given lambda function.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,6 +1,5 @@
 version: 2.1
 orbs:
-  aws-cli: circleci/aws-cli@1.3.0
   go: circleci/go@1.5.0
 jobs:
   build:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -126,15 +125,18 @@ jobs:
...
+    docker:
+        - image: cimg/base:2021.04
     steps:
-      - aws-cli/install
       - run:
           name: "Update Lambda version"
           command: |
-            aws lambda update-function-configuration --function-name $FUNCTION_NAME --environment "{\"Variables\":{\"LATEST_VERSION\":\"$CIRCLE_TAG\"}}"
+              wget "https://github.com/cloudskiff/lambda-env-updater/releases/download/v1.0.0/lambda-env-updater_linux_amd64" && chmod +x lambda-env-updater_linux_amd64
+              ./lambda-env-updater_linux_amd64\
```
sundowndev, line 137: `./lambda-env-updater_linux_amd64 \`

eliecharra: Why adding an additional space there ?

.circleci/config.yml
```diff
@@ -126,15 +125,18 @@ jobs:
...
+                -name ${FUNCTION_NAME}\
```
sundowndev, line 138: `-name ${FUNCTION_NAME} \`

sundowndev (changes requested):
Two small typos in the command

.circleci/config.yml
sundowndev: Because otherwise it makes the shell asking for an input (even with bash)
```
 $ ./bin/driftctl_linux_amd64\
> 
```

sundowndev: Well in your diff there's a space in the next line so it'll work anyway, nevermind 😄

sundowndev (approved):

Issue #442, opened 2021-04-16
Add execution time to telemetry


eliecharra:
**Description**
Following #441, we want to add the execution duration (in seconds) to telemetry.
It should include state enumeration, resource enumeration from the states, and remote enumeration.

sjourdan:
LGTM, definitely my test case

![image](https://user-images.githubusercontent.com/419121/117498975-d3402780-af7a-11eb-99bd-ff4ccd5c86dc.png)

PR #467, opened 2021-04-28
Telemetry duration

eliecharra:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | no
| 🚀 New feature?   | no
| ⚠ Deprecations?   | no
| ❌ BC Break       | no
| 🔗 Related issues | #442 
| ❓ Documentation  | yes


pkg/telemetry/telemetry.go
```diff
@@ -31,6 +32,7 @@ func SendTelemetry(analysis *analyser.Analysis) {
 		Arch:           runtime.GOARCH,
 		TotalResources: analysis.Summary().TotalResources,
 		TotalManaged:   analysis.Summary().TotalManaged,
+		Duration:       uint(analysis.Duration.Seconds() + 0.5),
```
sundowndev, line 35: Why not using the float64 type here ?

eliecharra: On the backend we expect to retrieve an `uint`

wbeuil: Why adding 0.5 here ?

eliecharra: This is an elegant way to round to the nearest integer. `1.4` will become `1.9` and casted to `uint 1` and `1.5` will become `2.0` casted to `uint 2`

wbeuil (approved):

sundowndev (approved):

PR #471, opened 2021-04-29
Refacto `aws_s3_bucket`

eliecharra:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | no
| 🚀 New feature?   | no
| ⚠ Deprecations?   | no
| ❌ BC Break       | no
| 🔗 Related issues | #347 
| ❓ Documentation  | no


pkg/resource/aws/aws_s3_bucket_test.go
```diff
@@ -21,8 +21,9 @@ func TestAcc_AwsS3Bucket_BucketInUsEast1(t *testing.T) {
 					if err != nil {
 						t.Fatal(err)
 					}
-					result.AssertManagedCount(1)
+					result.AssertManagedCount(2)
```
wbeuil, line 24: The filter is on type `aws_s3_bucket`, I count 3 bucket resources in your `s3.tf`. I don't understand why we assert on 2 here.

pkg/middlewares/aws_bucket_policy_expander.go
```diff
@@ -27,11 +27,14 @@ func (m AwsBucketPolicyExpander) Execute(_, resourcesFromState *[]resource.Resou
 			continue
 		}
 
-		bucket, _ := res.(*aws.AwsS3Bucket)
+		bucket, ok := res.(*resource.AbstractResource)
+		if !ok {
+			return nil
+		}
```
wbeuil, line 33: Not sure we should return `nil` here. What could be the reason here to not have an `AbstractResource` ?

moadibfr (approved):
LGTM

pkg/middlewares/aws_bucket_policy_expander.go
```diff
@@ -44,25 +47,29 @@ func (m AwsBucketPolicyExpander) Execute(_, resourcesFromState *[]resource.Resou
...
 
+	bucketAttr, _ := bucket.Attrs.Get("bucket")
+	bucketName := bucketAttr.(string)
 	data := map[string]interface{}{
-		"id":     bucket.Id,
-		"bucket": bucket.Bucket,
-		"policy": bucket.Policy,
+		"id":     bucket.TerraformId(),
+		"bucket": bucketName,
+		"policy": policy,
```
wbeuil, line 62: We could totally write it that way since we won't need bucketName or policy var anymore when `S3BucketPolicy` will be refactored.

```
data := map[string]interface{}{
		"id":     bucket.TerraformId(),
		"bucket": (*bucket.Attrs)["bucket"],
		"policy": policyAttr,
	}
```

pkg/middlewares/s3_bucket_acl_test.go
```diff
@@ -25,125 +25,174 @@ func TestS3BucketAcl_Execute(t *testing.T) {
...
+						Id:   "testgrant",
+						Type: aws.AwsS3BucketResourceType,
+						Attrs: &resource.Attributes{
+							"grant": []map[string]interface{}{
+								{
+									"id":          "356616ba70ebbea29732c95eef24f9ea326b9018c167651705348b5af406a6db",
+									"permissions": []string{"FULL_CONTROL"},
+									"type":        awssdk.String("CanonicalUser"),
+									"uri":         awssdk.String(""),
+								},
```
wbeuil, line 38: You can delete all the pointers for the attributes of each `grant` we won't need it anymore. Same for the other tests in this file.

wbeuil (changes requested):

pkg/resource/aws/aws_s3_bucket_test.go
eliecharra: Nice catch, I changed the tf file but forgot to run the test another time ...

wbeuil (approved):

Issue #419, opened 2021-04-12
Display instance name when tag name exist


eliecharra:
**Description**
We should display the instance name for `aws_ec2_instance` and other resource that use tags for a name.

**References**:
- https://d1.awsstatic.com/whitepapers/aws-tagging-best-practices.pdf (page 13)

sjourdan:
LGTM!
Not sure it's extensive, but the `aws_instance` works indeed.
```
Found resources not covered by IaC:
[...]
  aws_instance:
    - i-0a85fc94dad0eee9a (Name: aws-www-1)
[...]
```

Issue #439, opened 2021-04-15
clarify how to add some resources to the .driftignore file


sjourdan:
**Description**

Take this simple driftctl output: 

```shell 
$ driftctl scan 
Scanned resources:    (55)
Found resources not covered by IaC:
  aws_iam_user:
    - terraform
  aws_security_group_rule:
    - Type: ingress, SecurityGroup: sg-0ce251e7ce328547d, Protocol: tcp, Ports: 5432, Source: 96.202.220.106/32
  aws_sqs_queue:
    - https://sqs.us-east-1.amazonaws.com/141177182257/queue-6qaidu.fifo
  aws_lambda_event_source_mapping:
    - Source: arn:aws:sqs:us-east-1:141177182257:queue-6qaidu.fifo Dest: arn:aws:lambda:us-east-1:141177182257:function:lk6n03-lambda-demo

```

Let's say I want to add all those resources to the `.driftignore`. 

It's easy to infer from the output that adding `aws_iam_user.terraform` to the `.driftignore` file will work (and it does)

For the `aws_sqs_queue`, it's quite straightforward reading the [related documentation](https://docs.driftctl.com/0.7.1/usage/filtering/driftignore): (`aws_sqs_queue.https://sqs\.us-east-1\.amazonaws\.com/141177182257/queue-6qaidu\.fifo`) 

But what is the user supposed to ignore with the `aws_security_group_rule` or `aws_lambda_event_source_mapping` output?

**Environment**
 - OS: fedora 33
 - driftctl version: 0.7.1

**How to reproduce**

**Possible Solution**

**Additional context**

eliecharra:
We should always display IDs for every resource for ignore purpose and document this

sjourdan:
LGTM! Timely to test and confirm IDs from another issue:

```
Found missing resources:
  aws_kms_alias:
    - alias/key-alias
  aws_kms_key:
    - e492d07f-0201-4346-a33b-727454d0787b
```

PR #473, opened 2021-04-29
Display ec2 instance name when tag name exist

sundowndev:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | no
| 🚀 New feature?   | yes
| ⚠ Deprecations?   | no
| ❌ BC Break       | no
| 🔗 Related issues | #419
| ❓ Documentation  | no 

## Description

>  We should display the instance name for aws_ec2_instance and other resource that use tags for a name.

### DIFF
--- a/pkg/resource/aws/aws_instance_ext.go
+++ b/pkg/resource/aws/aws_instance_ext.go
@@ -1,6 +1,10 @@
 package aws
 
-import "github.com/cloudskiff/driftctl/pkg/resource"
+import (
+	"fmt"
+
+	"github.com/cloudskiff/driftctl/pkg/resource"
+)
 
 func (r *AwsInstance) NormalizeForState() (resource.Resource, error) {
 	if r.RootBlockDevice != nil && len(*r.RootBlockDevice) == 0 {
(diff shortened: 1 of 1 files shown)


lotoussa (commented):
We generally place String() method in a resourc_ext.go file, and with the #439 PR coming, stringer should always return something like:
"id\n        Field1: value, Field2: value" if fields exist, else "id"

pkg/resource/aws/aws_instance.go
```diff
@@ -102,3 +102,10 @@ func (r *AwsInstance) TerraformType() string {
 func (r *AwsInstance) CtyValue() *cty.Value {
 	return r.CtyVal
 }
+
+func (r *AwsInstance) String() string {
```
eliecharra, line 106: This should be put in `aws_instance_ext.go` file as this file is generated and should not be altered. This is gonna be deleted btw but it will be more straightforward for us if it follow the same architecture than other resources.

pkg/resource/aws/aws_instance.go
```diff
@@ -102,3 +102,10 @@ func (r *AwsInstance) TerraformType() string {
...
+	if name, ok := r.Tags["Name"]; ok {
+		return name
```
eliecharra, line 108: We should return the id in any case.

If a name exist : `i-abcdef12345678 (Name: my-instance)` if not only show `i-abcdef12345678`

eliecharra (changes requested):
A quick small test could be nice (not a lot of value we agree)

eliecharra:
@sundowndev Target branch should be `0.8`

eliecharra (approved):

PR #475, opened 2021-04-30
Issue 439: Always print ID in resource stringer

lotoussa:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | yes
| 🚀 New feature?   | no
| ⚠ Deprecations?   | no
| ❌ BC Break       | yes
| 🔗 Related issues | #439 
| ❓ Documentation  | yes - WIP

## Description

Every resource in the console output should show the id, and then on the next line indented fields to clarify how to add some resources to the .driftignore file.
Update documentation.

Deletion of stringer String() and create a humanizer interface with a attributes method to get attribute from a resource in a string form.


eliecharra (commented):
⚠️ You forgot a stringer on resource `FakeResourceStringer`

But this remark will no more make sense if we decide to change ho stringer work.
Actually we duplicate the print logic `%s\n        (Attr: val, etc ...)` in every stringer, we do not want to do that IMHO. What about moving the logic to always print the `ID` in the console and make the stringer only return additional fields if any ? WDYT @moadibfr @wbeuil @sundowndev

moadibfr:
> What about moving the logic to always print the ID in the console and make the stringer only return additional fields if any ? 

When we looked at it one of the probleme with that is that the error case for some stringer is to display the id only. Thus console output will have to check if the stringer return something and do some logic based on that. We thought if we do stringer the stringer should be the one who decide it's output and should not return empty string that have to be checked.

eliecharra:
> When we looked at it one of the probleme with that is that the error case for some stringer is to display the id only. Thus console output will have to check if the stringer return something and do some logic based on that. We thought if we do stringer the stringer should be the one who decide it's output and should not return empty string that have to be checked.

Maybe `Stringer` is not what we really want do do anymore, maybe we can introduce a new interface and a method to replace `Stringer` to return a `map[string]string` that will be displayed in the console output ?

```go
interface HumanizedResource {
  Attrs() map[string]string
}

...

func (r *AwsIamAccessKey) Attrs() map[string]string {
        attrs := make(map[string]string)
	if r.User != nil && *r.User != "" {
		attrs["User"] = *r.User
	}
	return attrs
}
```

pkg/resource/aws/aws_security_group_rule_ext.go
```diff
@@ -128,5 +128,5 @@ func (r *AwsSecurityGroupRule) String() string {
 	if r.PrefixListIds != nil && len(*r.PrefixListIds) > 0 {
 		attrs = append(attrs, fmt.Sprintf("%s: %s", sourceOrDestination, strings.Join(*r.PrefixListIds, ", ")))
 	}
-	return strings.Join(attrs, ", ")
+	return fmt.Sprintf("%s\n        %s", r.CreateIdHash(), strings.Join(attrs, ", "))
```
shouze, line 131: this `CreateIdHash` was definitively missing, thanks!

moadibfr:
> Maybe `Stringer` is not what we really want do do anymore, maybe we can introduce a new interface and a method to replace `Stringer` to return a `map[string]string` that will be displayed in the console output ?

Sounds nice to me yep

pkg/resource/aws/aws_security_group_rule_ext.go
lotoussa: I'm reworking this PR, ID will be print every time for each resource when i'm done.

pkg/resource/aws/aws_route_ext.go
```diff
@@ -8,16 +8,22 @@ import (
 	"github.com/pkg/errors"
 )
 
-func (r *AwsRoute) String() string {
-	var destination string
+func (r *AwsRoute) Attrs() map[string]string {
+	attrs := make(map[string]string)
+	routeId, err := CalculateRouteID(r.RouteTableId, r.DestinationCidrBlock, r.DestinationIpv6CidrBlock)
```
moadibfr, line 13: Quite sur the route should have an idea here instead of computing a new one ?

pkg/resource/aws/aws_route_table_association_ext.go
```diff
@@ -1,14 +1,42 @@
...
+	}
+	if buf.Len() > 0 {
+		return fmt.Sprintf("rtassociation-%d", hashcode.String(buf.String()))
+	}
+	return ""
+}
+
+func (r *AwsRouteTableAssociation) Attrs() map[string]string {
+	attrs := make(map[string]string)
+	if idHash := r.CreateIdHash(); idHash != "" {
```
moadibfr, line 29: same here. Printing or getting attribute should not create new attributes

pkg/resource/aws/aws_security_group_rule_ext.go
```diff
@@ -84,49 +84,49 @@ func (r *AwsSecurityGroupRule) CreateIdHash() string {
 	return fmt.Sprintf("sgrule-%d", hashcode.String(buf.String()))
 }
 
-func (r *AwsSecurityGroupRule) String() string {
-	attrs := []string{}
+func (r *AwsSecurityGroupRule) Attrs() map[string]string {
+	attrs := make(map[string]string)
+	if idHash := r.CreateIdHash(); len(idHash) > 0 {
```
moadibfr, line 89: same

moadibfr (changes requested):

pkg/resource/aws/aws_route_ext.go
lotoussa: Ok so i delete this field of the map and console.go will handle the Id himself.

pkg/resource/aws/aws_route_table_association_ext.go
lotoussa: CreateIdHash() deleted.

lotoussa:
This PR need to be squash when merged because the commit log is corrupted.

pkg/output/humanizer.go
```diff
@@ -0,0 +1,23 @@
...
+
+import (
+	"fmt"
+)
+
+type AttributesGetter interface {
+	Attributes() map[string]string
+}
+
+func HumanMapPrint(res AttributesGetter) string {
```
moadibfr, line 11: HumanizeAttribute

pkg/cmd/scan/output/console.go
```diff
@@ -34,43 +35,45 @@ func NewConsole() *Console {
...
-		fmt.Printf("Found changed resources:\n")
+		fmt.Println("Found changed resources:")
 		for _, difference := range analysis.Differences() {
-			humanString := difference.Res.TerraformId()
-			if stringer, ok := difference.Res.(fmt.Stringer); ok {
-				humanString = stringer.String()
+			humanString := fmt.Sprintf("    - %s (%s):", difference.Res.TerraformId(), difference.Res.TerraformType())
+			whiteSpace := "        "
+			if humanizerRes, ok := difference.Res.(output.AttributesGetter); ok {
+				humanString += fmt.Sprintf("\n        %s", output.HumanMapPrint(humanizerRes))
```
moadibfr, line 73: what will this do if HumanMapPrint == "" ??

pkg/cmd/scan/output/testdata/output_stringer_resources.txt
```diff
@@ -1,12 +1,15 @@
 Found missing resources:
```
moadibfr, line 1: would be a good idea to add a test for a res that have no attribute (HumanMapPrint == "")

moadibfr (changes requested):

pkg/cmd/scan/output/testdata/output_stringer_resources.txt
lotoussa: i add a new test with adapted modification in console.go

PR #492, opened 2021-05-07
migrate sns to new arch

moadibfr:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | no
| 🚀 New feature?   | yes
| ⚠ Deprecations?   | no
| ❌ BC Break       | yes
| 🔗 Related issues | #347 
| ❓ Documentation  | no

## Description

migrate
sns_topic
sns_topic_policy
sns_topic_subscription

### DIFF
--- a/pkg/driftctl_test.go
+++ b/pkg/driftctl_test.go
@@ -836,13 +836,11 @@ func TestDriftctlRun_Middlewares(t *testing.T) {
 				},
 			},
 			mocks: func(factory resource.ResourceFactory) {
-				factory.(*terraform.MockResourceFactory).On("CreateAbstractResource", "foo", "aws_sns_topic_policy", mock.MatchedBy(func(input map[string]interface{}) bool {
-					return matchByAttributes(input, map[string]interface{}{
-						"id":     "foo",
-						"arn":    "arn",
-						"policy": "{\"policy\":\"bar\"}",
-					})
-				})).Times(1).Return(&resource.AbstractResource{
+				factory.(*terraform.MockResourceFactory).On("CreateAbstractResource", "foo", "aws_sns_topic_policy", map[string]interface{}{
+					"id":     "foo",
+					"arn":    "arn",
+					"policy": "{\"policy\":\"bar\"}",
+				}).Times(1).Return(&resource.AbstractResource{
 					Id:   "foo",
 					Type: aws.AwsSnsTopicPolicyResourceType,
 					Attrs: &resource.Attributes{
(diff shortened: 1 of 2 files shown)


pkg/driftctl_test.go
```diff
@@ -822,51 +822,55 @@ func TestDriftctlRun_Middlewares(t *testing.T) {
...
 				},
 			},
 			mocks: func(factory resource.ResourceFactory) {
-				foo := cty.ObjectVal(map[string]cty.Value{
-					"id":     cty.StringVal("foo"),
-					"arn":    cty.StringVal("arn"),
-					"policy": cty.StringVal("{\"policy\":\"bar\"}"),
-				})
-				factory.(*terraform.MockResourceFactory).On("CreateResource", mock.MatchedBy(func(input map[string]interface{}) bool {
+				factory.(*terraform.MockResourceFactory).On("CreateAbstractResource", "foo", "aws_sns_topic_policy", mock.MatchedBy(func(input map[string]interface{}) bool {
```
eliecharra, line 847: `mock.MatchedBy` can be removed and you can use directly a `map[string]interface{}`

pkg/middlewares/aws_sns_topic_policy_expander.go
```diff
@@ -44,43 +47,35 @@ func (m AwsSNSTopicPolicyExpander) Execute(_, resourcesFromState *[]resource.Res
...
-	data := map[string]interface{}{
-		"arn":    topic.Arn,
-		"id":     topic.Id,
-		"policy": topic.Policy,
-	}
-	ctyVal, err := m.resourceFactory.CreateResource(data, "aws_sns_topic_policy")
-	if err != nil {
-		return err
+	arn, exist := topic.Attrs.Get("arn")
+	if !exist || arn == "" {
```
eliecharra, line 57: This case was not handled previously, can it really happen ?

pkg/middlewares/aws_sns_topic_policy_expander_test.go
```diff
@@ -27,83 +27,198 @@ func TestAwsSNSTopicPolicyExpander_Execute(t *testing.T) {
...
+						"arn":    "arn",
+						"id":     "ID",
+						"policy": "{\"policy\":\"coucou\"}",
+					},
 				},
 			},
 			mock: func(factory *terraform.MockResourceFactory) {
-				factory.On("CreateResource", mock.Anything, "aws_sns_topic_policy").Once().Return(nil, nil)
+				factory.On("CreateAbstractResource", "ID", awsresource.AwsSnsTopicPolicyResourceType, mock.MatchedBy(func(input map[string]interface{}) bool {
+					return input["id"] == "ID"
```
eliecharra, line 61: Same remark as in another PR, we should change the ID's or either change the matcher

moadibfr: I don't get this remark ? There is only one resource so only this id ?

pkg/middlewares/aws_sns_topic_policy_expander.go
eliecharra: Yeah it can't hurt it was just to be sure to understand.

PR #499, opened 2021-05-07
Add ecr | route_table | route

wbeuil:
```
aws_ecr_repository
aws_route_table
aws_route
```

⚠️ I don't recall we spoke about the Stringer so I commented for now the method for the `aws_route`.


eliecharra:
> ⚠️ I don't recall we spoke about the Stringer so I commented for now the method for the aws_route.

For now we do not have a mechanism to handle this, we'll do another pass of refactoring when @lotoussa's PR will be merged to main

Issue #461, opened 2021-04-26
driftctl should report when no tfstate found on S3


sjourdan:
**Description**

As a user, when I `driftctl scan` using an S3 bucket with no tfstate in it, I expect at least a warning or an error - certainly not the tool going on as if everything was smooth.

**Environment**
 - OS: Fedora 34
 - driftctl version: 0.7.1

**How to reproduce**

1. create an empty S3 bucket
2. run driftctl scan --from tfstate+s3://my-empty-bucket/
3. confirm a scan happened fully, even without no tfstate found

**Possible Solution**

```shell 
$ driftctl scan --from tfstate+s3://mybucket/ 
No Terraform states found on s3://mybucket, exiting
```

**Additional context**

Bug found by @sundowndev :)

Locally it's less bizarre - even if it's not super user-friendly: `lstat terraform.tfstate: no such file or directory`.

sundowndev:
For the record, I think this should be done for all enumerators here https://github.com/cloudskiff/driftctl/blob/main/pkg/iac/terraform/state/terraform_state_reader.go#L238

sjourdan:
LGTM 

```
❯  AWS_PROFILE="cloudskiff-demo" LOG_LEVEL=critical driftctl-0.10.0 --tf-provider-version=3.44.0 scan --from tfstate+s3://dctl-461-nostate/**/*.tfstate
Scanned states (0)      
no Terraform state was found in dctl-461-nostate/**/*.tfstate, exiting
```

This is a bit at odds with #623 though. Let's think about the feature again.

PR #506, opened 2021-05-11
Report when no tfstate found in backend

sundowndev:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | yes
| 🚀 New feature?   | no
| ⚠ Deprecations?   | no
| ❌ BC Break       | no
| 🔗 Related issues | #461
| ❓ Documentation  | no 

## Description

As a user, when I run driftctl scan against an S3 bucket or local directory with no tfstate in it, I expect an error - certainly not the tool going on as if everything was smooth.


eliecharra (approved):

AzySir:
@sundowndev  Hey Raph, have you added in a glob pattern test with this? I believe I've hit a defect in v0.11, this is giving the error no Terraform state exists when there is a terraform state. However I'm scanning via glob...

eliecharra:
@AzySir Check https://github.com/cloudskiff/driftctl/issues/695, this is currently be fixed. Should be released in `v0.11.1` in a couple of minutes.

PR #519, opened 2021-05-17
Refacto aws route53 resources

eliecharra:


pkg/resource/aws/aws_route53_health_check.go
```diff
@@ -41,3 +44,9 @@ func (r *AwsRoute53HealthCheck) TerraformType() string {
 func (r *AwsRoute53HealthCheck) CtyValue() *cty.Value {
 	return r.CtyVal
 }
+
+func initAwsRoute53HealthCheckMetaData(resourceSchemaRepository resource.SchemaRepositoryInterface) {
+	resourceSchemaRepository.SetNormalizeFunc(AwsRoute53HealthCheckResourceType, func(val *resource.Attributes) {
+
+	})
```
wbeuil, line 51: Why is this empty ? And if it is intentional, why ?

pkg/middlewares/route53_records_id_reconcilier.go
```diff
@@ -0,0 +1,49 @@
...
+// From Terraform state, we retrieve: 1234_toto_TXT
+type Route53RecordIDReconcilier struct{}
+
+func NewRoute53RecordIDReconcilier() Route53RecordIDReconcilier {
+	return Route53RecordIDReconcilier{}
+}
+
+func (m Route53RecordIDReconcilier) Execute(_, resourcesFromState *[]resource.Resource) error {
+
+	for _, remoteResource := range *resourcesFromState {
```
wbeuil, line 24: Should be `stateResource` instead of `remoteResource`

wbeuil (changes requested):

pkg/middlewares/route53_records_id_reconcilier.go
eliecharra: Nice catch thanks

pkg/resource/aws/aws_route53_health_check.go
eliecharra: I need to remove it nice catch thanks

PR #521, opened 2021-05-17
migrate  security_group & security_group_rules

moadibfr:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | no
| 🚀 New feature?   | yes
| ⚠ Deprecations?   | no
| ❌ BC Break       | yes
| 🔗 Related issues | #347 
| ❓ Documentation  | no

## Description

migrate  security_group & security_group_rules


pkg/iac/terraform/state/test/vpc_security_group/result.golden.json
```diff
@@ -1,20 +1,44 @@
...
-   "Name": "Foo SG"
-  },
-  "VpcId": "vpc-618f6e09",
-  "Timeouts": null,
-  "CtyVal": {}
+  "Type": "aws_security_group",
+  "Attrs": {
+   "arn": "arn:aws:ec2:eu-west-3:047081014315:security-group/sg-0254c038e32f25530",
+   "description": "Managed by Terraform",
+   "egress": [
```
eliecharra, line 8: Previously we got null for `egress` and `ingress`, we remove them in the normalize func, I don't get why we got them there 🤔

pkg/middlewares/vpc_security_group_default_test.go
```diff
@@ -68,8 +92,9 @@ func TestDefaultVPCSecurityGroupShouldNotBeIgnoredWhenManaged(t *testing.T) {
 	if len(remoteResources) != 3 {
 		t.Error("Default security group was ignored")
 	}
-	managedDefaultSecurityGroup := remoteResources[2].(*aws.AwsDefaultSecurityGroup)
-	if *managedDefaultSecurityGroup.Name != "default" {
+	managedDefaultSecurityGroup := remoteResources[2].(*resource.AbstractResource)
+	name, _ := managedDefaultSecurityGroup.Attrs.Get("name")
```
eliecharra, line 96: Can we use `GetString` there ?

pkg/resource/resource.go
```diff
@@ -182,11 +182,68 @@ func ToResourceAttributes(val *cty.Value) *Attributes {
...
+func (a *Attributes) GetInt(path string) *int {
+	val, exist := (*a)[path]
+	if !exist {
+		return nil
+	}
+	v := val.(int)
+	return &v
+}
+
+func (a *Attributes) Set(path string, value interface{}) {
```
eliecharra, line 243: Why do we need a new `Set` function ? Can we reuse `SafeSet` instead ?

pkg/resource/aws/aws_security_group_rule.go
```diff
@@ -32,3 +40,72 @@ func (r *AwsSecurityGroupRule) TerraformType() string {
...
+func initAwsSecurityGroupRuleMetaData(resourceSchemaRepository resource.SchemaRepositoryInterface) {
+	resourceSchemaRepository.SetNormalizeFunc(AwsDefaultSecurityGroupResourceType, func(val *resource.Attributes) {
+		val.SafeDelete([]string{"self"})
+		if sgid := val.GetString("security_group_id"); sgid != nil && *sgid == "" {
+			val.SafeDelete([]string{"security_group_id"})
+		}
+
+		// On first run, this field is set to null in state file and to "" after one refresh or apply
+		// This ensure that if we find a nil value we dont drift
+		if desc := val.GetString("description"); desc != nil && *desc == "" {
```
eliecharra, line 96: Maybe you can use `val.DeleteIfDefault("description")` ? For `security_group_id` too

pkg/middlewares/vpc_security_group_rules_test.go
```diff
@@ -14,61 +13,111 @@ import (
 func TestVPCSecurityGroupRuleSanitizer(t *testing.T) {
 
 	factory := &terraform.MockResourceFactory{}
-	factory.On("CreateResource", mock.Anything, "aws_security_group_rule").Times(8).Return(nil, nil)
+	factory.On("CreateAbstractResource", aws.AwsSecurityGroupRuleResourceType, "sgrule-1175318309", mock.Anything).Times(1).Return(
```
eliecharra, line 16: Just to understand, you use `mock.Anything` because if any of the fields does not match the hash will change and thus the mock will not match anymore ? Maybe it will be easier to maintain to use `mock.Anything` on the ID and assert the whole attribute map ?

eliecharra (commented):
@wbeuil I did a quick review pass, since you implemented sg groups and rules it can be valuable if you can do a review pass here

pkg/iac/terraform/state/test/vpc_security_group/result.golden.json
moadibfr: hmm good question. We still safe delete them in normalize function if I'm right. I'll check...

pkg/middlewares/vpc_security_group_rules_test.go
moadibfr: the test use to no check anything I added a check on the ID as it is generated on the field and it was easier to implement than testing with the whole attribute map...

wbeuil (changes requested):
Have casting errors on middlewares.

image

pkg/middlewares/vpc_security_group_rules_test.go
eliecharra: 🆗

PR #531, opened 2021-05-19
try to change region in vpc acc test to mitigate max number of vpc

moadibfr:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | yes
| 🚀 New feature?   | no
| ⚠ Deprecations?   | no
| ❌ BC Break       | no
| 🔗 Related issues | 
| ❓ Documentation  | no 

## Description

try to change region in vpc acc test to mitigate max number of vpc

### DIFF
--- a/pkg/resource/aws/testdata/acc/aws_vpc/terraform.tf
+++ b/pkg/resource/aws/testdata/acc/aws_vpc/terraform.tf
@@ -1,5 +1,5 @@
 provider "aws" {
-  region = "us-east-1"
+  region = "us-east-2"
 }
 
 terraform {


wbeuil (approved):

pkg/resource/aws/testdata/acc/aws_vpc/terraform.tf
```diff
@@ -1,5 +1,5 @@
 provider "aws" {
-  region = "us-east-1"
+  region = "us-east-2"
```
eliecharra, line 2: This should never work and should be reverted IMHO @moadibfr @wbeuil. We are targetting the symptoms but not the root cause. And BTW this should be changed in the test case too 

image

moadibfr: Yeah we wanted to try having the test pass this does not mean we consider this a fix for the other test that fail. But this if we isolate this one it would be easier to find the one that exhaust the vpc limit. Also it's not armfull to test on different region.

I didn't remember the aws_region but it's strange that the test still pass thought ? I tried it in my env and it runs ok ?1

PR #540, opened 2021-05-21
Remove old resources

eliecharra:


pkg/resource/resource.go
```diff
@@ -15,83 +15,19 @@ import (
...
-	return false
+	Attributes() *Attributes
+	Schema() *Schema
 }
 
 type AbstractResource struct {
 	Id    string
 	Type  string
 	Attrs *Attributes
+	Sch   *Schema `json:"-" diff:"-"`
```
eliecharra, line 26: Why `diff:"-"` ? 

BTW I really have responsibility issue with this. I don't get why we do not keep resource and schema separated. For me there is not need to inject the schema in the resource they are two distinct struct (like cty value and cty type)

We may have a vocal talk about it

pkg/remote/aws/test/iam_role_policy_attachment_multiple/.terraform.lock.hcl
```diff
@@ -0,0 +1,20 @@
+# This file is maintained automatically by "terraform init".
```
eliecharra, line 1: This file is useless

pkg/remote/aws/test/iam_user_policy_attachment_multiple/.terraform.lock.hcl
```diff
@@ -0,0 +1,21 @@
+# This file is maintained automatically by "terraform init".
```
eliecharra, line 1: This file is useless

pkg/resource/deserializer.go
```diff
@@ -0,0 +1,24 @@
...
+
+func NewDeserializer(factory ResourceFactory) *Deserializer {
+	return &Deserializer{factory}
+}
+
+func (s *Deserializer) Deserialize(ty string, rawList []cty.Value) ([]Resource, error) {
+	resources := make([]Resource, 0)
+	for _, rawResource := range rawList {
+		rawResource := rawResource
+		attrs := ToResourceAttributes(&rawResource)
```
eliecharra, line 19: What's the purpose of `ToResourceAttributes` ? This logic do not need to be in a dedicated exported method, it does a deserialization of cty values object and it is exactly the purpose of this deserializer.

pkg/resource/aws/testdata/acc/aws_security_group/terraform.tf
```diff
@@ -72,26 +67,6 @@ resource "aws_security_group" "sg-bis-tutu" {
     }
 }
 
-resource "aws_security_group_rule" "tutu-egress" {
```
eliecharra, line 75: I don't get why do we have diff on theses acceptance test here

pkg/resource/aws/testdata/acc/aws_iam_access_key/terraform.tf
```diff
@@ -16,3 +16,13 @@ resource "aws_iam_user" "testuser_access_key" {
 resource "aws_iam_access_key" "accesskey" {
     user = aws_iam_user.testuser_access_key.name
 }
+
+resource "aws_iam_access_key" "ak" {
+    user    = aws_iam_user.test.name
+    pgp_key = "keybase:test"
```
eliecharra, line 22: Is this related to this PR ? 🤔

pkg/resource/resource.go
```diff
@@ -102,12 +38,20 @@ func (a *AbstractResource) TerraformType() string {
 	return a.Type
 }
 
-func (a *AbstractResource) CtyValue() *cty.Value {
-	return nil
+func (a *AbstractResource) Attributes() *Attributes {
+	return a.Attrs
+}
+
+func (a *AbstractResource) HumanReadableAttributes() map[string]string {
```
eliecharra, line 45: I think this logic should not be related to the resource struct.
 
We put too many thinks in resource : the schema, the data itself, the human readable function that is bound to the schema 🤔 not very clear IMHO we should rethink this.

pkg/resource/resource.go
```diff
@@ -249,6 +201,26 @@ func (a *Attributes) GetFloat64(path string) *float64 {
 	return &v
 }
 
+func (a *Attributes) GetMap(path string) map[string]interface{} {
```
eliecharra, line 204: This should be unexported or done directly in `GetStringMap`.

BTW make a full copy of the map looks heavy for a simple type cast and I don't get why we don't do this :

image

pkg/resource/resource.go
```diff
@@ -215,11 +163,15 @@ func (a *Attributes) GetString(path string) *string {
...
 
 func (a *Attributes) GetStringSlice(path string) []string {
-	val, exist := (*a)[path]
-	if !exist {
+	val := a.GetSlice(path)
+	if val == nil {
 		return nil
 	}
-	return val.([]string)
+	slice := make([]string, 0, len(val))
```
eliecharra, line 170: Why do we need to make a copy ?

pkg/resource/resource.go
```diff
@@ -15,83 +15,19 @@ import (
...
-}
-
-func IsRefactoredResource(typ string) bool {
-	for _, refactoredResource := range refactoredResources {
-		if typ == refactoredResource {
-			return true
-		}
-	}
-	return false
+	Attributes() *Attributes
```
eliecharra, line 18: I had planned to remove the resource interface and rename `AbstractResource` to `Resource` WDYT ? I don't think this interface makes sense anymore

test/resource/resource.go
```diff
@@ -4,50 +4,48 @@ import (
 	"github.com/cloudskiff/driftctl/pkg/resource"
 	"github.com/cloudskiff/driftctl/test/schemas"
 	"github.com/hashicorp/terraform/providers"
-	"github.com/zclconf/go-cty/cty"
 )
 
 type FakeResource struct {
```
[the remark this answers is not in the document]
moadibfr: Yeah we asked ourselves the same question. I think we could remove this (and maybe remove resource interface and only use abstractresource named as resource). But this will mean re writing all the test that use it and this is not easy for every test :/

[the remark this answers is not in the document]
wbeuil: We could totally ditch the `FakeResource`

pkg/resource/aws/testdata/acc/aws_security_group/terraform.tf
wbeuil: I remove those security group rules since they are useless in this acceptance test. They were copy/pasted in the `security_group_rule` acc tests.

pkg/resource/resource.go
moadibfr: We diff AbstractResource entirely in some tests so we need to ignore schema.
I quite like that the schema is given with the concerned resource instead of having to inject a separate repository to get it. I thought there is no responsibility concern here as the resource and the schema are directly linked and we only build the resource in one place. But yeah let's talk about this tomorrow. (yet I have to add that cty value have their own type embedded :grimacing:  https://github.com/zclconf/go-cty/blob/main/cty/value.go )

pkg/resource/aws/testdata/acc/aws_iam_access_key/terraform.tf
wbeuil: That's what @sjourdan told us in the last vocal meeting `if you can put it in the refacto, go`. But I can revert the last commit.

pkg/remote/aws/test/iam_role_policy_attachment_multiple/.terraform.lock.hcl
moadibfr: should we add version in the associated tf to lock something or just delete it ?

pkg/resource/deserializer.go
moadibfr: yeah you're right. I think this is a legacy method that we did not move

pkg/resource/resource.go
wbeuil: I added both the `GetMap` and `GetMapString` methods to have more method utilities in our pocket. And it totally make sense to only have the GetMap and cast the value at the end, didn't think of it.

pkg/resource/resource.go
moadibfr: yeah I think this is the same discussion as for separation between the schema and the resource. It is handy that abstractresource just implement the HumanReadableAttributes interface I think

pkg/resource/resource.go
wbeuil: Just like the `GetMap` we can't use the `[]string` cast on `[]interface{}`. I don't remember the resource I got the error ..

pkg/resource/resource.go
wbeuil: I agree !

pkg/resource/resource.go
moadibfr: this was supposed to be one of the getter/autocast function like GetSlice, GetString, ...
And yeah I think we could replace the getstringmap you're right

pkg/remote/aws/test/iam_role_policy_attachment_multiple/.terraform.lock.hcl
wbeuil: Just delete it

pkg/resource/resource.go
moadibfr: same as getStringmap it is not possible to cast []inteferace{} to []string

pkg/resource/resource.go
moadibfr: yeah as I said in a response before this one I think it would be great but rewriting all the tests is really painfull :cry:

pkg/resource/resource.go
eliecharra: For me this is more related to the `View` than the `Model` part if we think MVC. I had in mind that the resource is our model and is only a data container object. This should not be bound to a resource.

test/resource/resource.go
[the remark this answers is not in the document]
eliecharra: We'll do this in another PR

moadibfr (approved):

PR #547, opened 2021-05-24
Implement caching in EC2 repository

sundowndev:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | no
| 🚀 New feature?   | yes
| ⚠ Deprecations?   | no
| ❌ BC Break       | no
| 🔗 Related issues | #421
| ❓ Documentation  | no 

## Description

>  Sometimes we enumerate multiple time the same kind of resource, for example S3 buckets enumeration is done in S3BucketMetricSupplier, etc ...
We want to cache resources enumeration in the repository layer.
It will also ensure atomicity during a scan for AWS endpoints that have inconsistency.

This PR implements cache package and implement cache in EC2 repository. More PRs will come to implement cache in other repositories.

Issue #421, opened 2021-04-12
We should cache resources enumeration


eliecharra:
**Description**
Sometimes we enumerate multiple time the same kind of resource, for example S3 buckets enumeration is done in  `S3BucketMetricSupplier`, etc ...
We want to cache resources enumeration in the repository layer.
It will also ensure atomicity during a scan for AWS endpoints that have inconsistency.

**Technical details**
We want to have a global cache mechanism that is used in multiple repositories.
We want to use the caching layer only for resources that are enumerated multiple time.

sjourdan:
looks very trustable to me :)

PR #558, opened 2021-05-27
Implement cache in AWS S3 repository

sundowndev:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | no
| 🚀 New feature?   | yes
| ⚠ Deprecations?   | no
| ❌ BC Break       | no
| 🔗 Related issues | #421 
| ❓ Documentation  | no 

## Description

This PR implements cache in S3 repository. **Should be reviewed after #547**

### DIFF
--- a/pkg/remote/aws/ec2_eip_association_supplier.go
+++ b/pkg/remote/aws/ec2_eip_association_supplier.go
@@ -20,11 +20,11 @@ type EC2EipAssociationSupplier struct {
 	runner       *terraform.ParallelResourceReader
 }
 
-func NewEC2EipAssociationSupplier(provider *AWSTerraformProvider, deserializer *resource.Deserializer) *EC2EipAssociationSupplier {
+func NewEC2EipAssociationSupplier(provider *AWSTerraformProvider, deserializer *resource.Deserializer, repo repository.EC2Repository) *EC2EipAssociationSupplier {
 	return &EC2EipAssociationSupplier{
 		provider,
 		deserializer,
-		repository.NewEC2Repository(provider.session),
+		repo,
 		terraform.NewParallelResourceReader(provider.Runner().SubRunner())}
 }
 
--- a/pkg/remote/aws/init.go
+++ b/pkg/remote/aws/init.go
@@ -5,6 +5,7 @@ import (
 	"github.com/cloudskiff/driftctl/pkg/output"
 	"github.com/cloudskiff/driftctl/pkg/remote/aws/client"
 	"github.com/cloudskiff/driftctl/pkg/remote/aws/repository"
+	"github.com/cloudskiff/driftctl/pkg/remote/cache"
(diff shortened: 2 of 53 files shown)


pkg/remote/aws/ec2_eip_association_supplier.go
```diff
@@ -21,11 +22,11 @@ type EC2EipAssociationSupplier struct {
 	runner       *terraform.ParallelResourceReader
 }
 
-func NewEC2EipAssociationSupplier(provider *AWSTerraformProvider) *EC2EipAssociationSupplier {
+func NewEC2EipAssociationSupplier(provider *AWSTerraformProvider, c cache.Cache) *EC2EipAssociationSupplier {
 	return &EC2EipAssociationSupplier{
 		provider,
 		awsdeserializer.NewEC2EipAssociationDeserializer(),
-		repository.NewEC2Repository(provider.session),
+		repository.NewEC2Repository(provider.session, c),
```
moadibfr, line 29: I wonder if we should build the repository before passing it to supplier instead of passing supplier dependencies ?

moadibfr: that way we could also pass the same repository to every client instead of building one each time ?

pkg/remote/aws/init.go
```diff
@@ -26,7 +27,9 @@ func Init(alerter *alerter.Alerter, providerLibrary *terraform.ProviderLibrary,
 		return err
 	}
 
-	s3Repository := repository.NewS3Repository(client.NewAWSClientFactory(provider.session))
+	repositoryCache := cache.New(100)
```
moadibfr, line 30: can't we have a bigger cache than 100 ?

moadibfr (changes requested):

pkg/remote/aws/init.go
sundowndev: From what I calculated (~5 methods per repository for 11 repositories), we should not reach more than 55 keys. I set capacity of 100 because we have a lot of margin. I think we can easily increase capacity up to 1000 without any negative impact on performances.

pkg/remote/aws/ec2_eip_association_supplier.go
sundowndev: You mean the same way we already do with the S3 repository ? https://github.com/cloudskiff/driftctl/blob/main/pkg/remote/aws/init.go#L29

pkg/remote/aws/init.go
moadibfr: Ho you're right I thought at first read this would be number of resource but this will be one by resource type :+1:

pkg/remote/aws/ec2_eip_association_supplier.go
moadibfr: yep exactly

Issue #446, opened 2021-04-19
Enable drift detection on `aws_instance` shutdown behavior


eliecharra:
**Description**
Following https://github.com/cloudskiff/driftctl/issues/403, we should think about removing the workaround for newest provider version as this is now fixed on terraform provider side.

**Possible Solution**
We should find a way to fetch the terraform provider version and compare semver version to disable the fix for provider version > `3.38.0`

sjourdan:
@eliecharra says: 

_Elie: @Stephane Small precision about this issue https://github.com/cloudskiff/driftctl/issues/446 (how to test it). First create an instance with tf with a provider version <= 3.38. then manually change shutdown behavior in the console => no drift should be detected. Try with a provider version > 3.38 and then it should be detected. Then rollback to the good value from the console => no drift should happen_

sjourdan:
LGTM! 

Detected: 

```
  Name: "vm-1"
} (computed)
        + throughput: <nil> => 0 (computed)
    - i-0cb27ba2d1fb8312d (aws_instance):
        + capacity_reservation_specification: <nil> => [map[capacity_reservation_preference:open]]
        ~ instance_initiated_shutdown_behavior: "" => "terminate" (computed)
        + tags_all: <nil> => {
  Name: "vm-1",
  Terraform: "true",
  Environment: "dev"
} (computed)

```

No detected: 

```
  Name: "vm-1"
} (computed)
        + throughput: <nil> => 0 (computed)
```

(the false positive is https://github.com/cloudskiff/driftctl/issues/612)

PR #561, opened 2021-05-28
Enable diff on field InstanceInitiatedShutdownBehavior

sundowndev:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | yes
| 🚀 New feature?   | no
| ⚠ Deprecations?   | no
| ❌ BC Break       | no
| 🔗 Related issues | #446
| ❓ Documentation  | no 

## Description

That field was ignored because of a bug in AWS Terraform provider. Since it was fixed, we can now enable diff on it and fix the issue the good way.

### DIFF
--- a/pkg/remote/aws/init.go
+++ b/pkg/remote/aws/init.go
@@ -24,7 +24,9 @@ func Init(version string, alerter *alerter.Alerter,
 	progress output.Progress,
 	resourceSchemaRepository *resource.SchemaRepository,
 	factory resource.ResourceFactory) error {
-
+	if version == "" {
+		version = "3.19.0"
+	}
 	provider, err := NewAWSTerraformProvider(version, progress)
 	if err != nil {
 		return err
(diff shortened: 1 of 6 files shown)


eliecharra:
@sundowndev I think this PR can be closed, the logic to fix this need to be applied to the new architecture. To achieve this we need to retrieve the provider version and enable this field only for specific version of the provider which contains the bugfix. This is something blocked by the current rework we are working on. Maybe we can open an issue instead ?

sundowndev:
> @sundowndev I think this PR can be closed, the logic to fix this need to be applied to the new architecture. To achieve this we need to retrieve the provider version and enable this field only for specific version of the provider which contains the bugfix. This is something blocked by the current rework we are working on. Maybe we can open an issue instead ?

Isn't it the point of your issue #446 ? I created this PR as a starting point for the fix

eliecharra:
> Isn't it the point of your issue #446 ? I created this PR as a starting point for the fix

Yeah my bad the issue was already there 🙏🏻

Issue #494, opened 2021-05-07
Scanning a dir using HTTP backend outputs an error


sjourdan:
**Description**

Scanning an empty dir with an HTTP backend should output something clean and meaningful to the user, as we do for other errors!

**Environment**
 - OS: fedora 34
 - driftctl version: 0.8.0 (pre)

**How to reproduce**

Inside an HTTP served directory, create some folder: 

```
mkdir bogus-dir
```

Then request this folder with driftctl 

```
  LOG_LEVEL=error driftctl-0.8.0 scan --from tfstate+https://***.com/bogus-dir/
Scanned states (1)      
2 problems:

- Unsupported state file format: The state file could not be parsed as JSON: syntax error at byte offset 1.
- Unsupported state file format: The state file does not have a "version" attribute, which is required to identify the format version.

```

**Possible Solution**

**Additional context**

sjourdan:
Context: a user wants to load all the tfstates from a remote directory available through HTTP
Problem: maybe the user mistypes the folder name, or gives an empty folder name
Driftctl UX problem: instead of telling the user we didn't find any readable state file, we tell him about syntax errors and wrong formats (which are not helpful for the user to find his own issue)

Idea as mentioned in the original issue: 

```
$ driftctl-0.8.0 scan --from tfstate+https://***.com/bogus-dir/
Scanned states (1)      
1 problem:

- No Terraform state file found in the specified directory.
```

AzySir:
Suggestion also for this, add a `--skips-if-no-state` type of flag. As I'm waiting for other devs to create statefiles but the buckets may exist so I can code in parallel. 

_(Work around for this is an empty .tfstate, however I believe that still leads to a fail)_

sjourdan:
confirmed ok in 0.9.1 and 0.10.0-pre

```
❯  DCTL_TF_PROVIDER_VERSION="3.44.0" AWS_REGION="eu-west-3" AWS_PROFILE="cloudskiff-demo" driftctl-0.10.0 scan --from tfstate+https://driftctl.com/bogus/
Scanned states (1)      
Failed to read state file: The state file could not be read: error requesting HTTP(s) backend state: status code: 404
``` 

@AzySir I referenced your initial suggestion in the issue that is more likely to solve your use-case https://github.com/cloudskiff/driftctl/issues/623#issuecomment-864073056

PR #575, opened 2021-06-02
Fix error reporting in http state reader

lotoussa:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | yes
| 🚀 New feature?   | no
| ⚠ Deprecations?   | no
| ❌ BC Break       | no
| 🔗 Related issues | #494 
| ❓ Documentation  | no

## Description

We can't recover summary of errors from the _.HTTPBackend because casting isn't possible with this structure.
So i just add an error check before returning at the end of the method.
Maybe it's need to add some test to keep track of this change, but i'll do it tomorrow after discuss it with the team.

### DIFF
--- a/pkg/iac/terraform/state/backend/http_reader.go
+++ b/pkg/iac/terraform/state/backend/http_reader.go
@@ -46,7 +46,11 @@ func (h *HTTPBackend) Read(p []byte) (n int, err error) {
 			return 0, errors.Errorf("error requesting HTTP(s) backend state: status code: %d", res.StatusCode)
 		}
 	}
-	return h.reader.Read(p)
+	read, err := h.reader.Read(p)
+	if err != nil {
+		return 0, err
+	}
+	return read, nil
 }
 
 func (h *HTTPBackend) Close() error {
--- a/pkg/iac/terraform/state/backend/http_reader.go
+++ b/pkg/iac/terraform/state/backend/http_reader.go
@@ -46,11 +46,7 @@ func (h *HTTPBackend) Read(p []byte) (n int, err error) {
 			return 0, errors.Errorf("error requesting HTTP(s) backend state: status code: %d", res.StatusCode)
 		}
 	}
-	read, err := h.reader.Read(p)
-	if err != nil {
-		return 0, err
-	}
-	return read, nil
+	return h.reader.Read(p)
 }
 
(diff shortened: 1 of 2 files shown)


pkg/iac/terraform/state/backend/http_reader.go
```diff
@@ -46,7 +46,11 @@ func (h *HTTPBackend) Read(p []byte) (n int, err error) {
 			return 0, errors.Errorf("error requesting HTTP(s) backend state: status code: %d", res.StatusCode)
 		}
 	}
-	return h.reader.Read(p)
+	read, err := h.reader.Read(p)
+	if err != nil {
+		return 0, err
+	}
+	return read, nil
```
sundowndev, line 53: What's the point of this change ? `h.reader.Read(p)` already returns an error so this shouldn't have any effect at all, should it?

lotoussa: It's work because i return 0 in the case err != nil, but it erase all error to print "no state".
So i'll rework this one.

pkg/iac/terraform/state/terraform_state_reader.go
```diff
@@ -220,6 +222,9 @@ func (r *TerraformStateReader) retrieveMultiplesStates() ([]resource.Resource, e
 func read(path string, reader backend.Backend) (*states.State, error) {
 	state, err := readState(path, reader)
 	if err != nil {
+		if _, ok := reader.(*backend.HTTPBackend); ok && strings.Contains(err.Error(), "The state file could not be parsed as JSON") {
+			return nil, errors.Errorf("given url is not a valid statefile")
```
sundowndev, line 226: I'm confused with that change as it's less explicit than the current error : 

```
2 problems:

- Unsupported state file format: The state file could not be parsed as JSON: syntax error at byte offset 1.
- Unsupported state file format: The state file does not have a "version" attribute, which is required to identify the format version.
```

I don't understand how it'll solve the original issue, can you elaborate on that? thanks

sundowndev (changes requested):

sundowndev (approved):

eliecharra (dismissed):

wbeuil (approved):

eliecharra (approved):

Issue #570, opened 2021-06-01
AWS TF provider version is customizable


sjourdan:
**Description**
As user 
I want to specify the Terraform AWS Provider version
So I can use one that matches my TF code

Default = `latest`

Source: https://releases.hashicorp.com/terraform-provider-aws/

We decided to not include the provider name, as we'll not scan 1+ different providers at once.

**Example**

```
$ DCTL_TF_PROVIDER_VERSION=3.41.0 driftctl scan 
[...]
$ ls -al ~/.driftctl/plugins/linux_amd64/terraform-provider-aws_v3.41.0_x5
``` 

# Acceptance Tests

(download only - to be improved soon)

* A user can use the tool with no Terraform provider settings configured and driftctl will download the latest available
  * from a clean env, run `driftctl scan` 
  * the user sees the latest Terraform AWS version downloaded under `~/.driftctl/plugins` (latest version is always available at https://releases.hashicorp.com/terraform-provider-aws/)
* A user can specify a 3.x version of the AWS provider and driftctl will download the requested version. 
  * from a clean env, run `DCTL_TF_PROVIDER_VERSION=3.41.0 driftctl scan` 
  * the user sees version 3.41 downloaded under `~/.driftctl/plugins`
* A user can specify a 2.x version of the AWS provider and driftctl will download the requested version. 
  * from a clean env, run `DCTL_TF_PROVIDER_VERSION=2.70.0 driftctl scan` 
  * the user sees version 2.70 downloaded under `~/.driftctl/plugins`
* A user can specify a GitHub Terraform provider version 
  * from a clean env, run `DCTL_TF_PROVIDER_VERSION=4.0.0 driftctl scan --to github+tf` 
  * the user sees version 4.0.0 downloaded under `~/.driftctl/plugins`
```

sjourdan:
Followup: https://github.com/cloudskiff/driftctl/pull/577

Context: 
- for driftctl to work properly, a TF provider should be used. 
- if none is specified (`DCTL_TF_PROVIDER_VERSION`), one should be downloaded and used. 
- if none is found locally under `~/.driftctl/`, one should be downloaded and used. 

Two questions: 
1. Which provider version should be downloaded when none is specified **AND** none is found locally 
    1. the latest of that day
    2. a fixed version we tried
2. Which provider version should be used afterwards
    1. always latest, downloading a new version every time
    2. keep using the one found in cached `~/.driftctl` until the user wants to update

[...]

PR #577, opened 2021-06-03
Allow to specify a terraform provider version

eliecharra:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | no
| 🚀 New feature?   | yes
| ⚠ Deprecations?   | no
| ❌ BC Break       | no
| 🔗 Related issues | #347 #570 
| ❓ Documentation  | yes https://github.com/cloudskiff/driftctl-docs/pull/67

### DIFF
--- a/pkg/cmd/scan.go
+++ b/pkg/cmd/scan.go
@@ -136,6 +136,11 @@ func NewScanCmd() *cobra.Command {
 		"Terraform Cloud / Enterprise API token.\n"+
 			"Only used with tfstate+tfcloud backend.\n",
 	)
+	fl.StringVar(&opts.ProviderVersion,
+		"tf-provider-version",
+		"",
+		"Terraform provider version to use.\n",
+	)
 	fl.BoolVar(&opts.StrictMode,
 		"strict",
 		false,
@@ -162,7 +167,7 @@ func scanRun(opts *pkg.ScanOptions) error {
 
 	resFactory := terraform.NewTerraformResourceFactory(resourceSchemaRepository)
 
-	err := remote.Activate(opts.To, alerter, providerLibrary, supplierLibrary, scanProgress, resourceSchemaRepository, resFactory)
+	err := remote.Activate(opts.To, opts.ProviderVersion, alerter, providerLibrary, supplierLibrary, scanProgress, resourceSchemaRepository, resFactory)
 	if err != nil {
 		return err
 	}
--- a/pkg/cmd/scan.go
+++ b/pkg/cmd/scan.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"os"
 	"os/signal"
+	"regexp"
 	"strings"
 	"syscall"
 
@@ -75,6 +76,12 @@ func NewScanCmd() *cobra.Command {
 				opts.Filter = expr
 			}
 
+			providerVersion, _ := cmd.Flags().GetString("tf-provider-version")
+			if err := validateTfProviderVersionString(providerVersion); err != nil {
+				return err
+			}
+			opts.ProviderVersion = providerVersion
+
 			opts.Quiet, _ = cmd.Flags().GetBool("quiet")
 			opts.DisableTelemetry, _ = cmd.Flags().GetBool("disable-telemetry")
 
@@ -136,7 +143,7 @@ func NewScanCmd() *cobra.Command {
 		"Terraform Cloud / Enterprise API token.\n"+
 			"Only used with tfstate+tfcloud backend.\n",
 	)
-	fl.StringVar(&opts.ProviderVersion,
+	fl.String(
 		"tf-provider-version",
 		"",
 		"Terraform provider version to use.\n",
@@ -345,4 +352,14 @@ func parseOutputFlag(out string) (*output.OutputConfig, error) {
 		Key:     o,
 		Options: options,
 	}, nil
+}
+
+func validateTfProviderVersionString(version string) error {
+	if version == "" {
+		return nil
+	}
+	if match, _ := regexp.MatchString("^\\d+\\.\\d+\\.\\d+$", version); !match {
+		return errors.Errorf("Invalid version argument %s, expected a valid semver string (e.g. 2.13.4)", version)
+	}
+	return nil
 }
--- a/pkg/terraform/provider_installer_test.go
+++ b/pkg/terraform/provider_installer_test.go
@@ -8,6 +8,8 @@ import (
 	"testing"
 
 	"github.com/cloudskiff/driftctl/mocks"
+	terraformError "github.com/cloudskiff/driftctl/pkg/terraform/error"
+	"github.com/stretchr/testify/mock"
 
 	"github.com/stretchr/testify/assert"
 )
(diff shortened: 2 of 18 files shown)


sjourdan (changes requested):
Running driftctl by default without any option should download the latest provider release version. Should it be in this PR or in another one? 

```shell
~/.../cloudskiff/tf-aws ( master ?:4 ⊘ )
❯  rm -rf ~/.driftctl

~/.../cloudskiff/tf-aws ( master ?:4 ⊘ )
❯  driftctl-0.10.0 scan 
[...]
DEBU[0000] New provider library created
DEBU[0000] provider not found, downloading ... path=/home/sjourdan/.driftctl/plugins/linux_amd64/terraform-provider-aws_v3.19.0
Downloading terraform provider: aws
DEBU[0000] Downloading provider path=/home/sjourdan/.driftctl/plugins/linux_amd64 url=https://releases.hashicorp.com/terraform-provider-aws/3.19.0/terraform-provider-aws_3.19.0_linux_amd64.zip
DEBU[0004] Decompressing archive dst=/home/sjourdan/.driftctl/plugins/linux_amd64 src=/tmp/terraform-provider104562435
DEBU[0005] Download successful
[...]
```

eliecharra:
> Running driftctl by default without any option should download the latest provider release version

I don't think it's a good idea to always use the latest version, in this PR we fix the version to the previously hardcoded ones. We should discuss about always use the latest version but this can cause uncontrolled behaviors imho

sjourdan:
Would it take long to add some minimal checks for the content of `DCTL_TF_PROVIDER_VERSION`? like at least it looks like x.y.z? 

```shell
❯  LOG_LEVEL=debug DCTL_TF_PROVIDER_VERSION=\..\ AWS_PROFILE=cloudskiff AWS_REGION=eu-west-3 driftctl-0.10.0 scan
DEBU[0000] Bound environment variable to flag env=tf_provider_version flag=tf-provider-version value=.. 
[...]
DEBU[0000] provider not found, downloading ... path=/home/sjourdan/.driftctl/plugins/linux_amd64/terraform-provider-aws_v.. 
Downloading terraform provider: aws
DEBU[0000] Downloading provider path=/home/sjourdan/.driftctl/plugins/linux_amd64 url=https://releases.hashicorp.com/terraform-provider-aws/.. AWS_PROFILE=cloudskiff/terraform-provider-aws_.. AWS_PROFILE=cloudskiff_linux_amd64.zip
unsuccessful request to https://releases.hashicorp.com/terraform-provider-aws/.. AWS_PROFILE=cloudskiff/terraform-provider-aws_.. AWS_PROFILE=cloudskiff_linux_amd64.zip: 403 Forbidden
```

pkg/terraform/provider_installer_test.go
```diff
@@ -147,3 +143,40 @@ func TestProviderInstallerInstallAlreadyExistButIsDirectory(t *testing.T) {
 	)
 
 }
+
+// Ensure that if a provider exist with a postfix (_x5) we properly detect it
```
wbeuil, line 147: `exists`

eliecharra:
> Would it take long to add some minimal checks for the content of DCTL_TF_PROVIDER_VERSION? like at least it looks like x.y.z?

No that's very easy, I need to handle a better validation of the input and also add a 404 / 403 error handling if the version does not exist

wbeuil:
> > Running driftctl by default without any option should download the latest provider release version
> 
> I don't think it's a good idea to always use the latest version, in this PR we fix the version to the previously hardcoded ones. We should discuss about always use the latest version but this can cause uncontrolled behaviors imho

I understand why you would not want to always download the latest version of the provider but we should speak/find a way to bump sometimes the provider version and not be stuck at a former version. For that we could use our acc tests to act as our decider. At the end when we will be ok with the way we bump the version, let's put it in the docs for our user to understand our reasoning.

pkg/cmd/scan.go
```diff
@@ -341,3 +353,13 @@ func parseOutputFlag(out string) (*output.OutputConfig, error) {
 		Options: options,
 	}, nil
 }
+
+func validateTfProviderVersionString(version string) error {
+	if version == "" {
+		return nil
+	}
+	if match, _ := regexp.MatchString("^\\d+\\.\\d+\\.\\d+$", version); !match {
+		return errors.Errorf("Invalid version argument %s, expected a valid semver string (e.g: 2.13.4)", version)
```
wbeuil, line 362: `e.g. 2.13.4` the convention is the dot after each letter and then the examples

wbeuil (changes requested):
Tested it with terraform provider aws 2.20.0 3.19.0 3.20.0 3.30.0 3.44.0 3.45.0.

Seems good to me.

Wonder what to do if someone tries to use driftctl with resource in one version and tf-provider in another and we find drifts.

Example I have: 

```
Found changed resources:
    - driftctl_assume_role (aws_iam_role):
        + inline_policy: <nil> => [map[name:driftctl_policy policy:{
  "Version": "2012-10-17",
  "Statement": [
    {
      "Sid": "",
      "Effect": "Allow",
      "Action": [
        "sqs:ListQueues",
        "sqs:ListQueueTags",
        "sqs:GetQueueAttributes",
        "sns:ListTopics",
        ...
      ],
      "Resource": "*"
    }
  ]
}]]
Found 2 resource(s)
 - 100% coverage
 - 2 covered by IaC
 - 0 not covered by IaC
 - 0 missing on cloud provider
 - 1/2 changed outside of IaC
```

Only for tf-provider >= 3.30.0

eliecharra:
> Wonder what to do if someone tries to use driftctl with resource in one version and tf-provider in another and we find drifts.

I don't know if we can do something for theses cases.

wbeuil (approved):

sjourdan:
LGTM! 

## Invalid version argument

- `DCTL_TF_PROVIDER_VERSION=a` 
- `DCTL_TF_PROVIDER_VERSION=1..1.1`
- `DCTL_TF_PROVIDER_VERSION=a.b.c`
- `DCTL_TF_PROVIDER_VERSION=1`
- `DCTL_TF_PROVIDER_VERSION=٣`
- `DCTL_TF_PROVIDER_VERSION=٣.٣.٣`
- `DCTL_TF_PROVIDER_VERSION=\..\`

## Provider does not exist: 

- `DCTL_TF_PROVIDER_VERSION=199999999999999999999999999999999999999999999999999999.29999.999999`
- `DCTL_TF_PROVIDER_VERSION=latest`
- `DCTL_TF_PROVIDER_VERSION=666`

## Works

- `VERSION=3.0.0 DCTL_TF_PROVIDER_VERSION=${VERSION}`

sjourdan (approved):
LGTM

pkg/cmd/scan.go
```diff
@@ -341,3 +353,13 @@ func parseOutputFlag(out string) (*output.OutputConfig, error) {
...
 	}, nil
 }
+
+func validateTfProviderVersionString(version string) error {
+	if version == "" {
+		return nil
+	}
+	if match, _ := regexp.MatchString("^\\d+\\.\\d+\\.\\d+$", version); !match {
+		return errors.Errorf("Invalid version argument %s, expected a valid semver string (e.g. 2.13.4)", version)
+	}
```
sundowndev, line 363: Can't you use go-version to check the version here? https://github.com/cloudskiff/driftctl/blob/main/pkg/iac/terraform/state/versions.go#L6

sundowndev (changes requested):

pkg/cmd/scan.go
eliecharra: I was wondering if we can use this lib but I don't think validation match what we really want there https://github.com/hashicorp/go-version/blob/master/version_test.go#L49

The simple regex was more relevant IMHO for this usage.

sundowndev (approved):

PR #589, opened 2021-06-04
Fix crash when capacity is set to 0 in cache

sundowndev:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | yes
| 🚀 New feature?   | no
| ⚠ Deprecations?   | no
| ❌ BC Break       | no
| 🔗 Related issues | #...
| ❓ Documentation  | no 

## Description

In some cases, we may want to not use the cache although it's required. It may be for testing purposes or simply because of a technical limitation. We already use a capacity of 0 in supplier unit tests, because we don't need to cache values. For now, a capacity of 0 result in a nil pointer dereference error. This PR fixes the issue simply by ignoring keys when capacity is set to 0.

### DIFF
--- a/pkg/remote/cache/cache_test.go
+++ b/pkg/remote/cache/cache_test.go
@@ -69,4 +69,31 @@ func TestCache(t *testing.T) {
 		}
 		assert.Equal(t, 5, cache.Len())
 	})
+
+	t.Run("should ignore keys when capacity is 0", func(t *testing.T) {
+		keys := []struct {
+			key   string
+			value interface{}
+		}{
+			{
+				"test",
+				[]string{"slice"},
+			},
+			{
+				"test",
+				[]string{},
+			},
+			{
+				"test2",
+				[]resource.FakeResource{},
+			},
+		}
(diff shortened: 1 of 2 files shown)


lotoussa (dismissed):

pkg/remote/cache/cache_test.go
```diff
@@ -69,4 +69,39 @@ func TestCache(t *testing.T) {
...
+				[]resource.FakeResource{},
+				false,
+				nil,
+			},
+		}
+		cache := New(0)
+
+		for _, k := range keys {
+			assert.Equal(t, k.overridden, cache.Put(k.key, k.value))
+			assert.Equal(t, k.expectedValue, cache.Get(k.key))
```
wbeuil, line 103: No need to have an `overriden` bool in your struct since with a cache of cap 0 it must be everytime `false`. Same for `expectedValue`.

wbeuil (changes requested):

wbeuil (approved):

Issue #160, opened 2021-01-28
Supplier interface could have a method to retrieve supplier supported type


moadibfr:
**Description**
We use their type in their own code an also 
 > It will be usefull in the future to avoid run filtered suppliers.

This come from https://github.com/cloudskiff/driftctl/pull/152#discussion_r565977250

wbeuil:
Before starting, in what way we would use this method ?

PR #605, opened 2021-06-08
add suppliedtype to suppliers

moadibfr:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | no
| 🚀 New feature?   | no
| ⚠ Deprecations?   | no
| ❌ BC Break       | no
| 🔗 Related issues | #160 
| ❓ Documentation  | no 

## Description
add suppliedtype to suppliers


pkg/remote/github/init.go
```diff
@@ -35,11 +35,16 @@ func Init(version string, alerter *alerter.Alerter,
...
-	supplierLibrary.AddSupplier(NewGithubTeamMembershipSupplier(provider, repository, deserializer))
-	supplierLibrary.AddSupplier(NewGithubBranchProtectionSupplier(provider, repository, deserializer))
+	var githubSuppliers []resource.SimpleTypeSupplier
+	githubSuppliers = append(githubSuppliers, NewGithubRepositorySupplier(provider, repository, deserializer))
+	githubSuppliers = append(githubSuppliers, NewGithubTeamSupplier(provider, repository, deserializer))
+	githubSuppliers = append(githubSuppliers, NewGithubMembershipSupplier(provider, repository, deserializer))
+	githubSuppliers = append(githubSuppliers, NewGithubTeamMembershipSupplier(provider, repository, deserializer))
+	githubSuppliers = append(githubSuppliers, NewGithubBranchProtectionSupplier(provider, repository, deserializer))
+
+	for _, supplier := range githubSuppliers {
```
eliecharra, line 45: Why not using `supplierLibrary.Suppliers()` to retrieve the list of suppliers instead of building a slice ?

pkg/resource/supplier.go
```diff
@@ -1,10 +1,15 @@
 package resource
 
-// Resource Supplier supply the list of resource.Resource, its the front to retrieve remote resources
+// Supplier Resource Supplier supply the list of resource.Resource, its the front to retrieve remote resources
 type Supplier interface {
 	Resources() ([]Resource, error)
 }
 
+type SimpleTypeSupplier interface {
+	Supplier
```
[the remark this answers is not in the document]
moadibfr: We have `ChainSupplier` and `Scanner` that implement supplier so I thought it would be easier to have a separate interface for this need

pkg/remote/github/init.go
moadibfr: It will be filtered in next PR before being added to supplierLibrary

pkg/resource/supplier.go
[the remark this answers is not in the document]
wbeuil: I'm good with having a new interface just for this need.

pkg/remote/github/init.go
wbeuil: Do you mean that the filtering will work as follow ?

If I only want to scan s3 buckets, only the s3 supplier will be inside the supplier library ?

eliecharra:
Switching this one to draft since we are reworking supplied for the deep mode in #619

Issue #456, opened 2021-04-21
Filtering should be considered prior to scans 


jackson-theisen-bind:
**Description**
When using filter rules, a scan still tries to look at resources that should be filtered. 

**Environment**
 - OS: macOS 
 - driftctl version: `v0.7.1` 

**How to reproduce**

Create an ec2 instance, remove the ec2 actions from https://docs.driftctl.com/0.7.1/providers/aws/authentication/#least-privileged-policy on the driftctl user's policy, then perform a scan with a filter to exclude ec2 instances:
```
$ driftctl scan --from tfstate+s3://foo/bar  --filter 'Type!="aws_instance"'
ERRO[0010] Unable to decode resource from state name=instance type=aws_instance
```

**Possible Solution**

Rewrite the logic so filtering occurs prior to a scan, not just against the results. 

**Additional context**

eliecharra:
Hi @jackson-theisen-bind thanks for your interest in driftctl 🙏🏻 

This is something we have in mind for the next releases since a couple of weeks. It will also lead to performance and rate limit optimization.

We currently have a lot of work in progress for `v0.8.0` and `v0.9.0` (a big core refactoring is in progress) so we do not have a deadline for now, but it will probably be the next big feature we'll work on.

sjourdan:
This is also linked to a correct realization of https://github.com/cloudskiff/driftctl/issues/468

eliecharra:
This is currently blocked by deep mode

eliecharra:
We will work on it but only with driftignore compatibility.

sjourdan:
LGTM

Nothing is scanned with a `driftignore` ignoring everything. 

- < 2s with 0.14
- > 8s with 0.13

```sh
❯  driftctl version
v0.13.0
❯  time driftctl scan --output json://stdout 
{
        "summary": {
                "total_resources": 0,
                "total_changed": 0,
                "total_unmanaged": 0,
                "total_missing": 0,
                "total_managed": 0
        },
        "managed": null,
        "unmanaged": null,
        "missing": null,
        "differences": null,
        "coverage": 0,
        "alerts": null
}
real    0m7.882s
user    0m1.036s
sys     0m0.201s


❯  time LOG_LEVEL=error driftctl-pre scan --output json://stdout 
{
        "summary": {
                "total_resources": 0,
                "total_changed": 0,
                "total_unmanaged": 0,
                "total_missing": 0,
                "total_managed": 0
        },
        "managed": null,
[...]

PR #619, opened 2021-06-11
Split suppliers to add deep mode

eliecharra:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | no
| 🚀 New feature?   | yes
| ⚠ Deprecations?   | no
| ❌ BC Break       | no
| 🔗 Related issues | #572 
| ❓ Documentation  | no

## Description

@moadibfr @wbeuil ⚠️ Attentive architectural review and naming required please 🙏🏻 
I reworker a little bit the `scanner.go` too

PR #620, opened 2021-06-11
driftignore prevent resource fetch

moadibfr:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | yes
| 🚀 New feature?   | yes
| ⚠ Deprecations?   | no
| ❌ BC Break       | no
| 🔗 Related issues | #456 
| ❓ Documentation  | yes 
## Description

driftignore prevent resource fetch


pkg/resource/supplier.go
```diff
@@ -1,10 +1,15 @@
 package resource
 
-// Resource Supplier supply the list of resource.Resource, its the front to retrieve remote resources
+// Supplier Resource Supplier supply the list of resource.Resource, its the front to retrieve remote resources
```
lotoussa, line 3: ?

lotoussa (approved):
it seems to me that everything is fine

pkg/resource/supplier.go
moadibfr: I'll revert this, my linter does not like when comment don't start with the object name as it does not respect godoc... https://blog.golang.org/godoc

wbeuil:
Focus is on the `deep mode` #619, let's put it in draft for now

moadibfr:
need too much rework we should consider writing differently for the new arch that came for deepmode

Issue #555, opened 2021-05-25
Can we change the ~/.driftctl location?


AzySir:
**Description**
  I am pushing the boundaries and currently trying to run driftctl via a lambda. But right now I am getting the below issue. 

```
  mkdir /home/sbx_user1051: read-only file system
```

I believe this is happening due to the fact that .driftctl is trying to access 

`/Users/<user>/.driftctl/plugins/darwin_amd64/terraform-provider-aws_v3.19.0_x5` on a lambda which is impossible as it can only use the `/tmp` folder. 

**Environment**
 - OS: AWS Lambda
 - driftctl version: v0.8.0

**How to reproduce**
Download driftctl in lambda and run scan command

**Possible Solution**
1) Take into account an Environment variable similar to the example on the terraform website for - 

```
export TF_PLUGIN_CACHE_DIR="$HOME/.terraform.d/plugin-cache"
```

2)  Take into a flag `--plugin_cache=` to set the cache directory

a-vorobiev:
If it helps, as a workaround I set `HOME` environment variable to `/tmp/` when invoking `driftctl` in a Lambda function:

```golang
cmd := exec.Command(lambdaPath+`/driftctl`, params...)
cmd.Env = append(
	os.Environ(),
	"HOME=/tmp",
)
```

eliecharra:
Thanks for the workaround @a-vorobiev 🙏🏻 

We should work on this and allow to override the default path.

lotoussa:
*Create a tutorial to use driftctl in a lambda*

sjourdan:
LGTM 

```
❯  DCTL_TF_PROVIDER_VERSION=3.44.0 driftctl-0.10.0 scan --config-dir ./tmp
[...]
❯  tree tmp/.driftctl/
tmp/.driftctl/
└── plugins
    └── linux_amd64
        └── terraform-provider-aws_v3.44.0_x5
```

Just added a companion issue: https://github.com/cloudskiff/driftctl/issues/669

PR #629, opened 2021-06-12
Add config-dir flag to change .driftctl location

lotoussa:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | no
| 🚀 New feature?   | yes
| ⚠ Deprecations?   | no
| ❌ BC Break       | no
| 🔗 Related issues | #555 
| ❓ Documentation  | yes - i must add this flag to the documentation

## Description

I add a flag 'config-dir' to override the .driftctl default path

### DIFF
--- a/pkg/cmd/scan.go
+++ b/pkg/cmd/scan.go
@@ -9,6 +9,7 @@ import (
 	"syscall"
 
 	"github.com/cloudskiff/driftctl/pkg/telemetry"
+	"github.com/mitchellh/go-homedir"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
@@ -155,9 +156,11 @@ func NewScanCmd() *cobra.Command {
 		false,
 		"Includes cloud provider service-linked roles (disabled by default)",
 	)
+
+	configDir, _ := homedir.Dir()
 	fl.String(
 		"config-dir",
-		"",
+		configDir,
 		".driftctl path to use.\n"+
 			"Default is home directory.\n",
 	)
--- a/pkg/cmd/scan.go
+++ b/pkg/cmd/scan.go
@@ -161,8 +161,7 @@ func NewScanCmd() *cobra.Command {
 	fl.String(
 		"config-dir",
 		configDir,
-		".driftctl path to use.\n"+
-			"Default is home directory.\n",
+		"Directory path that driftctl uses for configuration.\n",
 	)
 
 	return cmd
--- a/pkg/cmd/scan.go
+++ b/pkg/cmd/scan.go
@@ -157,7 +157,10 @@ func NewScanCmd() *cobra.Command {
 		"Includes cloud provider service-linked roles (disabled by default)",
 	)
 
-	configDir, _ := homedir.Dir()
+	configDir, err := homedir.Dir()
+	if err != nil {
+		configDir = os.TempDir()
+	}
 	fl.String(
 		"config-dir",
 		configDir,
--- a/pkg/cmd/scan.go
+++ b/pkg/cmd/scan.go
@@ -85,6 +85,8 @@ func NewScanCmd() *cobra.Command {
 			opts.Quiet, _ = cmd.Flags().GetBool("quiet")
 			opts.DisableTelemetry, _ = cmd.Flags().GetBool("disable-telemetry")
 
+			opts.ConfigDir, _ = cmd.Flags().GetString("config-dir")
+
 			return nil
 		},
 		RunE: func(cmd *cobra.Command, args []string) error {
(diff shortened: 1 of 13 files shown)


pkg/cmd/scan.go
```diff
@@ -153,6 +155,12 @@ func NewScanCmd() *cobra.Command {
 		false,
 		"Includes cloud provider service-linked roles (disabled by default)",
 	)
+	fl.String(
+		"config-dir",
+		"",
+		".driftctl path to use.\n"+
+			"Default is home directory.\n",
+	)
```
sundowndev, line 163: You can set the default value here directly, using `homedir.Dir()` or `os.Getenv("HOME")`

sundowndev (changes requested):

pkg/cmd/scan.go
lotoussa: i'm on it, it can be cool to avoid the logic in provider_installer.go

pkg/cmd/scan.go
```diff
@@ -154,6 +157,15 @@ func NewScanCmd() *cobra.Command {
 		"Includes cloud provider service-linked roles (disabled by default)",
 	)
 
+	configDir, _ := homedir.Dir()
+	// Error handled in getProviderDirectory() pkg/terraform/provider_installer.go
```
sundowndev, line 161: Comment to remove

pkg/cmd/scan.go
```diff
@@ -154,6 +157,15 @@ func NewScanCmd() *cobra.Command {
 		"Includes cloud provider service-linked roles (disabled by default)",
 	)
 
+	configDir, _ := homedir.Dir()
```
sundowndev, line 160: Are you sure it's safe to ignore the error here?

sundowndev (dismissed):

pkg/cmd/scan.go
```diff
@@ -154,6 +157,14 @@ func NewScanCmd() *cobra.Command {
...
+	fl.String(
+		"config-dir",
+		configDir,
+		".driftctl path to use.\n"+
+			"Default is home directory.\n",
```
wbeuil, line 165: I find this sentence really confusing.

I'd rather read: 
`Directory path that driftctl uses for configuration`
`Default is the home directory`

wbeuil (changes requested):
What if the `config-dir` is not a directory but a file ?

pkg/cmd/scan.go
sundowndev: FYI Cobra already handles default values in help message. You shoudn't have to write it manually.

pkg/cmd/scan.go
lotoussa: Yes the test can demonstrate multiple situations, and if configDir is "" et err != nil, this variable will be handle after that to call tmpDir()

lotoussa:
> What if the `config-dir` is not a directory but a file ?

DEBU[0024] Bound environment variable to flag env=disable_telemetry flag=disable-telemetry value=true
DEBU[0024] New provider library created
DEBU[0024] Starting gRPC client alias=eu-west-3
fork/exec testconfigdir/.driftctl/plugins/darwin_amd64/terraform-provider-aws_v3.19.0: not a directory
Exiting.

pkg/cmd/scan.go
sundowndev: Not sure to understand, I don't see any fallback to tmpDir() in the diff

lotoussa: pkg/terraform/provider_installer.go:78

wbeuil: I think what @sundowndev meant is that for the default variable just below `configDir` if `homedir.Dir()` returns an error, we won't have any default displayed inside the `--help` message. Is that it ?

sundowndev: Could you put the fallback here instead ? I think it's confusing that it's not close to the definition

sundowndev (changes requested):
Almost good, I think you forgot to remove this statement in `pkg/terraform/provider_installer.go` : 

```go
if p.homeDir == "" {
	p.homeDir = os.TempDir()
}
```

sundowndev (approved):

wbeuil (approved):
LGTM

sundowndev (dismissed):

sundowndev (approved):

sjourdan:
LGTM 

```
❯  DCTL_TF_PROVIDER_VERSION=3.44.0 driftctl-0.10.0 scan --config-dir ./tmp
[...]
❯  tree tmp/.driftctl/
tmp/.driftctl/
└── plugins
    └── linux_amd64
        └── terraform-provider-aws_v3.44.0_x5
```

Just added a companion issue: https://github.com/cloudskiff/driftctl/issues/669

Issue #646, opened 2021-06-15
Display Terraform provider version used for scan


sjourdan:
**Description**

_As a driftctl user, 
I want to know the version of the provider used by driftctl to scan my account against my tfstate
So that I can keep my Terraform config in sync with driftctl scan_

With 0.10, the provider version is no more hardcoded: users can now specify which one to use (ie: `driftctl scan --tf-provider-version=3.44.0`). 

But there has to be a default, and we decided to keep that default to 3.19 for AWS for now, so as to not break any existing automated run for current users. 

But existing and new users may not know about the feature and have a suboptimal overall experience (false positives).

The goal here is to notify the user that 

a) a specific version of the provider was used to scan
b) the user can change it to match the correct one

Adding this information should not break the JSON output (console output only).

**Example**

A default run: 

```shell 
$ driftctl scan 
[...]
Found 25 resource(s)
 - 68% coverage
 - 17 covered by IaC
 - 8 not covered by IaC
 - 0 missing on cloud provider
 - 2/17 changed outside of IaC
Provider version used to scan: 3.19.0 (default). Use --tf-provider-version to specify one. 
``` 

A run with a specific version: 

```
$ driftctl scan --tf-provider-version=3.44.0
[...]
Found 25 resource(s)
 - 68% coverage
 - 17 covered by IaC
 - 8 not covered by IaC
 - 0 missing on cloud provider
 - 2/17 changed outside of IaC
Provider version used to scan: 3.44.0
``` 

A run with a specific version using the env var: 

```
$ DTCL_TF_PROVIDER_VERSION=2.70.0 driftctl scan 
[...]
Found 25 resource(s)
 - 68% coverage
 - 17 covered by IaC
 - 8 not covered by IaC
 - 0 missing on cloud provider
 - 2/17 changed outside of IaC
Provider version used to scan: 2.70.0
```

sjourdan:
LGTM!

```shell
$ DCTL_TF_PROVIDER_VERSION=3.44.0 driftctl-0.10.0 scan
Found 84 resource(s)
 - 96% coverage
 - 81 covered by IaC
 - 3 not covered by IaC
 - 0 missing on cloud provider
 - 1/81 changed outside of IaC
You have diffs on computed fields, check the documentation for potential false positive drifts
Provider version used to scan: 3.44.0. Use --tf-provider-version to use another version.

# Using default

$ DCTL_TF_PROVIDER_VERSION="" driftctl-0.10.0 scan
[...]
[...]

PR #658, opened 2021-06-17
Display terraform provider version in output

sundowndev:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | no
| 🚀 New feature?   | yes
| ⚠ Deprecations?   | no
| ❌ BC Break       | no
| 🔗 Related issues | #646
| ❓ Documentation  | no 

## Description

> As a driftctl user,
> I want to know the version of the provider used by driftctl to scan my account against my tfstate
> So that I can keep my Terraform config in sync with driftctl scan

This PR simply add a log message at the end of scan with the provider's version. Also the default value is now set in the scan command controller, rather than the AWS provider init function.

image

### DIFF
--- a/pkg/cmd/scan.go
+++ b/pkg/cmd/scan.go
@@ -9,6 +9,7 @@ import (
 	"syscall"
 
 	"github.com/cloudskiff/driftctl/pkg/telemetry"
+	"github.com/fatih/color"
 	"github.com/mitchellh/go-homedir"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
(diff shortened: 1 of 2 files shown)


lotoussa (approved):

pkg/cmd/scan.go
```diff
@@ -77,6 +78,9 @@ func NewScanCmd() *cobra.Command {
 			}
 
 			providerVersion, _ := cmd.Flags().GetString("tf-provider-version")
+			if providerVersion == "" {
+				providerVersion = aws.DefaultProviderVersion
```
eliecharra, line 82: Think of github too, this approach cannot work as we are multi cloud

eliecharra (changes requested):

pkg/cmd/scan.go
wbeuil: You were too fast for me, i was writing my comment ^^

sundowndev: oh you're right, I totally forgot about github 😅

eliecharra (approved):

Issue #630, opened 2021-06-14
aws_route false positive (route rable): carrier_gateway_id


sjourdan:
**Description**

```shell
    - r-rtb-029d6da9479e1da701080289494 (aws_route):
        Destination: 0.0.0.0/0, Table: rtb-029d6da9479e1da70
            + carrier_gateway_id: <nil> => ""

```

**Environment**
 - OS: F34
 - driftctl version: 0.10.0-pre
 - terraform generated state version: 1.0.0
 - aws tf provider: 3.44

**How to reproduce**

```hcl
resource "aws_route_table" "production" {
  vpc_id = aws_vpc.production.id

  route {
    cidr_block = "0.0.0.0/0"
    gateway_id = aws_internet_gateway.gw.id
  }

  tags = {
    Name = "Production Route Table"
  }
}

resource "aws_route_table_association" "a" {
  subnet_id      = aws_subnet.production-subnet-1.id
  route_table_id = aws_route_table.production.id
}
```

**Possible Solution**

**Additional context**

sjourdan:
Note: doesn't happen when scanning with 3.19 or 2.70 (with a similar 3.44 generated state)

PR #662, opened 2021-06-17
Fix aws_route carrier_gateway_id false positive

lotoussa:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | yes
| 🚀 New feature?   | no
| ⚠ Deprecations?   | no
| ❌ BC Break       | no
| 🔗 Related issues | #630 
| ❓ Documentation  | no

## Description

Fix aws_route false positive

### DIFF
--- a/pkg/resource/aws/aws_route.go
+++ b/pkg/resource/aws/aws_route.go
@@ -27,6 +27,7 @@ func initAwsRouteMetaData(resourceSchemaRepository resource.SchemaRepositoryInte
 		val.DeleteIfDefault("vpc_peering_connection_id")
 		val.DeleteIfDefault("destination_prefix_list_id")
 		val.DeleteIfDefault("instance_owner_id")
+		val.DeleteIfDefault("carrier_gateway_id")
 	})
 	resourceSchemaRepository.SetHumanReadableAttributesFunc(AwsRouteResourceType, func(res *resource.AbstractResource) map[string]string {
 		val := res.Attrs


pkg/resource/aws/aws_route.go
```diff
@@ -14,6 +14,7 @@ func initAwsRouteMetaData(resourceSchemaRepository resource.SchemaRepositoryInte
 	resourceSchemaRepository.SetNormalizeFunc(AwsRouteResourceType, func(res *resource.AbstractResource) {
 		val := res.Attrs
 		val.SafeDelete([]string{"timeouts"})
+		val.SafeDelete([]string{"carrier_gateway_id"})
```
eliecharra, line 17: This field shouldn't be ignored totally, we only want to avoid diff from `nil` to `""` and vice-versa.

The method `DeleteIfDefault` is more appropriate

eliecharra (changes requested):

pkg/resource/aws/aws_route.go
lotoussa: Right, fix done.

pkg/resource/aws/aws_route.go
```diff
@@ -27,6 +27,7 @@ func initAwsRouteMetaData(resourceSchemaRepository resource.SchemaRepositoryInte
 		val.DeleteIfDefault("vpc_peering_connection_id")
 		val.DeleteIfDefault("destination_prefix_list_id")
 		val.DeleteIfDefault("instance_owner_id")
+        val.DeleteIfDefault("carrier_gateway_id")
```
eliecharra, line 30: Indentation fix required

eliecharra (changes requested):

pkg/resource/aws/aws_route.go
lotoussa: yes i didn't make before push

Issue #615, opened 2021-06-11
A different authentication mechanism for S3 bucket


sjourdan:
# User Story

As a driftctl CLI user,
I want to use a dedicated authentication for S3
So that driftctl can access Terraform states stored on a different AWS account than the one to scan

# Acceptance Tests

## Criteria

**The user can use one AWS profile to access S3 and another for the scan**

- export a dedicated AWS profile (or any other AWS env var) for accessing S3 as read-only

```bash
export DCTL_S3_AWS_PROFILE="<aws_profile_name>" 
```

- export the usual driftctl AWS profile (read-only too)

```bash
export AWS_PROFILE="driftctl-ro"
```

- confirm the scan succeeds

```bash
driftctl scan --from tfstate+s3://mybucket/terraform.tfstate 
```

_Originally posted by @kuanhn in https://github.com/cloudskiff/driftctl/discussions/560_

sjourdan:
couldn't use the feature on 0.12-pre build `ee029e9-dev`

```
❯  AWS_REGION=us-west-2 DCTL_S3_AWS_PROFILE="615-s3-only" AWS_PROFILE="615-iam-only" driftctl-pre scan --from tfstate+s3://615-usw2-states/**/*.tfstate 
DEBU[0000] Bound environment variable to flag env=disable_telemetry flag=disable-telemetry value=true
DEBU[0000] New provider library created
DEBU[0000] Found existing provider path=/home/sjourdan/.driftctl/plugins/linux_amd64/terraform-provider-aws_v3.19.0_x5
DEBU[0000] Starting gRPC client alias=us-west-2
DEBU[0001] New gRPC client started alias=us-west-2
DEBU[0001] Terraform provider initialized alias=us-west-2 name=aws
DEBU[0001] Found IAC supplier backend=s3 path=615-usw2-states/**/*.tfstate supplier=tfstate
INFO[0001] Start reading IaC
DEBU[0002] Stopping ParallelRunner
Scanned states (0)      
DEBU[0002] Closing providers
DEBU[0002] Closing provider key=aws
DEBU[0002] Closing gRPC client alias=us-west-2
AccessDenied: Access Denied
	status code: 403, request id: Q6YQQYKED4XHVV6X, host id: WOlGcILAUFS01nEvinXD/CsdkUbXgiWYkDl/nBaFZruuKdwYE7aizPRu/wAXDtFPwSxKc6kMEQ4=

```

[...]

PR #671, opened 2021-06-21
Add a way to override s3 config/profile with env var

lotoussa:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | no
| 🚀 New feature?   | yes
| ⚠ Deprecations?   | no
| ❌ BC Break       | no
| 🔗 Related issues | #615 
| ❓ Documentation  | yes - https://github.com/cloudskiff/driftctl-docs/pull/83

## Description

Creation of a struct to allow override of env var for state enumerator.
I'm not sure with this implementation, feel free to discuss about it.

### DIFF
--- a/pkg/envproxy/env_proxy.go
+++ b/pkg/envproxy/env_proxy.go
@@ -6,9 +6,9 @@ import (
 )
 
 type EnvProxy struct {
-	varPrefix  string
-	varPattern string
-	DefaultEnv map[string]string
+	fromPrefix string
+	toPrefix   string
+	defaultEnv map[string]string
 }
 
 func NewEnvProxy() *EnvProxy {
@@ -18,35 +18,35 @@ func NewEnvProxy() *EnvProxy {
 		envMap[tmp[0]] = tmp[1]
 	}
 	return &EnvProxy{
-		DefaultEnv: envMap,
+		defaultEnv: envMap,
 	}
 }
 
-func (s *EnvProxy) SetProxy(prefix, pattern string) {
-	s.varPrefix = prefix
-	s.varPattern = pattern
+func (s *EnvProxy) SetProxy(fromPrefix, toPrefix string) {
+	s.fromPrefix = fromPrefix
+	s.toPrefix = toPrefix
 }
 
 func (s *EnvProxy) Apply() {
-	if s.varPrefix == "" || s.varPattern == "" {
+	if s.fromPrefix == "" || s.toPrefix == "" {
 		return
 	}
-	for key, value := range s.DefaultEnv {
-		if strings.HasPrefix(key, s.varPrefix) {
-			key = strings.Replace(key, s.varPrefix, s.varPattern, 1)
+	for key, value := range s.defaultEnv {
+		if strings.HasPrefix(key, s.fromPrefix) {
+			key = strings.Replace(key, s.fromPrefix, s.toPrefix, 1)
 			os.Setenv(key, value)
 		}
 	}
 }
 
 func (s *EnvProxy) Restore() {
-	if s.varPrefix == "" || s.varPattern == "" {
+	if s.fromPrefix == "" || s.toPrefix == "" {
 		return
 	}
-	for key, value := range s.DefaultEnv {
-		if strings.HasPrefix(key, s.varPrefix) {
-			key = strings.Replace(key, s.varPrefix, s.varPattern, 1)
-			value = s.DefaultEnv[key]
+	for key, value := range s.defaultEnv {
+		if strings.HasPrefix(key, s.fromPrefix) {
+			key = strings.Replace(key, s.fromPrefix, s.toPrefix, 1)
+			value = s.defaultEnv[key]
 		}
 		os.Setenv(key, value)
 	}
--- a/pkg/envproxy/env_proxy.go
+++ b/pkg/envproxy/env_proxy.go
@@ -11,20 +11,17 @@ type EnvProxy struct {
 	defaultEnv map[string]string
 }
 
-func NewEnvProxy() *EnvProxy {
+func NewEnvProxy(fromPrefix, toPrefix string) *EnvProxy {
 	envMap := map[string]string{}
 	for _, variable := range os.Environ() {
 		tmp := strings.SplitN(variable, "=", 2)
 		envMap[tmp[0]] = tmp[1]
 	}
 	return &EnvProxy{
+		fromPrefix: fromPrefix,
+		toPrefix:   toPrefix,
 		defaultEnv: envMap,
 	}
-}
-
-func (s *EnvProxy) SetProxy(fromPrefix, toPrefix string) {
-	s.fromPrefix = fromPrefix
-	s.toPrefix = toPrefix
 }
 
 func (s *EnvProxy) Apply() {
--- a/pkg/envproxy/env_proxy.go
+++ b/pkg/envproxy/env_proxy.go
@@ -1,0 +1,53 @@
+package envproxy
+
+import (
+	"os"
+	"strings"
+)
+
+type EnvProxy struct {
+	varPrefix  string
+	varPattern string
+	DefaultEnv map[string]string
+}
+
+func NewEnvProxy() *EnvProxy {
+	envMap := map[string]string{}
+	for _, variable := range os.Environ() {
+		tmp := strings.SplitN(variable, "=", 2)
+		envMap[tmp[0]] = tmp[1]
+	}
+	return &EnvProxy{
+		DefaultEnv: envMap,
+	}
+}
+
+func (s *EnvProxy) SetProxy(prefix, pattern string) {
+	s.varPrefix = prefix
+	s.varPattern = pattern
+}
+
+func (s *EnvProxy) Apply() {
+	if s.varPrefix == "" || s.varPattern == "" {
+		return
+	}
+	for key, value := range s.DefaultEnv {
+		if strings.HasPrefix(key, s.varPrefix) {
+			key = strings.Replace(key, s.varPrefix, s.varPattern, 1)
+			os.Setenv(key, value)
+		}
+	}
+}
+
+func (s *EnvProxy) Restore() {
+	if s.varPrefix == "" || s.varPattern == "" {
+		return
+	}
+	for key, value := range s.DefaultEnv {
+		if strings.HasPrefix(key, s.varPrefix) {
+			key = strings.Replace(key, s.varPrefix, s.varPattern, 1)
+			value = s.DefaultEnv[key]
+		}
+		os.Setenv(key, value)
+	}
+}
--- a/pkg/envproxy/env_proxy_test.go
+++ b/pkg/envproxy/env_proxy_test.go
@@ -37,6 +37,12 @@ func TestEnvProxy(t *testing.T) {
 			initialEnv:  []string{"TEST_DCTL_S3_PROFILE=test_dctl_s3_profile", "TEST_AWS_PROFILE=test_aws_profile"},
 			modifiedEnv: []string{"TEST_DCTL_S3_PROFILE=test_dctl_s3_profile", "TEST_AWS_PROFILE=test_aws_profile"},
 		},
+        {
+            name:        "Without initialEnv",
+            proxyArgs:   []string{"TEST_DCTL_S3_", "TEST_AWS_"},
+            initialEnv:  []string{},
+            modifiedEnv: []string{},
+        },
 	}
 
 	for _, tt := range tests {
--- a/pkg/envproxy/env_proxy_test.go
+++ b/pkg/envproxy/env_proxy_test.go
@@ -4,95 +4,77 @@ import (
 	"os"
 	"strings"
 	"testing"
-
-	"github.com/stretchr/testify/assert"
 )
 
-func TestEnvProxy_Apply(t *testing.T) {
+func TestEnvProxy(t *testing.T) {
 	tests := []struct {
-		name      string
-		proxyArgs []string
-		modifier  bool
+		name        string
+		proxyArgs   []string
+		initialEnv  []string
+		modifiedEnv []string
 	}{
 		{
-			name:      "Without args on SetProxy",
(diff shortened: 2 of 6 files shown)


eliecharra (changes requested):
I don't think we should use flags for this. As we discussed in our planning schedule, we do not want to be able to override only AWS_PROFILE but any AWS SDK values https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-envvars.html (we forgot to update the issue according maybe). We should have a mechanism that pass every `DCTL_S3_XXXXX` to the s3 state reader client. WDYT @cloudskiff/engineering

sundowndev:
> I don't think we should use flags for this. As we discussed in our planning schedule, we do not want to be able to override only AWS_PROFILE but any AWS SDK values https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-envvars.html (we forgot to update the issue according maybe). We should have a mechanism that pass every `DCTL_S3_XXXXX` to the s3 state reader client. WDYT @cloudskiff/engineering

I agree. We need to think about S3 alternatives (minio, digitalocean, scaleway, ovh...), and other storage systems (gcs, azure blob, openstack swift, ...) we may introduce in the future. So the S3 env vars should allow the user to customize at least the access_key_id, secret_access_key, endpoint, and region.

pkg/iac/terraform/state/enumerator/s3.go
```diff
@@ -25,8 +26,27 @@ type S3Enumerator struct {
...
+
+	s3Config := aws.Config{}
+
+	if s3Region := os.Getenv("DCTL_S3_REGION"); s3Region != "" {
+		s3Config.Region = aws.String(s3Region)
+		if s3Endpoint := os.Getenv("DCTL_S3_ENDPOINT"); s3Endpoint != "" {
+			s3Config.Endpoint = aws.String(s3Endpoint)
+		}
+	}
+
```
sundowndev, line 45: Let me suggest a different approach: we could create a struct for those values with a struct tag `env` to define from which environment variable parse the value from. [example from configor](https://github.com/jinzhu/configor#usage)

### Example

```go
type S3EnumeratorConfig struct {
    Endpoint string `env:"DCTL_S3_ENDPOINT"`
    // ...
}

// ...

var s3Config S3EnumeratorConfig
load(&s3Config)

sess := session.Must(session.NewSessionWithOptions(session.Options{
		SharedConfigState: session.SharedConfigEnable,
		Profile:           s3Profile,
		Config:            aws.Config{
                Endpoint: aws.String(s3Config.Endpoint),
                // ...
        },
}))
```

This way we could reuse that logic for any other auth system, and easily add new config parameters.

WDYT @cloudskiff/engineering ?

eliecharra: I agree with this approach event if I would have preferred something less declarative and more flexible (with this approach we'll have to follow ever AWS SDK evolution). But I didn't get a better approach in mind. Maybe we can do something like in acc test where we override the whole env used to run TF operations ? 

Anyway I don't think this add a lot of value to introduce a config library only for this case, I'd rather prefer to have manual `os.GetEnv` in an `aws.Config` object directly.

sundowndev: > with this approach we'll have to follow ever AWS SDK evolution
I don't think we have any other option anyway, do we?

lotoussa: So do i try to create a structure like sundowndev recommandation ?

eliecharra: > > with this approach we'll have to follow ever AWS SDK evolution
> 
> I don't think we have any other option anyway, do we?

This approach can handle this :

> Maybe we can do something like in acc test where we override the whole env used to run TF operations ?

pkg/iac/terraform/state/enumerator/s3.go
```diff
@@ -15,19 +16,48 @@ import (
...
+				value = s.defaultEnv[key]
+			}
+			os.Setenv(key, value)
+		}
+	}
+}
+
 func NewS3Enumerator(config config.SupplierConfig) *S3Enumerator {
+	s3EnumConfig := NewS3EnumeratorConfig()
+	s3EnumConfig.setEnv(false)
```
sundowndev, line 56: I think you should set the prefix and destination prefix here, since it's directly related to this enumerator.

example (pseudo code)

```go
func NewS3Enumerator(p EnvProxy, config config.SupplierConfig) *S3Enumerator {
  p.SetProxy("DCTL_S3_", "AWS_")

  // ...
```

sundowndev (changes requested):
I think this fits more with what me and @eliecharra had in mind. Howether, I have few comments about the implementation. Instead of using `setEnv` I think you should pass the struct to `NewS3Enumerator` and set the prefix/pattern from there. Then, in the state reader, after calling `Enumerate()`, call a rollback/reset method to set the old values back in env.

What do you think?

pkg/iac/terraform/state/enumerator/s3.go
```diff
@@ -15,19 +16,48 @@ import (
...
 	sess := session.Must(session.NewSessionWithOptions(session.Options{
 		SharedConfigState: session.SharedConfigEnable,
 	}))
+	s3EnumConfig.setEnv(true)
```
sundowndev, line 60: You can't reset the values here because env vars are parsed later at enumeration, in the state reader

pkg/iac/terraform/state/enumerator/s3.go
lotoussa: yes, good point, that was my first implementation, i didn't think about reuse that logic

pkg/iac/terraform/state/enumerator/s3.go
lotoussa: as we said in peer review, env var are parsed here.

sundowndev (changes requested):

pkg/iac/terraform/state/enumerator/s3.go
sundowndev: @eliecharra what do you think of this new approach ?

pkg/iac/terraform/state/enumerator/state_enumerator.go
```diff
@@ -10,13 +13,49 @@ type StateEnumerator interface {
...
+	for _, variable := range os.Environ() {
+		tmp := strings.SplitN(variable, "=", 2)
+		envMap[tmp[0]] = tmp[1]
+	}
+	return &EnvProxy{
+		DefaultEnv: envMap,
+	}
+}
+
+func (s *EnvProxy) ReplaceEnv(resetMod bool) {
```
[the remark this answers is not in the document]
eliecharra: > also I think resetMod is not really explicit and should be renamed to resetDefault or useDefault

Totally agreed on this, we should separate reponsibilities.

I have two methods in mind :

```go
envProxy := NewEnvProxy("DCTL_S3_", "AWS_")
proxy.Apply()
// do something with overriden env 
proxy.Restore()
```

pkg/iac/terraform/state/enumerator/state_enumerator.go
```diff
@@ -10,13 +13,49 @@ type StateEnumerator interface {
 	Enumerate() ([]string, error)
 }
 
+type EnvProxy struct {
```
eliecharra, line 16: I think this should be put in another package as there is nothing related to IaC stuff. WDYT of `pkg/helpers` @sundowndev @lotoussa ? (even if I don't really like "helpers" naming since it can quickly become a hold-all ...)

pkg/iac/terraform/state/enumerator/state_enumerator.go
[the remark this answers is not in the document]
lotoussa: Yes, i'm on it

pkg/iac/terraform/state/enumerator/state_enumerator.go
lotoussa: This is convenient

eliecharra (commented):
@sundowndev @lotoussa Look like a nice approach, clean, reusable and flexible, I still have some minor remarks on `EnvProxy` interface and pkg location

pkg/iac/terraform/state/enumerator/state_enumerator.go
sundowndev: @eliecharra I prefer having a dedicated pkg for piece of code like this one, I think it fits more with the Go design patterns. But helpers can also be great, as long as it's not about extending every other packages (which can quickly become messy).

So I suggest we put this in `pkg/envproxy` or something like `pkg/iac/terraform/state/helpers`

pkg/iac/terraform/state/enumerator/s3.go
```diff
@@ -14,20 +14,17 @@ import (
...
 	config config.SupplierConfig
 	client s3iface.S3API
 }
 
-func NewS3Enumerator(config config.SupplierConfig) *S3Enumerator {
+func NewS3Enumerator(config config.SupplierConfig, envProxy *EnvProxy) *S3Enumerator {
+	envProxy.SetProxy("DCTL_S3_", "AWS_", false)
 	sess := session.Must(session.NewSessionWithOptions(session.Options{
 		SharedConfigState: session.SharedConfigEnable,
 	}))
```
sundowndev, line 26: You can initialize EnvProxy directly in NewS3Enumerator since you'll rollback in the same function

pkg/envproxy/env_proxy.go
```diff
@@ -0,0 +1,53 @@
...
+	for _, variable := range os.Environ() {
+		tmp := strings.SplitN(variable, "=", 2)
+		envMap[tmp[0]] = tmp[1]
+	}
+	return &EnvProxy{
+		DefaultEnv: envMap,
+	}
+}
+
+func (s *EnvProxy) SetProxy(prefix, pattern string) {
```
sundowndev, line 25: `from, to`

sundowndev (changes requested):
Good work 👍🏼 , I think testing needs a little refacto, but it shouldn't take too long

pkg/envproxy/env_proxy.go
```diff
@@ -0,0 +1,53 @@
...
+
+import (
+	"os"
+	"strings"
+)
+
+type EnvProxy struct {
+	varPrefix  string
+	varPattern string
+	DefaultEnv map[string]string
```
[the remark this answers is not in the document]
lotoussa: it's a mistake

pkg/envproxy/env_proxy_test.go
```diff
@@ -0,0 +1,98 @@
...
+		{
+			name:      "With no pattern on SetProxy",
+			proxyArgs: []string{"TEST_DCTL_S3_", ""},
+			modifier:  false,
+		},
+		{
+			name:      "With no prefix on SetProxy",
+			proxyArgs: []string{"", "TEST_AWS_"},
+			modifier:  false,
+		},
```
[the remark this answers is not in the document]
lotoussa: good point, i'm on it

pkg/envproxy/env_proxy_test.go
```diff
@@ -0,0 +1,80 @@
...
+			currentEnv = os.Environ()
+			if !compareEnv(currentEnv, tt.initialEnv) {
+				t.Errorf("Expected %v, got %v", tt.initialEnv, currentEnv)
+			}
+		})
+	}
+}
+
+func compareEnv(currentEnv, testEnv []string) bool {
+	isValid := 0
```
sundowndev, line 71: `if len(currentEnv) != testEnv {return false}`

sundowndev (changes requested):

pkg/envproxy/env_proxy_test.go
lotoussa: I can't because testEnv is only the field i want to test (2 in general) and currentEnv is os.Environ()

sundowndev: right, nevermind

pkg/iac/terraform/state/enumerator/s3.go
```diff
@@ -8,26 +8,26 @@ import (
...
+	envProxy := envproxy.NewEnvProxy()
+	envProxy.SetProxy("DCTL_S3_", "AWS_")
+	envProxy.Apply()
 	sess := session.Must(session.NewSessionWithOptions(session.Options{
 		SharedConfigState: session.SharedConfigEnable,
 	}))
+	envProxy.Restore()
 	return &S3Enumerator{
 		config,
 		s3.New(sess),
```
sundowndev, line 33: Isn't it possible to add a test case to ensure the session parses values from `DCTL_S3_` env vars ?

lotoussa: yes probably, i'll do it later today

pkg/iac/terraform/state/backend/s3_reader.go
```diff
@@ -21,6 +22,9 @@ type S3Backend struct {
 
 func NewS3Reader(path string) (*S3Backend, error) {
 
+	envProxy := envproxy.NewEnvProxy("DCTL_S3_", "AWS_")
+	envProxy.Apply()
+
```
wbeuil, line 27: I would have put it just before calling `session.Must`

wbeuil (changes requested):
This time it works with `DCTL_S3_DEFAULT_REGION=eu-west-3 AWS_DEFAULT_REGION=us-east-1 AWS_PROFILE=test driftctl scan -f "tfstate+s3://xxx/**/terraform.tfstate" --tf-provider-version 3.44.0`

pkg/iac/terraform/state/backend/s3_reader.go
lotoussa: done

wbeuil (approved):
Seems good to me. Need to check why the lint test failed.

sundowndev (approved):

wbeuil (approved):

moadibfr (approved):

Issue #679, opened 2021-06-22
ignore tags_all attribute


sjourdan:
**Description**

As discussed on https://github.com/cloudskiff/driftctl/issues/635

**Example**

sjourdan:
wen from (0.10.0-pre) 

```
Found changed resources:
    - vol-0e022e1567c9cec58 (aws_ebs_volume):
        + tags_all: <nil> => {
  Terraform: "true",
  Name: "aws-www-1-root"
} (computed)

```

to (0.10.0-pre build `797f527-dev`) 

```
# nothing
```

LGTM!!

Issue #635, opened 2021-06-14
`tags_all` attribute will produce drifts


wbeuil:
**Description**
Since terraform provider 3.38.0, we can enable provider-level tagging as an alternative to tagging each resources with the same tags. It produces drifts on driftctl analysis.

**How to reproduce**
```
provider "aws" {
  default_tags {
    tags = {
      Environment = "Test"
    }
  }
}

resource "aws_instance" "test" {
  ami           = "ami-0b2ca94b5b49e0132"
  instance_type = "t2.micro"
  root_block_device {
    tags = {
      "Name"      = "test"
    }
  }
}

resource "aws_s3_bucket" "test" {
  bucket_prefix = "test-bucket-"
}
```

```
Found changed resources:
    - i-08121f2f13a5cbe30 (aws_instance):
        + tags: <nil> => {
  Environment: "Test"
}
    - test-bucket-20210611121140651700000001 (aws_s3_bucket):
        + tags: <nil> => {
  Environment: "Test"
}
```

**Possible Solution**
We can create a middleware to add global tags to each resources' tags and then delete the `tags_all` attribute to avoid duplicate.

sjourdan:
I have this issue without enabling anything

```
Found changed resources:
    - vol-0e022e1567c9cec58 (aws_ebs_volume):
        + tags_all: <nil> => {
  Terraform: "true",
  Name: "aws-www-1-root"
} (computed)

```

My provider config: 

```
terraform {
  required_version = "= 1.0.0"
  required_providers {
    aws = "= 3.46.0"
  }
  backend "s3" {
  [...]
}

provider "aws" {
  region = var.aws_region
}

```

sjourdan:
report of what was debated on discord: 

- it will generate way too much noise to users if we keep those `tags_all` attributes
- future deep mode will lower the impact of this change
- step 1: ignore tags_all for now (_A mvp could be to add a middleware that remove theses fields in every resource (should be run after all middlewares)_ says @eliecharra) 
- step 2: make a better middleware (in sync with deep mode roadmap)

sjourdan:
won't fix: `tags_all` will stay ignored for now.

PR #680, opened 2021-06-22
Middleware to manage tags_all attribute

wbeuil:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | yes
| 🚀 New feature?   | no
| ⚠ Deprecations?   | no
| ❌ BC Break       | no
| 🔗 Related issues | #635 and #679
| ❓ Documentation  | no

## Description

On my first attempt, I tried to create a middleware to just `SafeDelete` each `tags_all` attribute on state and remote resources. But I had huge false positives on `tags` attribute because for remote resources `tags_all` == `tags`.

Thus I created a tags_all manager middleware in which, for the remote resources, I simply `SafeDelete` the `tags_all` attribute and for the state resources, since `tags_all` seems to be

```go
if err := d.Set("tags_all", tags.Map()); err != nil {
		return fmt.Errorf("error setting tags_all: %w", err)
	}
```

I `SafeSet` `tags_all` to `tags` and delete the former one.

driftctl needs to only find drifts on tags, we don't care if they were set as default_tags or not.

### DIFF
--- a/pkg/middlewares/tags_all_manager.go
+++ b/pkg/middlewares/tags_all_manager.go
@@ -1,0 +1,31 @@
+package middlewares
+
+import (
+	"github.com/cloudskiff/driftctl/pkg/resource"
+)
+
+// Manage tags_all attribute on each compatible resources
+type TagsAllManager struct{}
+
+func NewTagsAllManager() TagsAllManager {
+	return TagsAllManager{}
+}
+
+func (a TagsAllManager) Execute(remoteResources, resourcesFromState *[]resource.Resource) error {
+	for _, remoteRes := range *remoteResources {
+		if res, ok := remoteRes.(*resource.AbstractResource); ok {
+			if _, exist := res.Attrs.Get("tags_all"); exist {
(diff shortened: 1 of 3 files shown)


sundowndev (approved):

pkg/middlewares/tags_all_manager.go
```diff
@@ -0,0 +1,31 @@
...
+		if res, ok := remoteRes.(*resource.AbstractResource); ok {
+			if _, exist := res.Attrs.Get("tags_all"); exist {
+				res.Attrs.SafeDelete([]string{"tags_all"})
+			}
+		}
+	}
+	for _, stateRes := range *resourcesFromState {
+		if res, ok := stateRes.(*resource.AbstractResource); ok {
+			if allTags, exist := res.Attrs.Get("tags_all"); exist {
+				_ = res.Attrs.SafeSet([]string{"tags"}, allTags)
```
eliecharra, line 25: I think we should merge `all_tags` and `tags` there nope ?

eliecharra (approved):

Issue #562, opened 2021-05-28
false positive: aws_ebs_volume is found missing


sjourdan:
**Description**

driftctl 0.9.0 reports a missing `aws_ebs_volume`: 

```shell 
Scanned states (1)      
Scanned resources (51)      
Found missing resources:
  aws_ebs_volume:
    - vol-000e2dffd7f09fdc0
[...]
``` 

The said volume is available tho: 

```
$ aws ec2 describe-volumes --volume-ids vol-000e2dffd7f09fdc0

{
    "Volumes": [
        {
            "Attachments": [
                {
                    "AttachTime": "2021-05-05T06:26:01+00:00",
                    "Device": "/dev/xvdd",
                    "InstanceId": "i-0a85fc94dad0eee9a",
                    "State": "attached",
                    "VolumeId": "vol-000e2dffd7f09fdc0",
                    "DeleteOnTermination": false
                }
            ],
            "AvailabilityZone": "eu-west-3a",
            "CreateTime": "2021-05-05T06:25:39.455000+00:00",
            "Encrypted": true,
            "KmsKeyId": "arn:aws:kms:eu-west-3:943295019655:key/a53ba348-5a9a-4f86-8cd3-86698e80ac39",
            "Size": 125,
            "SnapshotId": "",
            "State": "in-use",
            "VolumeId": "vol-000e2dffd7f09fdc0",
            "Tags": [
                {
                    "Key": "Name",
                    "Value": "aws-www-1-data"
                }
            ],
            "VolumeType": "st1",
            "MultiAttachEnabled": false
        }
    ]
}
```

Table version for readability if needed: 

```
---------------------------------------------------------------------------------------------------------
|                                            DescribeVolumes                                            |
+-------------------------------------------------------------------------------------------------------+
||                                               Volumes                                               ||
|+--------------------+--------------------------------------------------------------------------------+|
||  AvailabilityZone  |  eu-west-3a                                                                    ||
||  CreateTime        |  2021-05-05T06:25:39.455000+00:00                                              ||
||  Encrypted         |  True                                                                          ||
||  KmsKeyId          |  arn:aws:kms:eu-west-3:943295019655:key/a53ba348-5a9a-4f86-8cd3-86698e80ac39   ||
||  MultiAttachEnabled|  False                                                                         ||
||  Size              |  125                                                                           ||
||  SnapshotId        |                                                                                ||
||  State             |  in-use                                                                        ||
[...]

PR #683, opened 2021-06-22
Fix duplicate ebs volume in state

wbeuil:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | yes
| 🚀 New feature?   | no
| ⚠ Deprecations?   | no
| ❌ BC Break       | no
| 🔗 Related issues | #562
| ❓ Documentation  | no

## Description

We should not create a new ebs_volume if the one inlined in `instance.ebs_block_device` was created using `ebs_volume`.


eliecharra (approved):
Nice fix 👍🏻

PR #688, opened 2021-06-23
Fix golden files update

eliecharra:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | no
| 🚀 New feature?   | no
| ⚠ Deprecations?   | no
| ❌ BC Break       | no
| 🔗 Related issues | none
| ❓ Documentation  | no

## Description

Since we moved defaults version of provider upper, test fails when updating golden files :

image

image

### DIFF
--- a/pkg/remote/aws/init_test.go
+++ b/pkg/remote/aws/init_test.go
@@ -1,6 +1,8 @@
 package aws
 
 import (
+	"os"
+
 	"github.com/cloudskiff/driftctl/pkg/output"
 	"github.com/cloudskiff/driftctl/pkg/terraform"
 )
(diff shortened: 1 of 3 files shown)


moadibfr (dismissed):

pkg/remote/aws/init_test.go
```diff
@@ -1,14 +1,16 @@
...
+
 	"github.com/cloudskiff/driftctl/pkg/output"
 	"github.com/cloudskiff/driftctl/pkg/terraform"
 )
 
 func InitTestAwsProvider(providerLibrary *terraform.ProviderLibrary) (*AWSTerraformProvider, error) {
 	progress := &output.MockProgress{}
 	progress.On("Inc").Maybe().Return()
-	provider, err := NewAWSTerraformProvider("", progress, "")
+	provider, err := NewAWSTerraformProvider("3.19.0", progress, os.TempDir())
```
sundowndev, line 13: May be we should put this in a constant ?

sundowndev (approved):
one question but LGTM

pkg/remote/aws/init_test.go
eliecharra: I don't think that is has the same sense that in the part of the code where we assign it as a default value
This value should be enforced by the test and does not have to follow the "latest default" from the code or it will break some test. This need to be reproductible

Issue #443, opened 2021-04-16
Add number of driftignore lines to telemetry


eliecharra:
**Description**
Following #441, we want to add the number of rules (ignore empty lines, invalid lines and comments) in the driftignore to telemetry.

sjourdan:
back to product investigation, closing.

PR #705, opened 2021-06-25
Implement in-memory store for analysis data

sundowndev:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | no
| 🚀 New feature?   | no
| ⚠ Deprecations?   | no
| ❌ BC Break       | no
| 🔗 Related issues | #443
| ❓ Documentation  | no 

## Description

Following a discussion with @eliecharra , we though it'd be nice to have a in-memory key-value store for storing data related to the analysis and use them to send telemetry data. The implementation of this is a simple store in which you can create buckets and store values in it. The bucket can then be marshaled to JSON (for example to be passed to an http client). This way, we can reuse that logic for other parts of the execution and various purposes.

If you have comments about the implementation details, let's discuss about it!


pkg/driftctl.go
```diff
@@ -45,6 +48,7 @@ type DriftCTL struct {
 	iacProgress              globaloutput.Progress
 	resourceSchemaRepository resource.SchemaRepositoryInterface
 	opts                     *ScanOptions
+	s                        memstore.Store
```
eliecharra, line 51: I'd rather like a more complete naming like `store` or `memoryStore`

pkg/cmd/scan.go
```diff
@@ -237,7 +239,7 @@ func scanRun(opts *pkg.ScanOptions) error {
 	globaloutput.Printf(color.WhiteString("Scan duration: %s\n", analysis.Duration.Round(time.Second)))
 
 	if !opts.DisableTelemetry {
-		telemetry.SendTelemetry(analysis)
+		telemetry.SendTelemetry(store)
```
eliecharra, line 242: I think this method should not have any argument, memory store should be passed in telemetry constructor, here we should only have to do `telemetry.SendTelemetry()` imho

pkg/driftctl.go
```diff
@@ -136,6 +142,13 @@ func (d DriftCTL) Run() (*analyser.Analysis, error) {
 	analysis.Duration = time.Since(start)
 	analysis.Date = time.Now()
 
+	d.s.Bucket(memstore.TelemetryBucket).Set("version", version.Current())
```
eliecharra, line 145: We can set theses values earlier, imagine one day if we want to send telemetry even if a scan fail, we should be able to send all telemetry data not related to scan result in any way. Btw, we do not need to pass the memstore to driftctl IMHO, we can set telemetry keys related to analysis results in the called code (`scan.go`).

pkg/memstore/buckets.go
```diff
@@ -0,0 +1,6 @@
+package memstore
+
+const (
+	// TelemetryBucket is the name of the store bucket used by the telemetry service
+	TelemetryBucket = "telemetry"
```
eliecharra, line 5: Maybe we can have a custom type to force bucket to something predetermined.

```go
type BucketType int

const (
	TelemetryBucket BucketType = iota
	OtherBucket
)

// Use BucketType here
func (s store) Bucket(name BucketType) Bucket {
```

pkg/memstore/memstore.go
```diff
@@ -0,0 +1,63 @@
+package memstore
+
+import (
+	"encoding/json"
+	"sync"
+)
+
+type Bucket interface {
```
eliecharra, line 8: Should be in a `bucket.go`

pkg/memstore/memstore.go
```diff
@@ -0,0 +1,63 @@
...
+	Set(string, interface{})
+	Get(string) interface{}
+	MarshallJSON() ([]byte, error)
+}
+
+type Store interface {
```
eliecharra, line 14: `memstore/memstore.go` sound a bit redundant, maybe we can rename it `memstore/store.go` ?

pkg/telemetry/telemetry.go
```diff
@@ -2,40 +2,14 @@ package telemetry
...
-		Os:             runtime.GOOS,
-		Arch:           runtime.GOARCH,
-		TotalResources: analysis.Summary().TotalResources,
-		TotalManaged:   analysis.Summary().TotalManaged,
-		Duration:       uint(analysis.Duration.Seconds() + 0.5),
-	}
-
-	body, err := json.Marshal(t)
+func SendTelemetry(s memstore.Store) {
+	body, err := s.Bucket(memstore.TelemetryBucket).MarshallJSON()
```
eliecharra, line 12: We should not call `MarshallJSON` directly, use `json.Masrshall(bucket)` instead.

Btw, I'm wondering if we should pass directly bucket content, I think this is not the responsibility of the bucket to be serialized directly. I think we should still build a `Telemetry` object with explicit types and then serialize it to control serialization type. I see a weakness here if by mistake you put the wrong type in a `bucket.Set`

pkg/memstore/memstore.go
```diff
@@ -0,0 +1,63 @@
...
+}
+
+func (b bucket) Set(key string, value interface{}) {
+	b.m.Lock()
+	defer b.m.Unlock()
+	b.values[key] = value
+}
+
+func (b bucket) Get(key string) interface{} {
+	b.m.RLock()
```
eliecharra, line 56: I don't think we need a read lock

eliecharra (changes requested):

pkg/memstore/memstore.go
sundowndev: There's a race condition here that'll panic if we don't lock, I think that's the proper usage of read/write mutex, isn't it ?

pkg/cmd/scan.go
sundowndev: Not sure to understand, how should we pass memory store to telemetry if we don't pass it to telemetry.SendTelemetry function ?

eliecharra: My bad, in my mind in `telemetry.SendTelemetry`, `telemetry` was a struct initialized earlier but it is only the pkg name. forget this remark.

pkg/memstore/memstore.go
eliecharra: Ok I wasn't aware of possible concurrency even from just reading from a map. In my mind map operation are atomic and thus you do not have to lock for a read. Good to know

sundowndev: I wasn't sure too so I launched tests with `-race` and it caused panic

pkg/cmd/scan.go
```diff
@@ -225,6 +227,9 @@ func scanRun(opts *pkg.ScanOptions) error {
 	}()
 
 	analysis, err := ctl.Run()
+	if !opts.DisableTelemetry {
+		telemetry.SendTelemetry(store)
+	}
```
sundowndev, line 232: I should revert this since we don't send any error related field in telemetry data

pkg/driftctl_test.go
```diff
@@ -591,9 +671,16 @@ func TestDriftctlRun_BasicFilter(t *testing.T) {
 					Attrs: &resource.Attributes{},
 				},
 			},
-			assert: func(result *test.ScanResult, err error) {
+			assert: func(t *testing.T, result *test.ScanResult, err error, store memstore.Store) {
 				result.AssertUnmanagedCount(1)
 				result.AssertResourceUnmanaged("res2", "filtered")
+
+				assert.Equal(t, "dev-dev", store.Bucket(memstore.TelemetryBucket).Get("version"))
```
eliecharra, line 678: I'm not sure that we need to test telemetry data in this test case. It make sense to test store in the basic test `TestDriftctlRun_BasicBehavior` but here we want to test filter and IMHO adding memstore assertions here will lead us to more maintenance for the same scope covered. Same remark for all other test cases. BTW adding a `assertStore` attribute to testcase may be less painfull since you do not have to update every assert func in unrelated tests

pkg/telemetry/telemetry.go
```diff
@@ -20,19 +18,31 @@ type telemetry struct {
 	Duration       uint   `json:"duration"`
 }
 
-func SendTelemetry(analysis *analyser.Analysis) {
+func SendTelemetry(s memstore.Store) {
```
eliecharra, line 21: Maybe we can pass directly the telemetry bucket here WDYT ?

eliecharra (changes requested):

eliecharra (approved):

Issue #571, opened 2021-06-01
False positive aws_security_group_rule


wbeuil:
**Description**
Found missing `aws_security_group_rule` resource in an `aws_default_security_group` resource.

```
# output json
{
  "id": "sgrule-3983462422",
  "type": "aws_security_group_rule"
}

#output console
Scanned states (1)      
Scanned resources (62)      
Found missing resources:
  aws_security_group_rule:
    - Type: egress, SecurityGroup: sg-0054329bbba384f1a, Protocol: All, Ports: All, Destination: 0.0.0.0/0
```

**Environment**
 - OS: macOS
 - driftctl version: 0.9.0

**How to reproduce**

```
resource "aws_default_security_group" "sg" {
  vpc_id = aws_vpc.vpc.id
  tags   = local.tags
}

resource "aws_security_group_rule" "i_ssh" {
  type              = "ingress"
  from_port         = 22
  to_port           = 22
  protocol          = "tcp"
  cidr_blocks       = ["0.0.0.0/0"]
  description       = "SSH"
  security_group_id = aws_default_security_group.sg.id
}

resource "aws_security_group_rule" "i_http" {
  type              = "ingress"
  from_port         = 80
  to_port           = 80
  protocol          = "tcp"
  cidr_blocks       = ["0.0.0.0/0"]
  description       = "HTTP"
  security_group_id = aws_default_security_group.sg.id
}

resource "aws_security_group_rule" "i_https" {
  type              = "ingress"
  from_port         = 443
  to_port           = 443
  protocol          = "tcp"
  cidr_blocks       = ["0.0.0.0/0"]
  description       = "HTTPS"
  security_group_id = aws_default_security_group.sg.id
}

resource "aws_security_group_rule" "e_all" {
  type              = "egress"
  from_port         = 0
  to_port           = 0
  protocol          = "-1"
  cidr_blocks       = ["0.0.0.0/0"]
  description       = "All protocols"
  security_group_id = aws_default_security_group.sg.id
}
```

wbeuil:
In the listing of the security group rules, we remove from the list 2 types of rule:

```
func (s *VPCSecurityGroupRuleSupplier) isDefaultIngress(rule *securityGroupRule) bool {
	return rule.Type == sgRuleTypeIngress &&
		rule.FromPort == 0 &&
		rule.ToPort == 0 &&
		rule.Protocol == "-1" &&
		rule.CidrBlocks == nil &&
[...]

PR #707, opened 2021-06-28
Fix default security group rules from default security group

wbeuil:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | yes
| 🚀 New feature?   | no
| ⚠ Deprecations?   | no
| ❌ BC Break       | no
| 🔗 Related issues | #571
| ❓ Documentation  | no

## Description

Default ingress and egress rules should be ignored in a middleware to avoid false positives in the case of managed ones.


lotoussa (approved):

sundowndev (approved):

eliecharra (approved):
Nice catch and nice fix

PR #761, opened 2021-06-29
Split s3_bucket_metrics_supplier

### LINKED ISSUE #743: Split s3_bucket_metrics_supplier

eliecharra:
Fix #743


pkg/remote/aws/s3_bucket_metrics_detail_fetcher.go
```diff
@@ -0,0 +1,38 @@
...
+func NewS3BucketMetricsDetailsFetcher(provider terraform.ResourceReader, deserializer *resource.Deserializer) *S3BucketMetricsDetailsFetcher {
+	return &S3BucketMetricsDetailsFetcher{
+		reader:       provider,
+		deserializer: deserializer,
+	}
+}
+
+func (r *S3BucketMetricsDetailsFetcher) ReadDetails(res resource.Resource) (resource.Resource, error) {
+	ctyVal, err := r.reader.ReadResource(terraform.ReadResourceArgs{
+		Ty: aws.AwsS3BucketMetricResourceType,
```
lotoussa, line 23: I'm not sure of the relevance of this remark but why don't we have a variable like in suppliers to call the type of the resource? (supportedType())

eliecharra: This variable will be used in enumerators (not in suppliers) to filter resources we do not want without the need to scan them. If the enumerator do not run, then the detail fetcher will not run too as it uses enumerators output. So this is not relevant here

moadibfr (approved):

PR #768, opened 2021-06-30
Add generic detail fetcher

eliecharra:


sundowndev (approved):
LGTM

eliecharra:
> Will it be used by default when resource hasn't a dedicated detail fetcher ?

Good point, I was wondering if we should add this, actually with the current implementation we should always be explicit with

```go
remoteLibrary.AddDetailsFetcher(aws.AwsS3BucketResourceType, NewGenericDetailsFetcher(aws.AwsS3BucketResourceType, provider, deserializer))
```

But it can make sense to use it by default WDYT @cloudskiff/engineering ?

PR #781, opened 2021-07-01
Split internet_gateway_supplier

### LINKED ISSUE #730: Split internet_gateway_supplier

wbeuil:
Fix #730


pkg/remote/aws/ec2_internet_gateway_enumerator.go
```diff
@@ -0,0 +1,49 @@
...
+	remoteerror "github.com/cloudskiff/driftctl/pkg/remote/error"
+	tf "github.com/cloudskiff/driftctl/pkg/remote/terraform"
+	"github.com/cloudskiff/driftctl/pkg/resource"
+	"github.com/cloudskiff/driftctl/pkg/resource/aws"
+)
+
+type EC2InternetGatewayEnumerator struct {
+	repository     repository.EC2Repository
+	factory        resource.ResourceFactory
+	providerConfig tf.TerraformProviderConfig
```
eliecharra, line 14: We do not need `providerConfig` here, it was used in s3 enumerators to get the current region but we don't care here. You can remove it

eliecharra (changes requested):

pkg/remote/aws/ec2_internet_gateway_enumerator.go
wbeuil: Nice catch! I will fix the other EC2 Enumerators then.

eliecharra (approved):

PR #792, opened 2021-07-05
Split route_table_supplier

### LINKED ISSUE #741: Split route_table_supplier

wbeuil:
Fix #741


pkg/remote/aws/ec2_route_table_enumerator.go
```diff
@@ -0,0 +1,55 @@
...
+func (e *EC2RouteTableEnumerator) Enumerate() ([]resource.Resource, error) {
+	routeTables, err := e.repository.ListAllRouteTables()
+	if err != nil {
+		return nil, remoteerror.NewResourceEnumerationError(err, string(e.SupportedType()))
+	}
+
+	var results []resource.Resource
+
+	for _, routeTable := range routeTables {
+		var isMain bool
```
eliecharra, line 35: Maybe we can mutualize this logic in a method shared by the two route enumerators

eliecharra (changes requested):

pkg/remote/aws/ec2_route_table_enumerator.go
wbeuil: We can add a `isMainRouteTable` method in the `ec2_repository` file, wdyt ?

eliecharra: I don't think that it is the repository responsibility to handle that

eliecharra (approved):

Issue #797, opened 2021-07-06
driftctl stops with "error in supplier aws_s3_bucket: AccessDenied" with 0.11


sjourdan:
**Description**

Using 0.11+: 

```
❯  dctlenv use 0.11.0
Switching version to v0.11.0
Switching completed

~/.../20210701-v0.12/615 
❯  LOG_LEVEL=error AWS_PROFILE="615-iam-only" driftctl scan 
Scanned states (1) 
error in supplier aws_s3_bucket: AccessDenied: Access Denied
	status code: 403, request id: PSAC44Y292PS3KT5, host id: 3KbW3ZWmvhYStqtWBnbuWWm2TjMRUDbby8i64mbPI3M9jRrfEmVARVBZlzaUfh81TAfOx3hAR8U=

```

Using 0.10.0: 

```
❯  dctlenv use 0.10.0
Switching version to v0.10.0
Switching completed

~/.../20210701-v0.12/615 
❯  LOG_LEVEL=error AWS_PROFILE="615-iam-only" driftctl scan 
Scanned states (1) 
Found resources not covered by IaC:
[...a bunch of IAM only resources...]
Found 36 resource(s)
 - 2% coverage
 - 1 covered by IaC
 - 35 not covered by IaC
 - 0 missing on cloud provider
 - 0/1 changed outside of IaC
Ignoring aws_route53_health_check from drift calculation: Listing aws_route53_health_check is forbidden.
Ignoring aws_route_table_association from drift calculation: Listing aws_route_table is forbidden.
Ignoring aws_internet_gateway from drift calculation: Listing aws_internet_gateway is forbidden.
Ignoring aws_instance from drift calculation: Listing aws_instance is forbidden.
Ignoring aws_ami from drift calculation: Listing aws_ami is forbidden.
Ignoring aws_s3_bucket_notification from drift calculation: Listing aws_s3_bucket is forbidden.
Ignoring aws_s3_bucket_policy from drift calculation: Listing aws_s3_bucket is forbidden.
Ignoring aws_eip_association from drift calculation: Listing aws_eip_association is forbidden.
Ignoring aws_route53_record from drift calculation: Listing aws_route53_zone is forbidden.
Ignoring aws_subnet from drift calculation: Listing aws_subnet is forbidden.
Ignoring aws_db_subnet_group from drift calculation: Listing aws_db_subnet_group is forbidden.
Ignoring aws_sns_topic from drift calculation: Listing aws_sns_topic is forbidden.
[...]

PR #798, opened 2021-07-06
Fix enumeration errors not handled

eliecharra:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | yes
| 🚀 New feature?   | no
| ⚠ Deprecations?   | no
| ❌ BC Break       | no
| 🔗 Related issues | #797 
| ❓ Documentation  | no

## Description

Fix enumeration errors not handled since we refactored scanner


lotoussa (approved):

wbeuil (commented):
2 comments:
- We could reuse for each error the `*-empty` file to not duplicate the same results.
- To test for 403 error we could remove the mock alerter and try to assert on the right alert inside the map of alerts

eliecharra:
I tried to do a bugfix with minimal impact, this can be improved a lot for sure, but it is not the goal of a bugfix PR IMHO.

> * We could reuse for each error the `*-empty` file to not duplicate the same results.

I don't think we need a golden file, maybe we can handle this in another way for theses empty cases

> * To test for 403 error we could remove the mock alerter and try to assert on the right alert inside the map of alerts

Yeah I started to do this but we want a quick bugfix here and it will take much time to assert on a proper `SendAlert` and fix 55 cases. So for now I fixed it like this.

How do you want to handle that ? Maybe we can do two separate tech issues ?

wbeuil:
Understood ! Let's release like that and on the other way let's create 2 separate issues.

wbeuil (approved):

PR #801, opened 2021-07-06
Split vpc_security_group_rule_supplier

### LINKED ISSUE #752: Split vpc_security_group_rule_supplier

sundowndev:
fixes #752


pkg/remote/aws/vpc_security_group_rule_enumerator.go
```diff
@@ -66,82 +64,59 @@ func toInterfaceSlice(val []string) []interface{} {
...
-func (s *VPCSecurityGroupRuleSupplier) Resources() ([]resource.Resource, error) {
-	securityGroups, defaultSecurityGroups, err := s.repository.ListAllSecurityGroups()
+func (e *VPCSecurityGroupRuleEnumerator) SupportedType() resource.ResourceType {
+	return resourceaws.AwsSecurityGroupRuleResourceType
+}
+
+func (e *VPCSecurityGroupRuleEnumerator) Enumerate() ([]resource.Resource, error) {
+	securityGroups, defaultSecurityGroups, err := e.repository.ListAllSecurityGroups()
 	if err != nil {
 		return nil, remoteerror.NewResourceEnumerationError(err, resourceaws.AwsSecurityGroupRuleResourceType)
```
wbeuil, line 81: This is an error from the former supplier. Let's fix it now. It should be:

```
return nil, remoteerror.NewResourceEnumerationErrorWithType(err, string(e.SupportedType()), resourceaws.AwsSecurityGroupResourceType)
```

wbeuil (changes requested):
You need to create a specific `DetailsFetcher` and you need to remove each `schema.golden.json` file of the test folder.

lotoussa (approved):
lgtm

pkg/remote/aws/vpc_security_group_rule_details_fetcher.go
```diff
@@ -0,0 +1,70 @@
...
+		attrs["prefix_list_ids"] = v
+	}
+	if v, ok := res.Attributes().Get("source_security_group_id"); ok {
+		attrs["source_security_group_id"] = v
+	}
+
+	ctyVal, err := r.reader.ReadResource(terraform.ReadResourceArgs{
+		Ty:         aws.AwsSecurityGroupRuleResourceType,
+		ID:         res.TerraformId(),
+		Attributes: flatmap.Flatten(attrs),
```
eliecharra, line 59: We should avoid to use terraform pkg function since they are now private.  You can make a `map[string]string` and use `res.Attributes().GetString("cidr_blocks");` instead

eliecharra (changes requested):
Need rebase

pkg/remote/aws/vpc_security_group_rule_details_fetcher.go
sundowndev: Should we internalize this package then ?

sundowndev: `GetString` does not convert to string but simply cast to string, I'm afraid this will not work as expected

pkg/remote/aws/vpc_security_group_rule_details_fetcher.go
```diff
@@ -20,43 +21,43 @@ func NewVPCSecurityGroupRuleDetailsFetcher(provider terraform.ResourceReader, de
...
-		attrs["to_port"] = int(v.(float64))
+		attrs["to_port"] = fmt.Sprintf("%d", int(v.(float64)))
 	}
 	if v, ok := res.Attributes().Get("security_group_id"); ok {
-		attrs["security_group_id"] = v
+		attrs["security_group_id"] = fmt.Sprintf("%s", v)
 	}
 	if v, ok := res.Attributes().Get("self"); ok {
-		attrs["self"] = v
+		attrs["self"] = fmt.Sprintf("%v", v.(bool))
```
eliecharra, line 42: Should be `%t` no ?

eliecharra: https://github.com/22Acacia/terraform/blob/v0.6.4/flatmap/flatten.go#L31

sundowndev: Yeah I think it's better

from the docs;

`%t	the word true or false`

eliecharra (approved):

PR #817, opened 2021-07-07
Split sqs queue policy supplier

lotoussa:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | no
| 🚀 New feature?   | no
| ⚠ Deprecations?   | no
| ❌ BC Break       | no
| 🔗 Related issues | #749 
| ❓ Documentation  | no

branch created from 750 split sqs queue supplier.


pkg/remote/aws/sqs_queue_policy_enumerator.go
```diff
@@ -0,0 +1,48 @@
...
+import (
+	"github.com/cloudskiff/driftctl/pkg/remote/aws/repository"
+	remoteerror "github.com/cloudskiff/driftctl/pkg/remote/error"
+	"github.com/cloudskiff/driftctl/pkg/resource"
+	"github.com/cloudskiff/driftctl/pkg/resource/aws"
+
+	awssdk "github.com/aws/aws-sdk-go/aws"
+)
+
+type SqsQueuePolicyEnumerator struct {
```
wbeuil, line 12: Can you change everywhere to `SQS` and not `Sqs`. Thanks !

pkg/remote/sqs_scanner_test.go
```diff
@@ -112,3 +112,97 @@ func TestSqsQueue(t *testing.T) {
 		})
 	}
 }
+
+func TestSqsQueuePolicySupplier_Resources(t *testing.T) {
```
wbeuil, line 116: Just `TestSQSQueuePolicy`

wbeuil (changes requested):

pkg/remote/aws/sqs_queue_policy_enumerator.go
lotoussa: i have changed Sqs to SQS everywhere in the project on another PR. Let me know if it's too much and you prefer to change only on this PR, but for example on #750 it's still Sqs so i think it's not clean to change only on #749.

pkg/remote/sqs_scanner_test.go
lotoussa: done

wbeuil (approved):

PR #820, opened 2021-07-08
Split dynamodb_table_supplier

### LINKED ISSUE #714: Split dynamodb_table_supplier

sundowndev:
fixes #714


pkg/remote/aws/dynamodb_table_enumerator.go
```diff
@@ -0,0 +1,46 @@
...
+	}
+
+	results := make([]resource.Resource, len(tables))
+
+	for _, table := range tables {
+		results = append(
+			results,
+			e.factory.CreateAbstractResource(
+				string(e.SupportedType()),
+				*table,
```
lotoussa, line 39: shouldn't we use aws.StringValue() ?

lotoussa (dismissed):

pkg/remote/aws/dynamodb_table_enumerator.go
sundowndev: I honestly don't know, according to other enumerators I think we do. Howether that means we could create a resource with empty string as ID 🤔

wbeuil (approved):

pkg/remote/aws/dynamodb_table_enumerator.go
sundowndev: Yes, also if `table` is `nil`, it'll not crash but use `""` as resource id, which sounds weird

Issue #355, opened 2021-03-19
Test goreleaser instead of combination of gox & gh-release


sjourdan:
**Description**

@sundowndev convinced us to try goreleaser :)

the auto changelog feature can be interesting.

**Examples**

- https://github.com/goreleaser/example

eliecharra:
Sam just pushed a PR on dagger side https://github.com/dagger/dagger/pull/388

May be inspiring ?

eliecharra:
This is now integrated, we created #894 to handle docker builds in a more efficient way

PR #829, opened 2021-07-09
Setup Goreleaser

sundowndev:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | no
| 🚀 New feature?   | no
| ⚠ Deprecations?   | no
| ❌ BC Break       | no
| 🔗 Related issues | #355 
| ❓ Documentation  | no 

## Description

This PR setup goreleaser in our workflow in order to ease release management and support more platforms. Here's an overview of what the release will looks like : https://github.com/sundowndev/goreleaser-template/releases

**NOTE:** attentive review required for the circleci config, as I'm not familiar with that CI system.


.goreleaser.yml
```diff
@@ -0,0 +1,68 @@
...
+      build_flag_templates:
+          - --label=org.opencontainers.image.title={{ .ProjectName }}
+          - --label=org.opencontainers.image.description={{ .ProjectName }}
+          - --label=org.opencontainers.image.url=https://github.com/cloudskiff/driftctl
+          - --label=org.opencontainers.image.source=https://github.com/cloudskiff/driftctl
+          - --label=org.opencontainers.image.version={{ .Version }}
+          - --label=org.opencontainers.image.created={{ time "2006-01-02T15:04:05Z07:00" }}
+          - --label=org.opencontainers.image.revision={{ .FullCommit }}
+          - --label=org.opencontainers.image.licenses=MIT
+          - --platform=linux/amd64
```
sundowndev, line 58: Apache License 2.0

Dockerfile_release
```diff
@@ -0,0 +1,3 @@
+FROM scratch
+COPY driftctl /usr/bin/driftctl
+ENTRYPOINT ["/usr/bin/driftctl"]
```
eliecharra, line 3: What's the purpose of this dockerfile ?

.goreleaser.yml
```diff
@@ -0,0 +1,52 @@
...
+    - signature: "${artifact}.gpg"
+      artifacts: checksum
+      args: [ "--batch", "-u", "{{ .Env.SIGNINGKEY }}", "--output", "${signature}", "--detach-sign", "${artifact}" ]
+builds:
+    - id: "driftctl-build"
+      binary: driftctl
+      dir: .
+      env:
+          - CGO_ENABLED=0
+          - GO111MODULE=on
```
eliecharra, line 22: `GO111MODULE` is useless since it is the default

.goreleaser.yml
```diff
@@ -0,0 +1,52 @@
...
+before:
+    hooks:
+        - go mod tidy
+        - go generate ./...
+signs:
+    - signature: "${artifact}.gpg"
+      artifacts: checksum
+      args: [ "--batch", "-u", "{{ .Env.SIGNINGKEY }}", "--output", "${signature}", "--detach-sign", "${artifact}" ]
+builds:
+    - id: "driftctl-build"
```
eliecharra, line 17: `id: "driftctl"` sounds less redundant

.goreleaser.yml
```diff
@@ -0,0 +1,52 @@
...
+      goos:
+          - linux
+          - windows
+          - darwin
+      goarch:
+          - amd64
+          - arm
+          - arm64
+          - 386
+      ldflags: "-X github.com/cloudskiff/driftctl/pkg/version.version={{ .Tag }}"
```
eliecharra, line 32: We should have a single entrypoint for the build to be able to reproduce the same behavior in CI and locally. Previously we have a `build.sh` script which responsibility was to pass proper flags. We should find a way to be able to build things locally too. What's the behavior of goreleaser if we are not on a tag ? Is there a way to wrap goreleaser in the `build.sh` script to get the same logic as before ? BTW it misses the flag `-X github.com/cloudskiff/driftctl/build.env=release`. I don't think we can avoid a custom build script and thus it should be updated to to not use gox anymore. We should absolutely have a single way to build binaries to avoid the `work on my machine` effect. I stop the review here since it will probably implies a lot of changes

eliecharra (changes requested):

.goreleaser.yml
sundowndev: I agree but goreleaser can be used in local environments too. I could make a script that simply uses goreleaser to build locally. WDYT ?

> What's the behavior of goreleaser if we are not on a tag ?

I think the default value is the current commit hash (I'll try to confirm)

Dockerfile_release
sundowndev: It was useful for goreleaser to build and push the Docker image, but it's not useful anymore since we decided to not use goreleaser for that.

.goreleaser.yml
sundowndev: https://goreleaser.com/cmd/goreleaser_build/

eliecharra: > I agree but goreleaser can be used in local environments too. I could make a script that simply uses goreleaser to build locally. WDYT ?

Yeah i think it's fine if you adapt the existing script to make it work with goreleaser that's the idea. We need to have a single way to build a binary for : 
- Docker: release mode, version tag
- local:  dev mode, commit hash (if not checked out on a tag), not cross platforrm (build only the current platform, but with the availability to cross build another specific platform, for example to build a binary with a hotfix to be tested by people like we already did before on windows) 
- CI: release mode, version tag, GPG signature

sundowndev: I created a script for each use case : build and release, let me know what you think about it

- Build script builds the project for the current platform (intended for devs)
- Release script builds the project for all platforms, sign using GPG, publish on github if needed (intended for CI)

.gitignore
```diff
@@ -1,5 +1,4 @@
 /bin/*
-!/bin/.gitkeep
```
eliecharra, line 2: If we decide not to keep the bin folder, you can change the first line to `/bin`

Makefile
```diff
@@ -21,11 +21,11 @@ all: fmt lint test build
 
 .PHONY: build
 build:
-	scripts/build.sh
+	ENV=dev scripts/build.sh
```
eliecharra, line 24: Why do we need and `ENV` var if we call two distinct scripts ?

.goreleaser.yml
```diff
@@ -0,0 +1,52 @@
...
+dist: bin
+release:
+    github:
+        owner: cloudskiff
+        name: driftctl
+    draft: false
+    prerelease: auto
+before:
+    hooks:
+        - go mod tidy
```
eliecharra, line 11: I don't think that it is the build to handle this. The go mod tidy should be something done by the developpers.

scripts/release.sh
```diff
@@ -0,0 +1,47 @@
...
+
+# Check configuration
+goreleaser check
+
+GRFLAGS=""
+
+# We sign every releases using PGP
+# We may not want to do so in dev environments
+if [ -z $SIGNINGKEY ]; then
+    GRFLAGS="--skip-sign ${GRFLAGS}"
```
eliecharra, line 32: Can be replaced with `GRFLAGS="--skip-sign"`. You do not need to concat here since we have empty var initialization just before

.circleci/config.yml
```diff
@@ -83,46 +83,17 @@ jobs:
...
-  sign_release:
-    docker:
-      - image: cimg/base:2020.01
-    steps:
-      - attach_workspace:
-          at: ~/project
-      - run:
-          name: "Ensure GnuPG is available"
-          command: gpg --version
-      - run:
```
eliecharra, line 95: This step should be kept in the build process

.goreleaser.yml
```diff
@@ -0,0 +1,52 @@
...
+        - go generate ./...
+signs:
+    - signature: "${artifact}.gpg"
+      artifacts: checksum
+      args: [ "--batch", "-u", "{{ .Env.SIGNINGKEY }}", "--output", "${signature}", "--detach-sign", "${artifact}" ]
```
eliecharra, line 16: We expect the Key ID or email instead `{{ .Env.SIGNINGKEY }}`. This env contains the whole private key that should be decoded from base64 (see my comment on the circleci config file)

We should have this here `args: [ "--batch", "-u", "security@cloudskiff.com", "--output", "${signature}", "--detach-sign", "${artifact}" ]`

.goreleaser.yml
```diff
@@ -0,0 +1,52 @@
...
+    algorithm: sha256
+snapshot:
+    name_template: "{{ .Tag }}-next"
+nfpms:
+    - maintainer: CloudSkiff <engineering@cloudskiff.com>
+      description: Detect, track and alert on infrastructure drift
+      homepage: https://github.com/cloudskiff/driftctl
+      license: Apache License 2.0
+      file_name_template: '{{ .ProjectName }}_{{ .Os }}_{{ .Arch }}'
+      formats:
```
eliecharra, line 49: I don't think that we need deb, apk and rpm packages right now since we do not have a release process on theses platform. It's good to know that we can easily generate them but we'll do the work in dedicated PR in the future.

.circleci/config.yml
```diff
@@ -243,26 +214,16 @@ workflows:
...
-          context: cloudskiff-signing
-          requires:
-            - build
-          filters:
-            tags:
-              only: /^v.*/
-            branches:
-              ignore: /.*/
       - release:
           context: driftctl
```
eliecharra, line 218: We also need the `cloudskiff-signing` context here where remain the private key for signature

scripts/build.sh
```diff
@@ -2,85 +2,32 @@
...
+SELFPATH="$(cd -P "$(dirname "$SOURCE")/.." && pwd)"
 
-# In release mode we don't want debug information in the binary
-# We also set the build env to release
-if [ $ENV = "release" ]; then
-    echo "+ Building env: release"
-    LD_FLAGS="-s -w -X github.com/cloudskiff/driftctl/build.env=release ${LD_FLAGS}"
-fi
+# Change into that directory
+cd "$DIR"
```
eliecharra, line 17: I don't get why we have to change dir 
image

eliecharra (changes requested):
You should test the signing process locally with a dummy GPG key

Makefile
sundowndev: This value is used by goreleaser `ldflags: "-s -w -X github.com/cloudskiff/driftctl/build.env={{ .Env.ENV }}`

For some reasons, defining that value in the script does not work as expected so I added it there

.goreleaser.yml
sundowndev: copy/paste mistake, it was supposed to be the fingerprint

.goreleaser.yml
sundowndev: Ok should I comment this for now ?

scripts/build.sh
sundowndev: So goreleaser can find the config file relatively to where the script is located, not from where it's ran

Makefile
eliecharra: In the script you have 
```
# By default build for dev
ENV=${ENV:-"dev"}
```

I think you can replace it with 
```
ENV="dev"
# and
ENV="release"
```

as we use two distinct script now we do not need for a defaut like it was done before

scripts/build.sh
eliecharra: Is it really usefull ? I mean the script is meant to be run by the Makefile so the PWD is always the Makefile PWD. I think we can remove theses lines

.circleci/config.yml
eliecharra: This is still incorrect, you should remove the `sign_release` **job** as it is done by goreleaser now. But you have to move the **step** `Import cloudskiff signing key` to the release **job** 

```yaml
  release:
    steps:
      - run:
          name: "Import cloudskiff signing key"
          command: |
            echo ${SIGNINGKEY} | base64 -d > signingkey
            gpg --import signingkey
      - run:
          name: "Build and publish release"
          command: make release
```

The key is stored in env as base64 string and thus need to be decoded and imported in gpg

.circleci/config.yml
```diff
@@ -111,18 +111,15 @@ jobs:
             - bin/
   release:
     environment:
-        CIRCLE_PROJECT_USERNAME: cloudskiff
-        CIRCLE_PROJECT_REPONAME: driftctl
+      CI: circleci
```
eliecharra, line 114: You should use predefined vars instead of defining a custom one (that override defaults btw), scripts should be updated according.

https://circleci.com/docs/2.0/env-vars/#built-in-environment-variables

.circleci/config.yml
```diff
@@ -111,18 +111,15 @@ jobs:
...
     docker:
-      - image: cibuilds/github:0.13
+      - image: golang:1.16
     steps:
       - attach_workspace:
```
eliecharra, line 118: This is now useless, it was previously used to share build artifact between jobs, but since we build, sign and publish in a single job this now useless

eliecharra (changes requested):
`build.sh` and `release.sh` are very similar, any reason to have made two distinct scripts ? There is a lot of redundancy between them maybe we can only use one script with switches on `ENV` ? You can also create a `commom.sh` script that only define functions, then in build and release you can do `source common.sh` and call these functions.

.circleci/config.yml
sundowndev: Isn't it useful to publish on AUR ?

eliecharra: AUR only need the SHA256SUM file, now that is is generated here we need to add a persist step after the build.

```yaml
      - persist_to_workspace:
          root: ~/project
          paths:
            - bin/driftctl_SHA256SUMS
```

sundowndev:
> You should test the signing process locally with a dummy GPG key

Tested locally and working fine

```
$ gpg --verify ./bin/driftctl_SHA256SUMS.gpg ./bin/driftctl_SHA256SUMS
gpg: Signature made jeu. 22 juil. 2021 11:02:50 CEST
gpg:                using RSA key 2ED1BE3E01BFA2C3161DFA9096605C74C2630B3F
gpg:                issuer "test@gmail.com"
gpg: Good signature from "dummy <test@gmail.com>" [ultimate]
```

scripts/build.sh
```diff
@@ -2,85 +2,46 @@
...
-    VERSION="${GIT_COMMIT}${GIT_DIRTY}"
+if ! which goreleaser >/dev/null; then
+    echo "+ Installing goreleaser..."
+    go install github.com/goreleaser/goreleaser@v0.173.2
 fi
 
-if [ -n "$OS_ARCH" ]; then
-  os_archs=("$OS_ARCH")
+if [ -z $ENV ]; then
+    echo "Error: ENV variable must be defined"
```
eliecharra, line 14: 👍🏻

scripts/build.sh
```diff
@@ -2,85 +2,46 @@
...
-# Build!
-echo "+ Building with flags: ${LD_FLAGS}"
-osarch="${os_archs[@]}"
-gox \
-    -osarch="$osarch" \
-    -parallel=2 \
-    -ldflags "${LD_FLAGS}" \
-    -output "bin/driftctl_{{.OS}}_{{.Arch}}" \
-    ./
+echo ${GRFLAGS}
```
eliecharra, line 42: Maybe this could be removed ?

eliecharra (changes requested):
Nice build script rework, when you'll fix circleci workflow issue we can merge it

eliecharra (approved):

Issue #846, opened 2021-07-16
Minimal enrichment for `aws_default_security_group_rule` middleware 


eliecharra:
We need almost of the field for `aws_security_group_rule`, check the middleware for an exhaustive list

sundowndev:
Not sure I understand this one correctly, doesn't enumerator contain needed attributes already ? https://github.com/cloudskiff/driftctl/blob/main/pkg/remote/aws/vpc_security_group_rule_enumerator.go#L42

eliecharra:
I didn't checked if middlewares already got their attributes, you can check if the middleware is working correctly without the detail fetcher and if it's ok then we can close this issue.

PR #857, opened 2021-07-19
Fix security group rule details fetcher

sundowndev:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | yes
| 🚀 New feature?   | no
| ⚠ Deprecations?   | no
| ❌ BC Break       | no
| 🔗 Related issues | #846
| ❓ Documentation  | no 

## Description

Default security group rules were not ignored by default anymore because some fields were of the wrong type.

```
DEBU[0009] Ignoring default unmanaged security group rule id=sgrule-3038934708 type=aws_security_group_rule
DEBU[0009] Ignoring default unmanaged security group rule id=sgrule-3415289939 type=aws_security_group_rule
```


lotoussa (approved):
lgtm

pkg/remote/aws/vpc_security_group_rule_enumerator.go
```diff
@@ -25,8 +25,8 @@ type securityGroupRule struct {
 	Type                  string
 	SecurityGroupId       string
 	Protocol              string
-	FromPort              int
-	ToPort                int
+	FromPort              float64
```
eliecharra, line 28: Why a float ? A TCP|UDP port cannot be a floating point number

eliecharra (changes requested):

pkg/remote/aws/vpc_security_group_rule_enumerator.go
sundowndev: As @moadibfr explained to me, we have to cast it as float64 because ctyjson deserialize integers as float64 and using int would mean to have a diff between serialized types and manual added types

eliecharra: 🆗

eliecharra (approved):

Issue #819, opened 2021-07-08
Add cli option to enable deep mode


sjourdan:
**Description**

- [ ] Non-deep mode immediately becomes the default behavior
- [ ] Add a command-line switch to enable the experimental deep mode.
- [ ] Add documentation about the supported resources for this experimental deep mode (all currently supported resources)

**Example**

```shell
$ driftctl scan
[...quickly lists resources]
$ driftctl scan --deep 
[0.12 behavior]
```

sjourdan:
LGTM

PR #860, opened 2021-07-20
Add deep mode flag

eliecharra:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | no
| 🚀 New feature?   | yes
| ⚠ Deprecations?   | yes
| ❌ BC Break       | yes
| 🔗 Related issues | #819 
| ❓ Documentation  | https://github.com/cloudskiff/driftctl-docs/pull/85

## Description

Add deep mode flag and make it not the default behavior

### DIFF
--- a/pkg/cmd/scan.go
+++ b/pkg/cmd/scan.go
@@ -33,7 +33,7 @@ import (
 )
 
 func NewScanCmd() *cobra.Command {
-	opts := &pkg.ScanOptions{Deep: true}
+	opts := &pkg.ScanOptions{}
 	opts.BackendOptions = &backend.Options{}
 
 	cmd := &cobra.Command{
@@ -99,6 +99,8 @@ func NewScanCmd() *cobra.Command {
 		},
 	}
 
+	warn := color.New(color.FgYellow, color.Bold).SprintfFunc()
+
 	fl := cmd.Flags()
 	fl.Bool(
 		"quiet",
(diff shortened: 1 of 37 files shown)


lotoussa (dismissed):
lgtm

pkg/cmd/scan.go
```diff
@@ -160,6 +162,13 @@ func NewScanCmd() *cobra.Command {
 		false,
 		"Includes cloud provider service-linked roles (disabled by default)",
 	)
+	fl.BoolVar(&opts.Deep,
+		"deep",
+		false,
+		fmt.Sprintf("%s Enable deep mode, disabled by default\n", warn("EXPERIMENTAL:"))+
```
wbeuil, line 168: I find it repetitive to say `disabled by default`. If we can enable it, it implicitly means that it is disabled.

pkg/cmd/scan.go
```diff
@@ -160,6 +162,13 @@ func NewScanCmd() *cobra.Command {
...
+			"In deep mode we compare resources internal values to expected ones (like a terraform plan).\n"+
+			warn("This can lead to unexpected behavior (false positives, hidden drifts) when using a tfstate as IaC source and thus this is not recommended"),
```
wbeuil, line 170: Maybe we can redirect our users to the docs instead of trying to explain what it is in the usage.

pkg/cmd/scan.go
sundowndev: yeah also Cobra already adds `(default: false)` to the help message

pkg/cmd/scan.go
```diff
@@ -160,6 +162,12 @@ func NewScanCmd() *cobra.Command {
 		false,
 		"Includes cloud provider service-linked roles (disabled by default)",
 	)
+	fl.BoolVar(&opts.Deep,
+		"deep",
+		false,
+		fmt.Sprintf("%s Enable deep mode\n", warn("EXPERIMENTAL:"))+
+			"You should check the documentation for more details : https://docs.driftctl.com/deep-mode\n",
```
wbeuil, line 169: Remove the space before `:`

wbeuil (approved):

Issue #831, opened 2021-07-09
Multiple output files out of same scan operation


suchakraborty:
Would it be possible to have 2 outputs for the same scan? like driftctl scan...-o json://... -o html://...

This would be helpful for me, as we have multiple environments, multiple regions, and would like to have regions specific html reports as well as json to easily create consolidated reports.

Another way probably more complicated would be to take multiple regions (and aws profiles) to create the consolidated report too - say working in terragrunt style foldering? But I can see that is way more complicated than just streaing the output through multiple channels to create different output files.

sjourdan:
LGTM!

I could generate at once, with multiple drifts and unmanaged stuff and folders and outputs 

- a console output for humans
- an HTML report to be stored on S3 somewhere
- a json file for a script to process it and send it to Slack

sounds really good

```
❯  LOG_LEVEL=error CLOUDSDK_CORE_PROJECT=driftctl-sj-tests GOOGLE_APPLICATION_CREDENTIALS=keys/driftctl.json driftctl-0.15-pre scan --from tfstate://**/*.tfstate --to gcp+tf --tf-provider-version 3.82.0 --output console:// --output html://report.html --output json://report.json
```

PR #865, opened 2021-07-21
Allow multiple output flags for a single scan

sundowndev:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | no
| 🚀 New feature?   | yes
| ⚠ Deprecations?   | no
| ❌ BC Break       | no
| 🔗 Related issues | #831
| ❓ Documentation  | yes https://github.com/cloudskiff/driftctl-docs/pull/86 

## Description

```bash
# Example
$ driftctl scan -o console://stdout -o html://reports/001/result.html -o json://reports/001/result.json
```

### DIFF
--- a/pkg/cmd/scan.go
+++ b/pkg/cmd/scan.go
@@ -61,12 +61,16 @@ func NewScanCmd() *cobra.Command {
 				)
 			}
 
-			outputFlag, _ := cmd.Flags().GetString("output")
-			out, err := parseOutputFlag(outputFlag)
+			outputFlag, _ := cmd.Flags().GetStringSlice("output")
+			if len(outputFlag) == 0 {
+				outputFlag = append(outputFlag, output.Example(output.ConsoleOutputType))
+			}
+
+			out, err := parseOutputFlags(outputFlag)
 			if err != nil {
 				return err
 			}
-			opts.Output = *out
+			opts.Output = out
 
 			filterFlag, _ := cmd.Flags().GetStringArray("filter")
 
@@ -117,10 +121,10 @@ func NewScanCmd() *cobra.Command {
 			"  - Type =='aws_s3_bucket && Id != 'my_bucket' (excludes s3 bucket 'my_bucket')\n"+
 			"  - Attr.Tags.Terraform == 'true' (include only resources that have Tag Terraform equal to 'true')\n",
 	)
-	fl.StringP(
+	fl.StringSliceP(
 		"output",
 		"o",
-		output.Example(output.ConsoleOutputType),
+		[]string{},
 		"Output format, by default it will write to the console\n"+
 			"Accepted formats are: "+strings.Join(output.SupportedOutputsExample(), ",")+"\n",
 	)
@@ -190,7 +194,6 @@ func NewScanCmd() *cobra.Command {
 
 func scanRun(opts *pkg.ScanOptions) error {
 	store := memstore.New()
-	selectedOutput := output.GetOutput(opts.Output, opts.Quiet)
 
 	c := make(chan os.Signal)
 	signal.Notify(c, os.Interrupt, syscall.SIGTERM)
@@ -256,9 +259,12 @@ func scanRun(opts *pkg.ScanOptions) error {
 	analysis.ProviderVersion = resourceSchemaRepository.ProviderVersion.String()
 	analysis.ProviderName = resourceSchemaRepository.ProviderName
 
-	err = selectedOutput.Write(analysis)
-	if err != nil {
-		return err
+	for _, o := range opts.Output {
+		selectedOutput := output.GetOutput(o, opts.Quiet)
+		err = selectedOutput.Write(analysis)
+		if err != nil {
+			return err
+		}
 	}
 
 	globaloutput.Printf(color.WhiteString("Scan duration: %s\n", analysis.Duration.Round(time.Second)))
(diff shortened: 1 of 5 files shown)


lotoussa (commented):
lgtm

moadibfr (approved):

pkg/cmd/scan.go
```diff
@@ -60,12 +60,16 @@ func NewScanCmd() *cobra.Command {
 				)
 			}
 
-			outputFlag, _ := cmd.Flags().GetString("output")
-			out, err := parseOutputFlag(outputFlag)
+			outputFlag, _ := cmd.Flags().GetStringSlice("output")
+			if len(outputFlag) == 0 {
```
eliecharra, line 64: Can we remove this if we specify the default to `[]string{output.Example(output.ConsoleOutputType)}` instead of `[]string{}` in `fl.StringSliceP` ?

pkg/cmd/scan_test.go
```diff
@@ -163,103 +164,149 @@ func Test_parseFromFlag(t *testing.T) {
...
+					Key: "json",
+					Options: map[string]string{
+						"path": "result.json",
+					},
+				},
+			},
+			err: nil,
+		},
+		{
+			name: "test multiple output values with invalid value",
```
eliecharra, line 294: It could be nice to add test case with two valid outputs to ensure they are not overriden `out: []string{"json:///test.json", "json://test2.json"},`. This does not change actual code coverage, but since it is a possible usecase from user point of view it should be covered IMHO

pkg/cmd/scan.go
```diff
@@ -231,9 +234,12 @@ func scanRun(opts *pkg.ScanOptions) error {
...
 	}
 
-	err = selectedOutput.Write(analysis)
-	if err != nil {
-		return err
+	for _, o := range opts.Output {
+		selectedOutput := output.GetOutput(o, opts.Quiet)
+		err = selectedOutput.Write(analysis)
+		if err != nil {
+			return err
```
eliecharra, line 241: I am wondering if we should break on the first error or if we should continue and log an error message. WDYT @wbeuil @moadibfr ? Compared on how we'll handle `--from` flags after @moadibfr's work, that may be more consistent not to break on first error.

eliecharra (changes requested):
Tested locally and it works great. Simple and efficient approach this is 👍🏻 

I only got small remarks

pkg/cmd/scan.go
sundowndev: I used this approach because user may not want to send output to console at all. Also the current behavior is to not send output to console when a different output is used. Do you think we should change that ?

pkg/cmd/scan.go
wbeuil: I would recommend your last remark which is to not break on first error. We could display an alert.

pkg/cmd/scan.go
eliecharra: But if you specify a default slice with console output, and some additional `--output` flags from CLI, they are not overwriting the whole default slice ?

moadibfr: yeah you would need two slice or to empty the slice before appending

eliecharra: It seems that it work well with a default slice :

```diff
diff --git a/pkg/cmd/scan.go b/pkg/cmd/scan.go
index c73fc265..e5f5bfb3 100644
--- a/pkg/cmd/scan.go
+++ b/pkg/cmd/scan.go
@@ -62,9 +62,6 @@ func NewScanCmd() *cobra.Command {
 			}
 
 			outputFlag, _ := cmd.Flags().GetStringSlice("output")
-			if len(outputFlag) == 0 {
-				outputFlag = append(outputFlag, output.Example(output.ConsoleOutputType))
-			}
 
 			out, err := parseOutputFlags(outputFlag)
 			if err != nil {
@@ -124,7 +121,7 @@ func NewScanCmd() *cobra.Command {
 	fl.StringSliceP(
 		"output",
 		"o",
-		[]string{},
+		[]string{output.Example(output.ConsoleOutputType)},
 		"Output format, by default it will write to the console\n"+
 			"Accepted formats are: "+strings.Join(output.SupportedOutputsExample(), ",")+"\n",
 	)

```

Without any flag we got : 



With a json flag :



This is exactly what we want for me, we should not change the default behavior when no flag are set, we should keep the console flag. If you do not want any output for me it's up to you to `> /dev/null`

sundowndev: Yeah you're right, just applied your patch

pkg/cmd/scan.go
sundowndev: Sending an alert there is impossible so I added an error log. Also added a fallback to console output when all outputs failed to compute.

pkg/cmd/scan.go
```diff
@@ -256,9 +256,21 @@ func scanRun(opts *pkg.ScanOptions) error {
 	analysis.ProviderVersion = resourceSchemaRepository.ProviderVersion.String()
 	analysis.ProviderName = resourceSchemaRepository.ProviderName
 
-	err = selectedOutput.Write(analysis)
-	if err != nil {
-		return err
+	validOutput := false
+	for _, o := range opts.Output {
+		if err = output.GetOutput(o, opts.Quiet).Write(analysis); err != nil {
+			logrus.Errorf("Computing output %s://%s failed: %v", o.Key, o.Options["path"], err.Error())
```
eliecharra, line 262: We are not really computing something here, `Error writing to output '%s'` may be more appropriate here.

The responsibility to print `%s://%s` could be a func of `output.OutputConfig` struct.

```go
type OutputConfig struct {
    Key     string
    Options map[string]string
}

func (o OutputConfig) String() string {
    return fmt.Sprintf("%s://%s", o.Key, o.Options["path"])
}
```

eliecharra: BTW I'm wondering why do we have a `Options map[string]string` since we only use a `path` key, maybe we should refacto this ? (in another PR of course)

eliecharra (changes requested):

pkg/cmd/scan.go
sundowndev: I agree, a map is a lack of type safety.

eliecharra (dismissed):

eliecharra (approved):

PR #868, opened 2021-07-22
fix crash when sqs queue is just deleted and still appears in listing

moadibfr:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | yes
| 🚀 New feature?   | no
| ⚠ Deprecations?   | no
| ❌ BC Break       | no
| 🔗 Related issues | 
| ❓ Documentation  | no 

## Description
fix crash when sqs queue is just deleted and still appears in listing


pkg/remote/aws/sqs_queue_details_fetcher.go
```diff
@@ -0,0 +1,43 @@
...
+}
+
+func (r *SQSQueueDetailsFetcher) ReadDetails(res resource.Resource) (resource.Resource, error) {
+	ctyVal, err := r.reader.ReadResource(terraform.ReadResourceArgs{
+		ID: res.TerraformId(),
+		Ty: aws.AwsSqsQueueResourceType,
+	})
+	if err != nil {
+		if strings.Contains(err.Error(), "NonExistentQueue") {
+			logrus.Debugf("Ignoring %+v", err)
```
eliecharra, line 31: Maybe a more explicit message with additional context could be more usefull in debug traces. We did it previously for EC2 instances

```go
			"id": res.TerraformId(),
		}).Debug("Instance read returned nil (instance may be terminated), ignoring ...")
```

pkg/remote/aws/sqs_queue_details_fetcher.go
```diff
@@ -0,0 +1,43 @@
...
+		deserializer: deserializer,
+	}
+}
+
+func (r *SQSQueueDetailsFetcher) ReadDetails(res resource.Resource) (resource.Resource, error) {
+	ctyVal, err := r.reader.ReadResource(terraform.ReadResourceArgs{
+		ID: res.TerraformId(),
+		Ty: aws.AwsSqsQueueResourceType,
+	})
+	if err != nil {
```
eliecharra, line 29: We should add a test to cover this case

eliecharra (changes requested):

pkg/remote/aws/sqs_queue_details_fetcher.go
moadibfr: Yeah I'd like to but we don't have test for details fetcher with mock and I couldn't reproduce this one to have goldenfile...

eliecharra: 🆗

eliecharra (approved):

Issue #666, opened 2021-06-21
Handle access denied exceptions on resources


eliecharra:
**Description**
We want to handle access denied exceptions when reading resources details.

**Environment**
 - OS: any
 - driftctl version: any
 - terraform version: not related
 - terraform providers versions: not related

**How to reproduce**
https://github.com/cloudskiff/driftctl/issues/665#issue-925486413

**Possible Solution**
We need to do string matching on error strings coming from `ReadResource`
We can reproduce an error behavior by enabling listing but deny every detail fetcher call (remove `lambda:GetFunction` from policy)
We can retrieve the error string with : `err.Error()`
Error should be handled in `scanner.go` when we iterate on details fetcher, like we did above for enumeration errors

```go
resourceWithDetails, err := fetcher.ReadDetails(res)
				if err != nil {
                                       // HERE
					return nil, err
				}
```

We should test behavior for every kind of resource since AWS does not return the same error code.
If we match an access denied exception, we should send an alert and skip the resource.

eliecharra:
Blocked by deep mode refactorisation

sjourdan:
I have a similar use case. One of my top-secret S3 buckets has a strict access policy that will prevent driftctl to scan it (and return a very expected 403).

Here's the policy: 

```json
{
    "Version": "2012-10-17",
    "Id": "S3PolicyId1",
    "Statement": [
        {
            "Sid": "IPAllow",
            "Effect": "Deny",
            "Principal": "*",
            "Action": "s3:*",
            "Resource": [
                "arn:aws:s3:::yaddabucket0130",
                "arn:aws:s3:::yaddabucket0130/*"
            ],
            "Condition": {
                "NotIpAddress": {
                    "aws:SourceIp": "1.2.3.4/32"
                }
            }
        }
    ]
```

sjourdan:
LGTM 

Context: 1 S3 bucket in the account has an explicit ban for my IP ("_yaddabucket0130_").

**Using 0.13** 

```sh
❯  LOG_LEVEL=error driftctl scan 
Scanned states (1) 
AccessDenied: Access Denied
        status code: 403, request id: PC50JPFPQJSP0735, host id: /GGji2ootu3WqRQNX7hEQxdAmXJSP5B1uYo3Dpo8unwmCXjLQJUF1fLsnTACjg70PMuN+VK8LvU=
```

**Using 0.14-pre** 

```sh
❯  LOG_LEVEL=error driftctl-pre scan 
[...]

PR #883, opened 2021-07-26
Ignore resource prior to scan

eliecharra:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | no
| 🚀 New feature?   | yes
| ⚠ Deprecations?   | no
| ❌ BC Break       | no
| 🔗 Related issues | #456 
| ❓ Documentation  | false

## Description

Make driftctl does not enumerate ignored resources types from driftignore.


lotoussa (approved):
lgtm

sundowndev (approved):
tested locally, works fine, good job!

moadibfr:
quick question is it intented that it wont ignore resource filtered by id ? If I have no read right on one kms for example and ignore just this one it would still try to scan it.

When I implemented it prior to our refacto. I filtered in ReadResource that way you can filter by type and id.

eliecharra:
> quick question is it intented that it wont ignore resource filtered by id ? If I have no read right on one kms for example and ignore just this one it would still try to scan it.
> 
> When I implemented it prior to our refacto. I filtered in ReadResource that way you can filter by type and id.

Yeah since deep mode is not our priority anymore I decided to put this behavior aside.
We can discuss about this with the team, and if we absolutely want it, we should do this in another dedicated PR IMHO

sjourdan:
Let's take the KMS case with a user without read rights: how can this user explicitly exclude the key?

sjourdan:
TLDR from Zoom call: 

- it's a deep mode only issue, so it's not prioritized too high
- the access denied resources cases will be handled by #666 so there's no real blocker here

pkg/remote/scanner.go
```diff
@@ -57,6 +60,12 @@ loop:
 
 func (s *Scanner) scan() ([]resource.Resource, error) {
 	for _, enumerator := range s.remoteLibrary.Enumerators() {
+		if s.filter.IsTypeIgnored(enumerator.SupportedType()) {
```
wbeuil, line 63: Could `s.filter` be nil ?

eliecharra: No, we pass a driftignore struct every time, if the driftignore is empty or does not exist we still pass a driftignore struct in `scan.go` but it will have empty matcher. So this should work

wbeuil (approved):
Tested it and i worked like a charm <3

PR #887, opened 2021-07-26
    add retry for destroy on some acceptancy tests

moadibfr:


test/acceptance/testing.go
```diff
@@ -203,6 +210,21 @@ func (c *AccTestCase) terraformApply() error {
...
 func (c *AccTestCase) terraformDestroy() error {
+	if c.RetryDestroy.NumberOfTry <= 0 {
+		return c.doDestroy()
+	}
+
+	r := retrier.New(retrier.ConstantBackoff(c.RetryDestroy.NumberOfTry, c.RetryDestroy.RetryDelay), nil)
+
+	err := r.Run(func() error {
+		return c.doDestroy()
+	})
```
sundowndev, line 221: you can pass doDestroy func directly here `r.Run(c.doDestroy)`

pkg/resource/aws/aws_eip_association_test.go
```diff
@@ -12,6 +13,10 @@ func TestAcc_Aws_EipAssociation(t *testing.T) {
 		TerraformVersion: "0.14.9",
 		Paths:            []string{"./testdata/acc/aws_eip_association"},
 		Args:             []string{"scan", "--filter", "Type=='aws_eip' || Type=='aws_eip_association'", "--tf-provider-version", "3.44.0", "--deep"},
+		RetryDestroy: acceptance.RetryConfig{
+			NumberOfTry: 3,
```
sundowndev, line 17: Not a big deal but, shouldn't it be `NumberOfTries` or `NumberOfAttempts` instead ?

test/acceptance/testing.go
```diff
@@ -203,6 +210,19 @@ func (c *AccTestCase) terraformApply() error {
...
 func (c *AccTestCase) terraformDestroy() error {
+	if c.RetryDestroy.NumberOfTry <= 0 {
+		return c.doDestroy()
+	}
+
+	r := retrier.New(retrier.ConstantBackoff(c.RetryDestroy.NumberOfTry, c.RetryDestroy.RetryDelay), nil)
+
+	err := r.Run(c.doDestroy)
+
+	return err
```
sundowndev, line 221: `return r.Run(c.doDestroy)`

sundowndev (changes requested):

pkg/resource/aws/aws_eip_association_test.go
eliecharra: +1 for `Attempts`

test/acceptance/testing.go
```diff
@@ -203,6 +210,19 @@ func (c *AccTestCase) terraformApply() error {
 }
 
 func (c *AccTestCase) terraformDestroy() error {
+	if c.RetryDestroy.NumberOfTry <= 0 {
+		return c.doDestroy()
+	}
+
+	r := retrier.New(retrier.ConstantBackoff(c.RetryDestroy.NumberOfTry, c.RetryDestroy.RetryDelay), nil)
```
eliecharra, line 217: Maybe we can update our existing `acceptance.RetryFor` to be able to use `RetryConfig` instead of adding a new dependency just for this ? I think it is quite overkill to add a dependency for a such simple code.

test/acceptance/testing.go
```diff
@@ -203,6 +210,19 @@ func (c *AccTestCase) terraformApply() error {
 }
 
 func (c *AccTestCase) terraformDestroy() error {
+	if c.RetryDestroy.NumberOfTry <= 0 {
```
eliecharra, line 213: If we use `uint8` instead of `int` for `NumberOfTry` we can do a `== 0` check, it does not make sense to have a number of attempts < 0.

eliecharra (changes requested):

test/acceptance/testing.go
moadibfr: it is not a new dependency we already use retrier in readresource, if it where new it would only be used when compiled the test anyway.
I would have prefere retryfor to use the dependency we already have I think. Also the api is quite good and I don't really see the use to implement something new just for our tests ?

test/acceptance/testing.go
moadibfr: retrier.ConstantBackoff( is waiting fo a int so that will need a cast. I just test like that to prevent a crash when initializing retrier but I can change it

moadibfr: also goland does not forbid to put -1 in a uint so yeah it would be a utin8

sundowndev (dismissed):

wbeuil (dismissed):

test/acceptance/testing.go
eliecharra: Ok, I have seen update on `go.sum` file and in my mind it was a new dependency

wbeuil (approved):

PR #895, opened 2021-07-27
Allow to build single binary in release mode

### LINKED ISSUE #894: Allow single os/arch build in build script
**Description**
We want to build a single `os/arch` version with the build script. It will be used in docker context to build a `linux/amd64` version.
It can be acceptable to build the current os/arch only but it will be better to be able to specify an os/arch.
A usecase can be to build a binary for a user for debug purpose, if someone is on windows we should be able to quickly build a win binary.

**Example**
Previously we were passing `OS_ARCH=linux/amd64` env var to the build script to force a single version to be builded.

eliecharra:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | no
| 🚀 New feature?   | no
| ⚠ Deprecations?   | no
| ❌ BC Break       | no
| 🔗 Related issues | Fix #894 
| ❓ Documentation  | no

## Description

Does not allow to specify a single os/arch to be built. It is a gorelease limitation, actually to achieve this you must comment unwanted os and arches in goreleaser config file


scripts/build.sh
```diff
@@ -10,15 +10,15 @@ if ! which goreleaser >/dev/null; then
...
+if [ -z $MODE ]; then
+    echo "Error: MODE variable must be defined"
     exit 1
 fi
 
 # Check configuration
 goreleaser check
 
-if [ "$ENV" == "dev" ]; then
+if [ "$MODE" == "dev" ]; then
```
sundowndev, line 21: I suggest we name this var `SINGLE_TARGET` with boolean value so we can run the script like so in Makefile : 

- Local build: `SINGLE_TARGET=true ./scripts/build.sh` (here ENV is empty and should be default to `dev`)
- Docker build: `SINGLE_TARGET=true ENV=release ./scripts/build.sh`
- Release build: `SINGLE_TARGET=false ENV=release ./scripts/build.sh`

I think running `MODE=dev make release` in the Dockerfile is a bit confusing, building for a single target is not specific to dev environments.

WDYT ?

eliecharra: > I think running MODE=dev make release in the Dockerfile is a bit confusing, building for a single target is not specific to dev environments.

This is very relevant, I totally agree on this.
That mean that for the docker build we should do : `ENV=release SINGLE_TARGET=true make build` ?

sundowndev: >  That mean that for the docker build we should do : ENV=release SINGLE_TARGET=true make build ?

Yes, but `SINGLE_TARGET` can be set directly in the Makefile

eliecharra: Something like that ?

```dockerfile
# Dockerfile
RUN MODE=release ENV=release make build
```


```makefile
# Makefile
build:
    SINGLE_TARGET=true ./scripts/build.sh
release:
    MODE=release ENV=release ./scripts/build.sh
```


```bash
# build.sh
# If ENV is not set, then it should be set to `dev`
# If SINGLE_TARGET is not set, then it should be set to `false`
ENV="${ENV:-dev}"
SINGLE_TARGET="${SINGLE_TARGET:-false}"
```

sundowndev: What'd be the purpose of `MODE` here ? I think you can drop it

But yeah that's what I had in mind

eliecharra: Yeah sorry we can drop the MODE of course.

moadibfr (approved):

Dockerfile
```diff
@@ -5,9 +5,9 @@ ARG ARCH="amd64"
 
 WORKDIR /go/src/app
 COPY go.mod go.sum Makefile ./
-RUN make deps
+RUN go mod download
 COPY . .
-RUN make release
+RUN make build
```
sundowndev, line 10: missing ENV=release here

sundowndev (changes requested):
Tried locally but `ENV=release make build` does not build for a single target as expected and `make build` fails because ENV is not defined

PR #897, opened 2021-07-29
Add resource source for terraform state

### LINKED ISSUE #873: Add resource source when scanning IaC
**Description**
Be able to identify resources origin for resources coming from IaC. The final goal is to display resource source in the output.

**Solution**
We should create a new composed struct `StateResource` : 

```go
type ResourceSource struct {
	Module string
	Name   string
	State  string
}

type StateResource struct {
	AbstractResource
	Source ResourceSource
}
```

and when we read IaC code we should add `Source` to be able to track down the resource during the execution.

eliecharra:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | no
| 🚀 New feature?   | no
| ⚠ Deprecations?   | no
| ❌ BC Break       | no
| 🔗 Related issues | Fix #873 
| ❓ Documentation  | no

### DIFF
--- a/pkg/iac/terraform/state/terraform_state_reader.go
+++ b/pkg/iac/terraform/state/terraform_state_reader.go
@@ -25,6 +25,11 @@ import (
 
 const TerraformStateReaderSupplier = "tfstate"
 
+type decodedRes struct {
+	source resource.Source
+	val    cty.Value
+}
+
 type TerraformStateReader struct {
 	library        *terraform.ProviderLibrary
 	config         config.SupplierConfig
@@ -50,7 +55,7 @@ func NewReader(config config.SupplierConfig, library *terraform.ProviderLibrary,
 	return &reader, nil
 }
 
-func (r *TerraformStateReader) retrieve() (map[string][]cty.Value, error) {
+func (r *TerraformStateReader) retrieve() (map[string][]decodedRes, error) {
 	b, err := backend.GetBackend(r.config, r.backendOptions)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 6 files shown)


pkg/resource/resource.go
```diff
@@ -17,11 +17,40 @@ type Resource interface {
...
 type AbstractResource struct {
-	Id    string
-	Type  string
-	Attrs *Attributes
-	Sch   *Schema `json:"-" diff:"-"`
+	Id     string
+	Type   string
+	Attrs  *Attributes
+	Sch    *Schema `json:"-" diff:"-"`
+	Source Source  `json:"-"`
```
moadibfr, line 53: don't we need a diff:"-" ?

eliecharra: We do not use compare the whole resource anymore, we should remove the diff annotation on `Sch` since this is useless. We only compare `res.Attributes()` in the analyzer now

pkg/iac/terraform/state/terraform_state_reader.go
```diff
@@ -165,16 +172,20 @@ func (r *TerraformStateReader) convertInstance(instance *states.ResourceInstance
...
-	for ty, val := range values {
-		decodedResources, err := r.deserializer.Deserialize(ty, val)
-		if err != nil {
-			logrus.WithField("ty", ty).Warnf("Could not read from state: %+v", err)
-			continue
+	for ty, val := range valFromState {
+		for _, stateVal := range val {
+			res, err := r.deserializer.DeserializeOne(ty, stateVal.val)
+			if err != nil {
+				logrus.WithField("ty", ty).Warnf("Could not read from state: %+v", err)
```
wbeuil, line 182: Maybe we can display here in which state the error came from

eliecharra: Yeah good idea, that was not the original goal of this PR but it can't hurt. Let's do this !

wbeuil (approved):
I'm good with that

pkg/resource/resource.go
moadibfr: :+1:

moadibfr (approved):

Issue #902, opened 2021-08-02
Add `google_storage_bucket`


eliecharra:
**Description**
Add support for `google_storage_bucket` in **non-deep** mode.

**Prerequisites :**
- [x] QA GCP account for ACC tests @eliecharra 

**Solution :**
- Add new provider GCP.
- During the creation of the new provider, take time to write a documentation on how to add a new provider.
- Add Enumerator for `google_storage_bucket`, with an acceptance test.
- Follow the contributor doc during the new resource implementation to check if it misses something.

sjourdan:
LGTM!

it looks good in default and deep mode, with different versions of the provider (v2 and v3 tested).
Comparing a v3 generated state with a v2 scanned infra doesn't work though but is not expected too (same situation than AWS)

HTML reports look good too: 

![image](https://user-images.githubusercontent.com/419121/132346926-2944a25e-3002-422f-8569-d7ead77895ea.png)

![image](https://user-images.githubusercontent.com/419121/132346965-3b96fe08-46f1-47c1-9ddb-0549d88b9155.png)

CLI

```
❯  LOG_LEVEL=error CLOUDSDK_CORE_PROJECT=driftctl-sj-tests GOOGLE_APPLICATION_CREDENTIALS=keys/driftctl.json driftctl-0.15-pre scan --to gcp+tf --tf-provider-version 3.82.0 
Scanned states (1) 
Found resources not covered by IaC:
  google_storage_bucket:
    - driftctl-qa-bucket-manual-1
    - manual_bucket_in_europe
Found 3 resource(s)
 - 33% coverage
 - 1 resource(s) managed by terraform
     - 0/1 resource(s) out of sync with Terraform state
 - 2 resource(s) not managed by Terraform
 - 0 resource(s) found in a Terraform state but missing on the cloud provider
Scan duration: 0s
Provider version used to scan: 3.82.0. Use --tf-provider-version to use another version.
```

Deep mode

```
❯  LOG_LEVEL=error CLOUDSDK_CORE_PROJECT=driftctl-sj-tests GOOGLE_APPLICATION_CREDENTIALS=keys/driftctl.json driftctl-0.15-pre scan --to gcp+tf --tf-provider-version 3.82.0 --deep
Scanned states (1) 
Found resources not covered by IaC:
  google_storage_bucket:
    - driftctl-qa-bucket-manual-1
    - manual_bucket_in_europe
Found changed resources:
  From tfstate://terraform.tfstate
[...]

PR #913, opened 2021-08-04
Add google_storage_bucket

eliecharra:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | no
| 🚀 New feature?   | yes
| ⚠ Deprecations?   | no
| ❌ BC Break       | no
| 🔗 Related issues | #902 
| ❓ Documentation  | https://github.com/cloudskiff/driftctl-docs/pull/92 

## Description

Add terraform google provider and `google_storage_bucket` in deep mode


pkg/remote/google/repository/asset.go
```diff
@@ -0,0 +1,81 @@
...
+	req := &assetpb.SearchAllResourcesRequest{
+		Scope: fmt.Sprintf("projects/%s", s.config.Project),
+		AssetTypes: []string{
+			storageBucketAssetType,
+		},
+	}
+	var results []*assetpb.ResourceSearchResult
+
+	s.lock.Lock()
+	defer s.lock.Unlock()
```
sundowndev, line 49: What's the purpose of this locker ?

eliecharra: It is to ensure that we do not call `SearchAllResources` multiples times, since this method can be called concurrently we want to avoid to make two call to AWS API if no results are cached initially.

pkg/resource/google/google_storage_bucket_test.go
```diff
@@ -0,0 +1,31 @@
...
+)
+
+func TestAcc_Google_StorageBucket(t *testing.T) {
+	acceptance.Run(t, acceptance.AccTestCase{
+		TerraformVersion: "0.15.0",
+		Paths:            []string{"./testdata/acc/google_storage_bucket"},
+		Args: []string{
+			"scan",
+			"--to", "gcp+tf",
+			"--filter", "Type=='google_storage_bucket'",
```
sundowndev, line 17: Shouldn't we pin provider version to `3.78.0` in ACC tests ?

sundowndev (changes requested):
Great work!

I just left few comments

pkg/resource/google/google_storage_bucket_test.go
eliecharra: We do not pin version in other ACC test so I followed the same behavior. 

The day we want to update the default provider version we can run the full ACC test suite on the new version if we do not pin version, and thus ensure backward compatibility since the version is pinned in the terraform files attached to theses tests. I think it can make sense to not pin them by default, an only pin them when we want to add an acceptance test  for a specific bug on a specific provider version. We may discuss this point with the team on discord after the summer break.

sundowndev: We did it for some acceptance tests but not all of them, not sure if we should do it for all of them then

sundowndev (approved):

wbeuil (dismissed):

pkg/remote/google/repository/asset_test.go
```diff
@@ -0,0 +1,62 @@
...
+			DisplayName: "driftctl-unittest-1",
+		},
+	}
+	assetClient, err := google.NewFakeAssetServer(expectedResults, nil)
+	if err != nil {
+		t.Fatal(err)
+	}
+	c := &cache.MockCache{}
+	c.On("Get", "SearchAllResources").Return(nil).Times(1)
+	c.On("Put", "SearchAllResources", mock.IsType([]*assetpb.ResourceSearchResult{})).Return(true).Times(1)
```
sundowndev, line 55: this is minor but it should be `.Return(false)`, because the key isn't supposed to exist yet

pkg/remote/google_storage_scanner_test.go
```diff
@@ -0,0 +1,153 @@
...
+			testFilter := &filter.MockFilter{}
+			testFilter.On("IsTypeIgnored", mock.Anything).Return(false)
+
+			s := NewScanner(remoteLibrary, alerter, scanOptions, testFilter)
+			got, err := s.Resources()
+			assert.Equal(tt, err, c.wantErr)
+			if err != nil {
+				return
+			}
+			alerter.AssertExpectations(tt)
```
sundowndev, line 149: missing `AssertExpectations` for testFilter as well

sundowndev (changes requested):
LGTM!

tests just misses a little assertion

sundowndev (approved):

Issue #921, opened 2021-08-09
JSON formatting is broken in HTML output


sundowndev:
When we have a diff on a JSON field, the result in the HTML report is ugly. The reason is we use the same formatting function than console output. We should create a custom formatting function.

### To reproduce

- Apply terraform file below
```hcl
provider "aws" {
    region = "us-east-1"
}

terraform {
    required_providers {
        aws = {
            version = "3.47.0"
        }
    }
}

resource "random_string" "prefix" {
    length  = 6
    upper   = false
    special = false
}

resource "aws_s3_bucket" "foobar" {
    bucket = "${random_string.prefix.result}.foobar.driftctl-test.com"
}

resource "aws_s3_bucket_policy" "foobar" {
    bucket = aws_s3_bucket.foobar.bucket
    policy = <<POLICY
{
  "Version":"2012-10-17",
  "Statement":[
    {
      "Sid":"PublicRead",
      "Effect":"Allow",
      "Principal": "*",
      "Action":["s3:GetObject","s3:GetObjectVersion"],
      "Resource":"arn:aws:s3:::${random_string.prefix.result}.foobar.driftctl-test.com/*"
    }
  ]
}
POLICY
}

```
- Go to AWS console and edit the bucket policy
```diff
- "Effect":"Allow",
+ "Effect":"Deny",
```
- run driftctl in deep mode with html report
- check changed resources in HTML report, you'll see ASCII colors switched in the json diff
- **We expect here a valid json diff with colors**
### Screenshots

![image](https://user-images.githubusercontent.com/6154987/128708623-f40ed60c-8140-4c7c-84e2-bd968ecfff17.png)

eliecharra:
Not that this should be covered by a test, and BTW this test change in test file can be removed :

```go
						{
							Change: diff.Change{
								Type: diff.UPDATE,
								Path: []string{"Policies", "0"},
								From: resource.Resource{},
								To:   resource.Resource{Id: "093cd6ba-cf6d-4800-b252-6a50ca8903cd", Type: "aws_iam_policy"},
							},
						},
```

sjourdan:
LGTM 

crystal clear now 

![image](https://user-images.githubusercontent.com/419121/131985447-e0c27ce4-6c8b-4e3a-8a05-14736e8b5c57.png)

PR #927, opened 2021-08-16
Fix JSON diff formatting in HTML output

sundowndev:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | yes
| 🚀 New feature?   | no
| ⚠ Deprecations?   | no
| ❌ BC Break       | no
| 🔗 Related issues | #921
| ❓ Documentation  | no 

## Description

When we have a diff on a JSON field, the result in the HTML report is ugly. The reason is we use the same formatting function than console output. We should create a custom formatting function.

This PR disables TTY coloring and add missing html tags to the JSON diff.

## Example

image

### DIFF
--- a/pkg/cmd/scan/output/console.go
+++ b/pkg/cmd/scan/output/console.go
@@ -98,7 +98,7 @@ func (c *Console) Write(analysis *analyser.Analysis) error {
 					if change.Type == diff.UPDATE {
 						if change.JsonString {
 							prefix := "           "
-							fmt.Printf("%s%s\n%s%s\n", whiteSpace, pref, prefix, jsonDiff(change.From, change.To, prefix))
+							fmt.Printf("%s%s\n%s%s\n", whiteSpace, pref, prefix, jsonDiff(change.From, change.To, isatty.IsTerminal(os.Stdout.Fd())))
 							continue
 						}
 					}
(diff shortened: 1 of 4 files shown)


pkg/cmd/scan/output/testdata/output.html
```diff
@@ -510,19 +512,35 @@ <h2>Jun 10, 2021</h2>
...
+<span class="code-box-line-delete">-      "Effect": "Allow",</span>
+<span class="code-box-line-create">+      "Effect": "Deny",</span>
+       "Principal": "*",
+<span class="code-box-line-delete">-      "Resource": "arn:aws:s3:::tmxxrn.foobar.driftctl-test.com/*",</span>
+<span class="code-box-line-create">+      "Resource": "arn:aws:s3:::tmxxrn.foobar.driftctl<span class="code-box-line-delete">-test.com/b/*",</span></span>
+<span class="code-box-line-delete">-      "Sid": "PublicRead"</span>
+<span class="code-box-line-create">+      "Sid": "PublicReadWrite"</span>
+     }
+   ],
+<span class="code-box-line-delete">-  "Version": "2012<span class="code-box-line-delete">-10-17"</span></span>
```
sundowndev, line 535: bad regex matching to fix

lotoussa (dismissed):

wbeuil (approved):
LGTM

pkg/cmd/scan/output/console.go
```diff
@@ -98,7 +98,7 @@ func (c *Console) Write(analysis *analyser.Analysis) error {
 					if change.Type == diff.UPDATE {
 						if change.JsonString {
 							prefix := "           "
-							fmt.Printf("%s%s\n%s%s\n", whiteSpace, pref, prefix, jsonDiff(change.From, change.To, prefix))
+							fmt.Printf("%s%s\n%s%s\n", whiteSpace, pref, prefix, jsonDiff(change.From, change.To, true))
```
eliecharra, line 101: Maybe we can pass `isatty.IsTerminal(os.Stdout.Fd())` instead of true here, and thus in `jsonDiff` we only need to pass the `coloring` parameter instead of doing an `&&` WDYT ?

pkg/cmd/scan/output/html_test.go
```diff
@@ -94,8 +94,8 @@ func TestHTML_Write(t *testing.T) {
 				)
 				a.AddDifference(analyser.Difference{
 					Res: &resource.Resource{
-						Id:   "diff-id-2",
-						Type: "aws_diff_resource",
+						Id:   "tmxxrn.foobar.driftctl-test.com",
```
eliecharra, line 97: This is minor but we could have kept the old `Id` and `Type` here

eliecharra (changes requested):
Looks fine for now, we may do a proper implementation in the future if we decide to go further with HTML report. I just have a small remark

eliecharra (approved):
👍🏻

PR #930, opened 2021-08-19
Embed images in HTML output

sundowndev:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | no
| 🚀 New feature?   | no
| ⚠ Deprecations?   | no
| ❌ BC Break       | no
| 🔗 Related issues | #...
| ❓ Documentation  | no 

## Description

The HTML output produce a simple HTML file that only needs two remote resources : driftctl logo and favicon.

This PR embed those resources in the binary so the HTML file does not rely on any external resource anymore. I also did small changes to the template.

I put provider version next to its name so it's less confusing for the user as they could think this is the driftctl version.

before:
image

after:
image


pkg/cmd/scan/output/assets/style.css
```diff
@@ -211,6 +211,7 @@ input[type="search"], select {
 
 .div-right {
     margin: 12px 0;
+    text-align: right;
```
wbeuil, line 214: Can you change that to `center` and add inside the media query `@media (min-width: 768px)` this line between classes `card` and `panels`:

```css
.div-right {
    text-align: right;
}
```

wbeuil (changes requested):
I don't have personally the cloud provider in uppercase letters. Wonder how you did that in your tests?

sundowndev:
> I don't have personally the cloud provider in uppercase letters. Wonder how you did that in your tests?

It's set directly in my test cases

```go
analysis: func() *analyser.Analysis {
    a := &analyser.Analysis{}
    a.Date = time.Date(2021, 06, 10, 0, 0, 0, 0, &time.Location{})
    a.ProviderName = "AWS"
    a.ProviderVersion = "3.19.0"
    return a
},
```

wbeuil (approved):

Issue #876, opened 2021-07-23
Display resource source in HTML output


sundowndev:
**Description**

We want to display resource source in HTML output so we always know which IaC source the drift come from.

**Implementation details**

We can create a select input in which we'd have states. Exactly how the resource type select works. We must create a distinct list of IaC sources in order to use it in the HTML template. The select should filter on states.

We also want to remove the `Resource Type`

![image](https://user-images.githubusercontent.com/6154987/126764618-e252d7fe-d298-4c95-8ef7-3b4b965296c2.png)

In the old resource type column, change the format to : `MODULE.TYPE.NAME` like in console output.

**Example**

![image](https://user-images.githubusercontent.com/16480203/126764067-425f584d-ec17-4f29-a302-6416c8ebb90b.png)

Blocked by #873

sjourdan:
LGTM

for a missing resource: 

![image](https://user-images.githubusercontent.com/419121/131985250-e00e1921-0eb4-46eb-ad1d-eb2ebdb23cb0.png)

for a changed resource: 

![image](https://user-images.githubusercontent.com/419121/131985312-0436fcc7-96a9-4bbb-8601-835a7da63e63.png)

PR #931, opened 2021-08-19
Display resource source in HTML output

sundowndev:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | no
| 🚀 New feature?   | yes
| ⚠ Deprecations?   | no
| ❌ BC Break       | no
| 🔗 Related issues | #876
| ❓ Documentation  | no 

## Description

This PR adds IaC source to drifts in HTML output.


pkg/cmd/scan/output/html.go
```diff
@@ -81,6 +81,16 @@ func (c *HTML) Write(analysis *analyser.Analysis) error {
 
 			return distinctResourceTypes(resources)
 		},
+		"getIaCSources": func() []string {
```
eliecharra, line 84: This could be changed to 
```go
		"getIaCSources": func() []string {
			resources := make([]*resource.Resource, 0)
            resources = append(resources, analysis.Deleted()...)
            resources = append(resources, analysis.Managed()...)
			return distinctIaCSources(resources)
		},
```

We may want to filter on IaC source for resource that may not have diffs (thinking of non deep mode which is now the default)

pkg/cmd/scan/output/html_test.go
```diff
@@ -151,6 +150,24 @@ func TestHTML_Write(t *testing.T) {
 							},
 						},
 					}})
+				a.AddDifference(analyser.Difference{
```
eliecharra, line 153: Should be tested with managed resources without diff too

pkg/cmd/scan/output/html_test.go
```diff
@@ -97,9 +97,8 @@ func TestHTML_Write(t *testing.T) {
 						Id:   "diff-id-2",
 						Type: "aws_diff_resource",
 						Source: &resource.TerraformStateSource{
-							State:  "tfstate://state.tfstate",
-							Module: "module",
```
eliecharra, line 101: It can have sense to keep Module here

eliecharra (changes requested):

pkg/cmd/scan/output/html.go
sundowndev: We don't display managed resources in html output, but yeah it make sense to display there's no drift for a single IaC source for example

pkg/cmd/scan/output/html.go
```diff
@@ -96,6 +96,17 @@ func (c *HTML) Write(analysis *analyser.Analysis) error {
...
+			resources := make([]*resource.Resource, 0)
+			resources = append(resources, analysis.Deleted()...)
+			resources = append(resources, analysis.Managed()...)
+
+			for _, d := range analysis.Differences() {
```
eliecharra, line 104: This loop is useless since diff always happen on managed resources

eliecharra (changes requested):
I only one last comment

pkg/cmd/scan/output/html.go
sundowndev: right, thanks

eliecharra (dismissed):

pkg/cmd/scan/output/html_test.go
```diff
@@ -281,3 +335,69 @@ func TestHTML_DistinctResourceTypes(t *testing.T) {
...
+		resources []*resource.Resource
+		value     []string
+	}{
+		{
+			name:      "test empty array",
+			resources: []*resource.Resource{},
+			value:     []string{},
+		},
+		{
+			name: "test empty array",
```
wbeuil, line 351: You need to change this name

wbeuil (changes requested):

pkg/cmd/scan/output/assets/index.tmpl
```diff
@@ -119,15 +126,19 @@
...
                         </div>
                         <div role="rowgroup" class="table-body">
                             {{range $diff := .Differences}}
                             <div role="row" data-kind="resource-changed" class="resource-item">
                                 <div class="row">
-                                    <span role="cell" data-type="resource-id">{{$diff.Res.TerraformId}}</span>
-                                    <span role="cell" data-type="resource-type">{{$diff.Res.TerraformType}}</span>
+                                    <span role="cell">
+                                        <span data-type="resource-id">{{$diff.Res.TerraformId}}</span>
+                                        {{ if $diff.Res.Src }}(<span>{{$diff.Res.SourceString}}</span>){{ end }}
```
wbeuil, line 138: Miss the `TerraformType` if there is no source I think

sundowndev: I think we always have iac source for drifts? but I can add a check anyway

Issue #942, opened 2021-08-26
RFC: How to handle source for expanded resources


eliecharra:
# Status

**DRAFT**

# Problem

driftctl can now display additional information of managed resource (when they appears as missing or changed when using deep mode).

The actual console output look like this

```
Found changed resources:
  From tfstate://terraform.tfstate
    - sfhzii.foobar.driftctl-test.com (my_module.aws_s3_bucket.foobar):
        ~ versioning.0.enabled: false => true
```

A source is composed of 3 strings : **SOURCE**, **NAMESPACE** and **INTERNAL_NAME**

We have a representation of this through the following interface

```go
type Source interface {
	Source() string
	Namespace() string
	InternalName() string
}
```

For Terraform we mapped these fields like this:
- We consider the state file path (`tfstate://terraform.tfstate`) to be the **SOURCE**
- We consider the module (`my_module`) to be the **NAMESPACE**
- We consider the resource name (`foobar`) to be the **INTERNAL_NAME**

Here is the actual implementation of a source of kind `TerraformState`:

```go
type TerraformStateSource struct {
	State  string // mapped to source
	Module string // mapped to namespace
	Name   string // mapped to internal name
}
```

For now we use a method bound to a resource to print a string representation of the source : `my_module.aws_s3_bucket.foobar`

```go
func (r *Resource) SourceString() string {
	if r.Source.Namespace() == "" {
		return fmt.Sprintf("%s.%s", r.ResourceType(), r.Source.InternalName())
	}
	return fmt.Sprintf("%s.%s.%s", r.Source.Namespace(), r.ResourceType(), r.Source.InternalName())
}
```

To be able to match remote and IaC resource, for Terraform we sometimes **expand** resources. 
Expanding a resource mean to create a child resource from a parent resource for the diff process to be consistent. We do not want to debate here for the need of expanders.

The actual problem is : **how to handle virtual resource created inside driftctl, in middlewares**.
For now resource created in expanders do not have a source which can cause crash.

# Proposal : Create a new source type `DriftctlSource`

The first step is to change the actual `Source` interface to this

```go
type Source interface {
	Source() string // mapped to "tfstate://terraform.tfstate"
	Address() string // mapped to "my_module.aws_s3_bucket.foobar"
}
```

We do not really need to split namespace and internal name.

The Terraform state implementation will became:

```go
type TerraformStateSource struct {
	State  string
	Module string
	Type   string // This is a new field
	Name   string
}

[...]

PR #962, opened 2021-08-27
Revert "feat: add iac source to drift in console output"

functionstackx:
Reverts cloudskiff/driftctl#934

## Bug

`deletedResource.Source` == nil if it is deleted.

## Steps to reproduce:
- create ec2 through terraform
- delete ec2 through the console
- run it and `deletedResource.Source` is `nil`

## Steps to Resolve
- resource.Source needs to be declared even when it is a deleted resource

## relates to #901

### DIFF
--- a/pkg/cmd/scan/output/console.go
+++ b/pkg/cmd/scan/output/console.go
@@ -35,27 +35,13 @@ func NewConsole() *Console {
 
 func (c *Console) Write(analysis *analyser.Analysis) error {
 	if analysis.Summary().TotalDeleted > 0 {
-		groupedBySource := make(map[string][]*resource.Resource)
-
-		for _, deletedResource := range analysis.Deleted() {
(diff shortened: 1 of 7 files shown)


eliecharra:
@OrenLeung I'm not sure to get it, a deleted resource mean that it exist in terraform but not in your cloud provider. I mean even if the resource is deleted on the cloud provider, we are still able to determine the source since we determine the source based on IaC (terraform state).

The only kind of resource that should not have a source are unmanaged ones.

Are you sure that you are not in the case covered here : https://github.com/cloudskiff/driftctl/issues/942 ?
We identified some case where the source can be null and the issue #942 is here to debate how can we solve this.

OrenLeung:
@eliecharra i am not sure if it is covered by #942 let me debug a bit further. It might because of a "virtual" EBS

Issue #954, opened 2021-08-27
Support aws_rds_cluster_instance


sjourdan:
**Description**

`aws_rds_cluster_instance` provides an RDS Cluster Instance Resource

**Example**

Doc: https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/rds_cluster_instance

sundowndev:
Following https://github.com/cloudskiff/driftctl/issues/947#issuecomment-911777671, we should create a middleware to handle false positives on `aws_db_instance` resources according to `aws_rds_cluster_instance` resources present in the state. For each `aws_rds_cluster_instance` in the state, ignore any remote `aws_db_instance` that matches the same resource id.

sjourdan:
LGTM 

< 0.16

```sh
❯  LOG_LEVEL=error driftctl scan --tf-provider-version 3.61.0
Found resources not covered by IaC:
  aws_db_instance:
    - aurora-cluster-demo-0
    - aurora-cluster-demo-1
    - aurora-cluster-demo-2
[...]
Found 9 resource(s)
 - 0% coverage
 - 0 resource(s) managed by terraform
     - 0/0 resource(s) out of sync with Terraform state
 - 9 resource(s) not managed by Terraform
 - 0 resource(s) found in a Terraform state but missing on the cloud provider
[...]
```

w/0.16: 

```sh
❯  LOG_LEVEL=error driftctl-0.16-pre scan --tf-provider-version 3.61.0
Found resources not covered by IaC:
[...]
Found 13 resource(s)
 - 30% coverage
 - 4 resource(s) managed by terraform
     - 0/4 resource(s) out of sync with Terraform state
 - 9 resource(s) not managed by Terraform
 - 0 resource(s) found in a Terraform state but missing on the cloud provider

```

PR #1011, opened 2021-09-03
Implement aws_rds_cluster_instance resource

sundowndev:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | no
| 🚀 New feature?   | yes
| ⚠ Deprecations?   | no
| ❌ BC Break       | no
| 🔗 Related issues | #954 
| ❓ Documentation  | https://github.com/cloudskiff/driftctl-docs/pull/126 

## Description

This PR implements the aws_rds_cluster_instance resource. This resource exists in the AWS API as a `aws_db_instance`, that's the reason why it doesn't have an enumerator. I created a middleware to import remote db instances when they match a rds cluster instance from state.


wbeuil (changes requested):
The way you're doing it in the middleware will not let us do a correct diff since you're just copying the remote one in both the state resources and in the remote resources.

We should instead map each rds_cluster_instance into rds_instance.

sundowndev:
> We should instead map each rds_cluster_instance into rds_instance.

@wbeuil not sure to understand what you mean exactly by "mapping into". How should we achieve that ?

wbeuil:
> > We should instead map each rds_cluster_instance into rds_instance.
> 
> @wbeuil not sure to understand what you mean exactly by "mapping into". How should we achieve that ?

Just like we do in other expander middleware like `AwsBucketPolicyExpander` or `AwsInstanceBlockDeviceResourceMapper` we want to add in the state resource list a new resource that you create from `AwsRDSClusterInstanceResourceType`. Otherwise in the case where we want to do a `deep-mode` scan it will never have drifts.

pkg/middlewares/aws_rds_cluster_instance_expander.go
```diff
@@ -0,0 +1,59 @@
...
+		var found bool
+		for _, remoteRes := range dbInstances {
+			// If the db instance's id matches the rds cluster instance's id, import it in the state
+			if remoteRes.ResourceId() == stateRes.ResourceId() {
+				found = true
+				newDbInstance := m.resourceFactory.CreateAbstractResource(aws.AwsDbInstanceResourceType, stateRes.ResourceId(), *stateRes.Attributes())
+				newResourcesFromState = append(newResourcesFromState, newDbInstance)
+				logrus.WithFields(logrus.Fields{
+					"id": newDbInstance.ResourceId(),
+				}).Debug("Created new db instance from RDS cluster instance")
```
moadibfr, line 48: I think you can break here when you found the corresponding id ?

moadibfr (changes requested):

wbeuil (dismissed):
LGTM

Issue #929, opened 2021-08-18
Feature: Add ability to read provider version from .terraform.lock.hcl lock file


nantiferov:
**Description**

Hello. Thank you for this wonderful tool 👍 

I think that it would be nice to have ability to read provider version from .terraform.lock.hcl lock file (if it exists). This way drift will be checked with the same version as used by terraform.

**Example**

By default it seems `3.19.0` is used in `driftctl scan` and it's possible to override it with `--tf-provider-version`.
While .terraform.lock.hcl lock file always has provider version in it, so it could be easily read from there.

sjourdan:
Great idea, we are prioritizing it. More specs and use-cases are coming

eliecharra:
**Note for maintainers:**

We can use https://github.com/hashicorp/hcl#hcl to read the `.terraform.lock.hcl` file if it exist and use the decoded version from it.

We should change the `ProviderInstaller` constructor `NewProviderInstaller` to remove the version, and instead pass the version to the `Install()` func : `func (p *ProviderInstaller) Install(version string) (string, error)`.

And so we can handle the new logic to retrieve the version from the lock file in `func (p *TerraformProvider) configure(alias string)` at the begining of the func.

We should also change each `resourceSchemaRepository.Init`  call to use `provider.Version()`

The default version statement should be moved within provider constructor method, for AWS it should be in `NewAWSTerraformProvider`

sjourdan:
I couldn't replicate the required behavior. 

w/0.15 as expected, 3.19 default version was loaded: 

```sh
❯  head .terraform.lock.hcl 
provider "registry.terraform.io/hashicorp/aws" {
  version     = "3.44.0"
  constraints = ">= 3.36.0, 3.44.0"
[...]
❯  LOG_LEVEL=debug driftctl scan 
[...]
DEBU[0000] Found existing provider path=/home/sjourdan/.driftctl/plugins/linux_amd64/terraform-provider-aws_v3.19.0_x5
```

w/0.16 unexpectedly loaded 3.19 too: 

```sh
❯  head .terraform.lock.hcl 
provider "registry.terraform.io/hashicorp/aws" {
  version     = "3.44.0"
  constraints = ">= 3.36.0, 3.44.0"
[...]
❯  driftctl-0.16-pre scan
DEBU[0000] Found existing provider path=/home/sjourdan/.driftctl/plugins/linux_amd64/terraform-provider-aws_v3.19.0_x5
```

[...]

PR #1016, opened 2021-09-06
Read provider version from .terraform.lock.hcl lock file when possible

sundowndev:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | no
| 🚀 New feature?   | yes
| ⚠ Deprecations?   | no
| ❌ BC Break       | no
| 🔗 Related issues | #929
| ❓ Documentation  | https://github.com/cloudskiff/driftctl-docs/pull/130 

## Description

This PR implement a lock package that parses the content of a Terraform lock file. This logic is used in the Terraform supplier in order to install a specific version when we found a Terraform lock file in at the given path.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -7,7 +7,6 @@ require (
 	github.com/Azure/azure-sdk-for-go/sdk/azcore v0.19.0
 	github.com/Azure/azure-sdk-for-go/sdk/azidentity v0.11.0
 	github.com/Azure/azure-sdk-for-go/sdk/storage/armstorage v0.2.0
-	github.com/Azure/azure-sdk-for-go/sdk/to v0.1.4
 	github.com/Azure/go-autorest/autorest v0.11.3
 	github.com/aws/aws-sdk-go v1.38.68
 	github.com/bmatcuk/doublestar/v4 v4.0.1
@@ -19,6 +18,7 @@ require (
 	github.com/hashicorp/go-hclog v0.9.2
 	github.com/hashicorp/go-plugin v1.3.0
 	github.com/hashicorp/go-version v1.3.0
+	github.com/hashicorp/hcl/v2 v2.7.2
 	github.com/hashicorp/terraform v0.14.0
 	github.com/hashicorp/terraform-exec v0.14.0
 	github.com/jarcoal/httpmock v1.0.6
@@ -40,6 +40,7 @@ require (
 	github.com/yudai/gojsondiff v1.0.0
 	github.com/zclconf/go-cty v1.8.4
 	go.uber.org/atomic v1.4.0
+	golang.org/x/net v0.0.0-20210913180222-943fd674d43e // indirect
 	golang.org/x/oauth2 v0.0.0-20210805134026-6f1e6394065a
 	golang.org/x/sync v0.0.0-20210220032951-036812b2e83c
 	google.golang.org/api v0.54.0
--- a/pkg/remote/aws/provider.go
+++ b/pkg/remote/aws/provider.go
@@ -2,7 +2,6 @@ package aws
 
 import (
 	"github.com/aws/aws-sdk-go/aws/session"
-
 	"github.com/cloudskiff/driftctl/pkg/output"
 	"github.com/cloudskiff/driftctl/pkg/remote/terraform"
 	tf "github.com/cloudskiff/driftctl/pkg/terraform"
@@ -45,6 +44,9 @@ type AWSTerraformProvider struct {
 }
 
 func NewAWSTerraformProvider(version string, progress output.Progress, configDir string) (*AWSTerraformProvider, error) {
+	if version == "" {
+		version = "3.19.0"
+	}
 	p := &AWSTerraformProvider{
 		version: version,
 		name:    "aws",
--- a/pkg/remote/terraform/provider.go
+++ b/pkg/remote/terraform/provider.go
@@ -9,7 +9,6 @@ import (
 	"time"
 
 	"github.com/cloudskiff/driftctl/pkg/output"
-
 	"github.com/eapache/go-resiliency/retrier"
 	"github.com/hashicorp/terraform/plugin"
 	"github.com/hashicorp/terraform/plugin/discovery"
@@ -88,7 +87,6 @@ func (p *TerraformProvider) Runner() *parallel.ParallelRunner {
 }
 
 func (p *TerraformProvider) configure(alias string) error {
-
 	providerPath, err := p.providerInstaller.Install()
 	if err != nil {
 		return err
--- a/pkg/terraform/lock/lockfile.go
+++ b/pkg/terraform/lock/lockfile.go
@@ -1,0 +1,54 @@
+package lock
+
+import (
+	"strings"
+
+	"github.com/hashicorp/hcl/v2/gohcl"
+	"github.com/hashicorp/hcl/v2/hclparse"
+)
+
+type ProviderBlock struct {
+	Address     string   `hcl:"address,label"`
+	Version     string   `hcl:"version,attr"`
+	Constraints string   `hcl:"constraints,optional"`
+	Hashes      []string `hcl:"hashes,optional"`
+}
+
+// ProviderAddress encapsulates a single provider type. In the future this will be
+// extended to include additional fields including Namespace and SourceHost
+type ProviderAddress struct {
+	Type      string
+	Namespace string
+	Hostname  string
+}
+
+func (p *ProviderAddress) String() string {
+	return strings.Join([]string{p.Hostname, p.Namespace, p.Type}, "/")
+}
+
+type Lockfile struct {
+	Providers []ProviderBlock `hcl:"provider,block"`
+}
+
+func (l *Lockfile) GetProviderByAddress(addr *ProviderAddress) *ProviderBlock {
+	for _, p := range l.Providers {
+		if p.Address == addr.String() {
+			return &p
+		}
+	}
+	return nil
+}
+
+func ReadLocksFromFile(filename string) (*Lockfile, error) {
+	var lock Lockfile
+
+	parser := hclparse.NewParser()
+	f, diags := parser.ParseHCLFile(filename)
+	if diags.HasErrors() {
+		return &lock, diags
+	}
+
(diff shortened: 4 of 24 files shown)


pkg/remote/aws/provider.go
```diff
@@ -2,6 +2,7 @@ package aws
 
 import (
 	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/hashicorp/terraform/addrs"
```
eliecharra, line 5: We cannot use this struct anymore since in the newer version this is unexported (moved under `internal` pkg)

We should have our own struct instead under `pkg/terraform`

pkg/remote/terraform/lockfile.go
```diff
@@ -0,0 +1,223 @@
...
+	"github.com/hashicorp/hcl/v2/gohcl"
+	"github.com/hashicorp/hcl/v2/hclparse"
+	"github.com/hashicorp/terraform/addrs"
+	"github.com/hashicorp/terraform/tfdiags"
+)
+
+type LockFileProvider struct {
+	Addr        addrs.Provider
+	Constraints string   `hcl:"constraints"`
+	Hashes      []string `hcl:"hashes"`
```
eliecharra, line 17: Can we remove theses fields from this struct ? We only want the version.

pkg/remote/terraform/provider.go
```diff
@@ -24,13 +25,16 @@ import (
 	tf "github.com/cloudskiff/driftctl/pkg/terraform"
 )
 
+const LockFileName = ".terraform.lock.hcl"
```
eliecharra, line 28: I don't think we need this exported const since this is only used at 1 place below.

pkg/remote/terraform/lockfile.go
```diff
@@ -0,0 +1,223 @@
+package terraform
```
eliecharra, line 1: It could be nice to put this in `pkg/terraform/lock` instead of remote, this is not really related to driftctl remote stuff.

We should probably only expose a single method from this pkg to return the related provider config from this pkg.

We may return an error (or terraform diags) too and log it in debug

pkg/remote/terraform/provider.go
```diff
@@ -88,8 +92,16 @@ func (p *TerraformProvider) Runner() *parallel.ParallelRunner {
 }
 
 func (p *TerraformProvider) configure(alias string) error {
+	var version string
+	lockFile, diags := LoadLocksFromFile(LockFileName)
```
eliecharra, line 96: We should log in debug the diags, it will help us investigate some case where we wont be able to read the lockfile for some reason

eliecharra (changes requested):
I left some architectural comments, I will do another deeper review pass after that

pkg/remote/terraform/provider.go
```diff
@@ -88,8 +89,14 @@ func (p *TerraformProvider) Runner() *parallel.ParallelRunner {
 }
 
 func (p *TerraformProvider) configure(alias string) error {
+	var version string
+	lockFile, _ := lock.LoadLocksFromFile(".terraform.lock.hcl")
```
eliecharra, line 93: My first comment is still relevant:

> We should log in debug the diags, it will help us investigate some case where we wont be able to read the lockfile for some reason

We should not log something only when everything is fine, if there is some diags err we should log them in debug too.

eliecharra (changes requested):
We have a looot of useless code in the lockfile read logic. Can you check and remove useless stuff, I think we can have simpler hcl decode logic

pkg/remote/terraform/provider.go
sundowndev: I did add log messages in the LoadLocksFromFile func https://github.com/cloudskiff/driftctl/blob/4a455a1b9e25cf5ff0e91f3b042fcc6037b7931d/pkg/remote/terraform/lock/lockfile.go#L44-L55

eliecharra: My bad I missed this one 🙏🏻 Anyway, I think that we should not bind the lockfile pkg to a logging strategy. Since we log here when everything is fine, it makes more sense to log errors at the same place IMHO. Let's keep the lockfile logic return it's errors without any error handling logic within.

eliecharra:
> We have a looot of useless code in the lockfile read logic. Can you check and remove useless stuff, I think we can have simpler hcl decode logic.

Maybe I wasn't enough explicit whith this sentence. In the actual code we are still depending of terraform code (for diags), we should not have any dependency.

We also do not need to do the whole lock file validation process while reading it, here we only want to read the version. The validation of the file is done by terraform.

I think we could simplify things a lot, I tried to write a minimal decoding code below as an example :

```go
func main() {
	parser := hclparse.NewParser()
	f, diags := parser.ParseHCLFile(".terraform.lock.hcl")
	if diags.HasErrors() {
		fmt.Errorf("%+v", diags)
		os.Exit(1)
	}

	schema := &hcl.BodySchema{
		Blocks: []hcl.BlockHeaderSchema{
			{
				Type:       "provider",
				LabelNames: []string{"address"},
			},
		},
	}

	content, diags := f.Body.Content(schema)
	if diags.HasErrors() {
		fmt.Errorf("%+v", diags)
		os.Exit(1)
	}
	for _, provider := range content.Blocks.OfType("provider") {
		providerAddr := provider.Labels[0]
		attrs, _ := provider.Body.JustAttributes()
		for _, attr := range attrs {
			val, diags := attr.Expr.Value(nil)
			if diags.HasErrors() {
				fmt.Errorf("%+v", diags)
				os.Exit(1)
			}
			if attr.Name == "version" {
				fmt.Printf("%s@%s\n", providerAddr, val.AsString())
			}
		}
	}
}
```

eliecharra:
Even simpler way with this approach :

```go
type Provider struct {
	Address string `hcl:"address,label"`
	Version string `hcl:"version"`
}

type Lockfile struct {
	Providers []Provider `hcl:"provider,block"`
}

func main() {
	parser := hclparse.NewParser()
	f, diags := parser.ParseHCLFile(".terraform.lock.hcl")
	if diags.HasErrors() {
		fmt.Errorf("%+v", diags)
		os.Exit(1)
	}

	lock := &Lockfile{}
	_ = gohcl.DecodeBody(f.Body, nil, lock)

	fmt.Printf("%s : %s", lock.Providers[0].Address, lock.Providers[0].Version)
}
```

pkg/terraform/lock/lockfile.go
```diff
@@ -0,0 +1,58 @@
...
+	var lock Lockfile
+
+	parser := hclparse.NewParser()
+	f, diags := parser.ParseHCLFile(filename)
+	if diags.HasErrors() {
+		return &lock, diags
+	}
+
+	diags = gohcl.DecodeBody(f.Body, nil, &lock)
+	if diags != nil {
```
eliecharra, line 53: This is useless, you can return `return &lock, diags` directly

pkg/terraform/lock/lockfile_test.go
```diff
@@ -0,0 +1,122 @@
...
+			filepath: "testdata/lockfile_invalid_type-3.hcl",
+			assert: func(locks *Lockfile, err error) {
+				provider := locks.Provider(&ProviderAddress{
+					Type:      "google-beta",
+					Namespace: "hashicorp",
+					Hostname:  "registry.terraform.io",
+				})
+
+				assert.Len(t, locks.Providers, 2)
+				assert.Equal(t, "2.71.0", provider.Version)
```
eliecharra, line 108: You should assert on `tt` and pass it as a parameter since you are running subtests `c.assert(tt, locks, err)`

```go
assert: func(tt *testing.T, locks *Lockfile, err error) {
```

pkg/terraform/lock/lockfile.go
```diff
@@ -0,0 +1,58 @@
...
+
+func (p *ProviderAddress) String() string {
+	return fmt.Sprintf("%s/%s/%s", p.Hostname, p.Namespace, p.Type)
+}
+
+type Lockfile struct {
+	Providers []Provider `hcl:"provider,block"`
+}
+
+func (l *Lockfile) Provider(addr *ProviderAddress) *Provider {
```
eliecharra, line 34: We cant name everything provider 😅 `s/Provider/GetProviderByAddress/`

pkg/terraform/lock/lockfile.go
```diff
@@ -0,0 +1,58 @@
...
+// ProviderAddress encapsulates a single provider type. In the future this will be
+// extended to include additional fields including Namespace and SourceHost
+type ProviderAddress struct {
+	Type      string
+	Namespace string
+	Hostname  svchost.Hostname
+}
+
+func (p *ProviderAddress) String() string {
+	return fmt.Sprintf("%s/%s/%s", p.Hostname, p.Namespace, p.Type)
```
eliecharra, line 27: `strings.Join(string[]{p.HostName, p.Namespace, p.Type}, "/")`

eliecharra (changes requested):

go.mod
```diff
@@ -7,7 +7,6 @@ require (
 	github.com/Azure/azure-sdk-for-go/sdk/azcore v0.19.0
 	github.com/Azure/azure-sdk-for-go/sdk/azidentity v0.11.0
 	github.com/Azure/azure-sdk-for-go/sdk/storage/armstorage v0.2.0
-	github.com/Azure/azure-sdk-for-go/sdk/to v0.1.4
```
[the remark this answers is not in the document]
sundowndev: Are you sure? I feel like it's unused, running `go mod tidy` removed it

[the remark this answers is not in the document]
eliecharra: https://github.com/cloudskiff/driftctl/blob/main/pkg/remote/azurerm/repository/storage_test.go#L192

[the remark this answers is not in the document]
sundowndev: Ok the diff will disappear  after a rebase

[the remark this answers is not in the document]
sundowndev: It's `github.com/Azure/azure-sdk-for-go/sdk/azcore/to` so we can safely remove it

eliecharra (approved):
That looks great 👍🏻 I'm gonna merge it when the doc PR will be ready

PR #1067, opened 2021-09-15
Add google_compute_router

sundowndev:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | no
| 🚀 New feature?   | yes
| ⚠ Deprecations?   | no
| ❌ BC Break       | no
| 🔗 Related issues | #1004 
| ❓ Documentation  | https://github.com/cloudskiff/driftctl-docs/pull/122

### DIFF
--- a/pkg/iac/terraform/state/test/google_compute_router/terraform.tfstate
+++ b/pkg/iac/terraform/state/test/google_compute_router/terraform.tfstate
@@ -1,0 +1,147 @@
+{
+  "version": 4,
+  "terraform_version": "0.15.5",
+  "serial": 6,
+  "lineage": "5dbb3a54-27a7-9a8d-0c76-8d44f1873015",
+  "outputs": {},
+  "resources": [
+    {
+      "mode": "managed",
+      "type": "google_compute_router",
+      "name": "router-0",
+      "provider": "provider[\"registry.terraform.io/hashicorp/google\"]",
+      "instances": [
+        {
+          "schema_version": 0,
+          "attributes": {
+            "bgp": [
+              {
+                "advertise_mode": "CUSTOM",
+                "advertised_groups": [
+                  "ALL_SUBNETS"
+                ],
+                "advertised_ip_ranges": [
+                  {
+                    "description": "",
+                    "range": "1.2.3.4"
+                  },
+                  {
+                    "description": "",
+                    "range": "6.7.0.0/16"
+                  }
+                ],
+                "asn": 64514
+              }
+            ],
+            "creation_timestamp": "2021-09-16T07:40:46.101-07:00",
+            "description": "",
+            "id": "projects/cloudskiff-dev-raphael/regions/us-central1/routers/test-router-0",
+            "name": "test-router-0",
+            "network": "https://www.googleapis.com/compute/v1/projects/cloudskiff-dev-raphael/global/networks/my-network",
+            "project": "cloudskiff-dev-raphael",
+            "region": "us-central1",
+            "self_link": "https://www.googleapis.com/compute/v1/projects/cloudskiff-dev-raphael/regions/us-central1/routers/test-router-0",
+            "timeouts": null
+          },
+          "sensitive_attributes": [],
+          "private": "eyJlMmJmYjczMC1lY2FhLTExZTYtOGY4OC0zNDM2M2JjN2M0YzAiOnsiY3JlYXRlIjoyNDAwMDAwMDAwMDAsImRlbGV0ZSI6MjQwMDAwMDAwMDAwLCJ1cGRhdGUiOjI0MDAwMDAwMDAwMH19",
+          "dependencies": [
+            "google_compute_network.foobar"
+          ]
+        }
+      ]
+    },
+    {
+      "mode": "managed",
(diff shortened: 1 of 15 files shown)


.gitignore
```diff
@@ -31,3 +31,4 @@
 
 ### driftctl
 /.driftignore
+.driftctl
```
eliecharra, line 34: This can be removed

pkg/iac/terraform/state/test/google_compute_router/terraform.tfstate
```diff
@@ -0,0 +1,173 @@
+{
+  "version": 4,
+  "terraform_version": "0.15.5",
+  "serial": 6,
+  "lineage": "5dbb3a54-27a7-9a8d-0c76-8d44f1873015",
+  "outputs": {},
+  "resources": [
+    {
+      "mode": "managed",
+      "type": "google_compute_network",
```
eliecharra, line 10: You can remove this resource from the state, we want to keep this test minimalist even if we have an inconsistent state.

pkg/remote/google/util.go
```diff
@@ -0,0 +1,23 @@
...
+)
+
+type googleResourceAttributes struct {
+	ProjectName string
+	Region      string
+}
+
+func parseResourceName(name string) (googleResourceAttributes, error) {
+	attrs := googleResourceAttributes{}
+	splittedName := strings.Split(name, "/")
```
eliecharra, line 15: As we talked during our live session the other day, we should use a regex here to be sure to extract things at the right index. We cannot predict that we'll always get the same structure for resource names.

For example for `google_compute_firewall` we do not have a region `//compute.googleapis.com/projects/cloudskiff-dev-elie/global/firewalls/test-firewall-0`

For project we should match something like `projects\/(.*?)\/`, same idea for region

pkg/remote/google_compute_scanner_test.go
```diff
@@ -148,3 +148,143 @@ func TestGoogleComputeFirewall(t *testing.T) {
...
+	}
+
+	providerVersion := "3.78.0"
+	schemaRepository := testresource.InitFakeSchemaRepository("google", providerVersion)
+	googleresource.InitResourcesMetadata(schemaRepository)
+	factory := terraform.NewTerraformResourceFactory(schemaRepository)
+
+	for _, c := range cases {
+		t.Run(c.test, func(tt *testing.T) {
+			shouldUpdate := c.dirName == *goldenfile.Update
```
eliecharra, line 238: You could simplify this test a lot since we are not in deep mode, you can remove dirName and other deep mode related stuff. Check my PR on `google_compute_instance` for example

eliecharra (changes requested):

pkg/remote/google/util.go
sundowndev: Since it's a non deep-mode implementation I'll undo that change and do it in another PR

.gitignore
sundowndev: I added it because terraform reader tests were creating a `.driftctl` folder within codebase and I committed it by mistake multiple times

eliecharra: We should investigate why test are creating theses folders instead of just ignoring the consequences. Can you create a maintenance issue for that ?

pkg/remote/google/util.go
```diff
@@ -0,0 +1,9 @@
+package google
+
+import (
+	"strings"
+)
+
+func trimResourceName(name string) string {
+	return strings.TrimPrefix(name, "//compute.googleapis.com/")
```
eliecharra, line 8: This should for for every kind of service and not only compute if you name the func like this.

We should either change the behavior to be able to trim any previx regardless of the service type or rename the func name for something like `trimComputeResourceName`

eliecharra (changes requested):
One last comment

pkg/remote/google/util.go
sundowndev: With a regex we could match `\/\/[\w]+.googleapis.com\/` WDYT ?

eliecharra: > With a regex we could match \/\/[\w]+.googleapis.com\/ WDYT ?

We should be sure that any service ends with `.googleapis.com` 

Another solution could be to remove every prefix `^\/\/.*?\/`, I don't know if it can be dangerous ...

sundowndev: Yes this is a common suffix for all resource types (except for k8s). See https://cloud.google.com/asset-inventory/docs/resource-name-format

We can also create a function for each service and change that later when we'll have more experience with GCP

eliecharra: Ok so I think `^\/\/[\w]+.googleapis.com\/` can do the job (do not forget the `^` to match the beginning of the string) since everything ends with `googleapis.com`.

I think you missread since kube related resources are under `container.googleapis.com`

image

sundowndev: > I think you missread since kube related resources are under container.googleapis.com

yes indeed

eliecharra (dismissed):

eliecharra:
@sundowndev Can you add a doc PR too 🙏🏻

Issue #960, opened 2021-08-27
Support AWS API Gateway


sjourdan:
**Description**

This service is highly split into small resources, let group them here first. 

- [x] #1110
- [x] #1111
- [x] #1112
- [x] #1113
- [x] #1114
- [x] #1125
- [x] #1115
- [ ] aws_api_gateway_deployment (dropped)
- [x] #1157
- [x] #1179
- [x] #1199
- [x] #1140
- [x] #1164
- [x] #1170
- [x] #1150
- [x] #1133
- [x] #1117
- [x] #1116
- [x] #1109

**Example**

TF doc: https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/api_gateway_resource

eliecharra:
@sjourdan How to you want to handle testing for this one ? We merged `rest_api` resource, should I put this issue in review ?

wbeuil:
All v1 resources merged

PR #1075, opened 2021-09-21
Add api_gateway_account

wbeuil:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | no
| 🚀 New feature?   | yes
| ⚠ Deprecations?   | no
| ❌ BC Break       | no
| 🔗 Related issues | #960
| ❓ Documentation  | yes

## Description

Add aws_api_gateway_account resource

### DIFF
--- a/pkg/remote/aws/api_gateway_account_enumerator.go
+++ b/pkg/remote/aws/api_gateway_account_enumerator.go
@@ -1,0 +1,46 @@
+package aws
+
+import (
+	"github.com/cloudskiff/driftctl/pkg/remote/aws/repository"
+	remoteerror "github.com/cloudskiff/driftctl/pkg/remote/error"
+	"github.com/cloudskiff/driftctl/pkg/resource"
+	"github.com/cloudskiff/driftctl/pkg/resource/aws"
+)
+
+type ApiGatewayAccountEnumerator struct {
+	repository repository.ApiGatewayRepository
+	factory    resource.ResourceFactory
+}
+
+func NewApiGatewayAccountEnumerator(repo repository.ApiGatewayRepository, factory resource.ResourceFactory) *ApiGatewayAccountEnumerator {
+	return &ApiGatewayAccountEnumerator{
+		repository: repo,
+		factory:    factory,
+	}
+}
+
+func (e *ApiGatewayAccountEnumerator) SupportedType() resource.ResourceType {
+	return aws.AwsApiGatewayAccountResourceType
+}
+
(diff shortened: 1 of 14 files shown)


pkg/remote/aws/api_gateway_account_enumerator.go
```diff
@@ -0,0 +1,51 @@
...
+	if err != nil {
+		return nil, remoteerror.NewResourceListingError(err, string(e.SupportedType()))
+	}
+
+	results := make([]*resource.Resource, 1)
+
+	if account != nil {
+		attrs := map[string]interface{}{}
+		if account.CloudwatchRoleArn != nil && *account.CloudwatchRoleArn != "" {
+			attrs["cloudwatch_role_arn"] = *account.CloudwatchRoleArn
```
eliecharra, line 37: Why do we need this attribute ?

pkg/remote/aws/api_gateway_account_enumerator.go
```diff
@@ -0,0 +1,51 @@
...
+		}
+
+		results = append(
+			results,
+			e.factory.CreateAbstractResource(
+				string(e.SupportedType()),
+				"api-gateway-account",
```
eliecharra, line 44: Is this hardcoded on terraform provider side too ?

eliecharra (commented):
Some small comments

pkg/remote/aws/api_gateway_account_enumerator.go
wbeuil: Yes exactly it is hardcoded in the provider 
image

pkg/remote/aws/api_gateway_account_enumerator.go
wbeuil: We could totally ditch it, it was more for preparing an eventual "deep-mode" compatibility since there is only one attribute in this resource that is needed by the provider.

image

pkg/remote/aws/api_gateway_account_enumerator.go
eliecharra: 🆗 👌🏻

pkg/remote/aws/api_gateway_account_enumerator.go
eliecharra: Yeah I think we should not add useless code for now since it is only a deep mode required field.

eliecharra (approved):

eliecharra (approved):

PR #1080, opened 2021-09-28
Fix github acceptance test

eliecharra:
Reference : https://github.com/integrations/terraform-provider-github/issues/908


sundowndev (approved):

majormoses:
I am confused are we supposed to specify a `node_id` or a `name`? The experience as a consumer of this provider resource has been confusing. I am not sure that this is correct, we are "fixing" a test by not testing an expected use case according to the documentation. If we no longer expect people to use a `node_id` and should only use a name then we need to update the documentation to help people understand what they are expected to change to upgrade.

https://registry.terraform.io/providers/integrations/github/latest/docs/resources/branch_protection#repository_id

> (Required) The name or node ID of the repository associated with this branch protection rule.

PR #1196, opened 2021-10-28
Create changelog generator script

sundowndev:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | no
| 🚀 New feature?   | no
| ⚠ Deprecations?   | no
| ❌ BC Break       | no
| 🔗 Related issues | #...
| ❓ Documentation  | no 

## Description

This bash script uses the GH CLI to generate a changelog from the previous tag to a base branch. It must be ran **before** creating the new release's tag.

### DIFF
--- a/scripts/changelog.sh
+++ b/scripts/changelog.sh
@@ -3,11 +3,23 @@
 # Please note that this script only work with Github repositories.
 # Prerequisites: git, github cli
 
-GHCLI_BIN="ghcli"
-REPO="cloudskiff/driftctl"
+GHCLI_BIN="gh"
+REPO="snyk/driftctl"
 LATEST_TAG=$(git describe --abbrev=0) # Get the least created tag
-DEFAULT_BRANCH=$(git symbolic-ref refs/remotes/origin/HEAD | sed 's@^refs/remotes/origin/@@')
+DEFAULT_BRANCH=origin/HEAD
 BASE=$DEFAULT_BRANCH # Change this if you don't want to use the default branch as base
+
+# Check GH cli is installed
+if ! which $GHCLI_BIN &> /dev/null; then
+    echo "GitHub CLI ($GHCLI_BIN) is not installed, visit https://github.com/cli/cli#installation"
+    exit 1
+fi
+
+# Check GH authentication
+if ! $GHCLI_BIN auth status &> /dev/null; then
+    echo "You are not logged into any GitHub hosts. Run gh auth login to authenticate."
+    exit 1
+fi
 
 # Compare $BASE branch with the latest tag
 # Keep IDs of merged pull requests
@@ -44,5 +56,3 @@ for change in "${CHANGES[@]}"; do
         echo $change | sed "s/\[map\[$PARTITION_COLUMN.*//"
     fi
 done
-
-exit 0
--- a/scripts/changelog.sh
+++ b/scripts/changelog.sh
@@ -1,0 +1,48 @@
+#!/usr/bin/env bash
+
+# Please note that this script only work with Github repositories.
+# Prerequisites: git, github cli
+
+GHCLI_BIN="ghcli"
+REPO="cloudskiff/driftctl"
+LATEST_TAG=$(git describe --abbrev=0) # Get the least created tag
+DEFAULT_BRANCH=$(git symbolic-ref refs/remotes/origin/HEAD | sed 's@^refs/remotes/origin/@@')
+BASE=$DEFAULT_BRANCH # Change this if you don't want to use the default branch as base
+
+# Compare $BASE branch with the latest tag
+# Keep IDs of merged pull requests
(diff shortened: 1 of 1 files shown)


scripts/changelog.sh
```diff
@@ -0,0 +1,48 @@
+#!/usr/bin/env bash
+
+# Please note that this script only work with Github repositories.
+# Prerequisites: git, github cli
+
+GHCLI_BIN="ghcli"
+REPO="cloudskiff/driftctl"
+LATEST_TAG=$(git describe --abbrev=0) # Get the least created tag
+DEFAULT_BRANCH=$(git symbolic-ref refs/remotes/origin/HEAD | sed 's@^refs/remotes/origin/@@')
+BASE=$DEFAULT_BRANCH # Change this if you don't want to use the default branch as base
```
craigfurman, line 10:
```suggestion
BASE=origin/HEAD
```

Is there any reason we need to do this default branch logic? Can't we just log from `origin/HEAD`? If I understand the sed expression correctly, we're using the local pointer to `main`, rather than the remote-tracking origin/main or origin/HEAD updated by `git fetch`, and I'm not sure what the benefit of that is.

scripts/changelog.sh
```diff
@@ -0,0 +1,48 @@
+#!/usr/bin/env bash
+
+# Please note that this script only work with Github repositories.
+# Prerequisites: git, github cli
+
+GHCLI_BIN="ghcli"
```
craigfurman, line 6: On what system is this the default name for the GitHub CLI? Is this https://github.com/cli/cli, commonly known on PATH as `gh`? I had to edit the script to test this because GHCLI_BIN is not overridable, but before we make it overridable, is there any reason we can't hardcode `gh` instead?

scripts/changelog.sh
```diff
@@ -0,0 +1,48 @@
...
+REPO="cloudskiff/driftctl"
```
craigfurman, line 7:
```suggestion
REPO="snyk/driftctl"
```

craigfurman (commented):
A couple of notes, and also would you mind pasting some expected output so that users can review it? I had trouble getting this script running in the first place, so I'm not sure I'm seeing the right thing 🙂

sundowndev:
> A couple of notes, and also would you mind pasting some expected output so that users can review it? I had trouble getting this script running in the first place, so I'm not sure I'm seeing the right thing slightly_smiling_face

@craigfurman Thanks for your review, this script was initially created for myself only, it can be improved for sure. Expected output can be found in [the release page](https://github.com/snyk/driftctl/releases), we actually run this script at each release to generate the changelog.

scripts/changelog.sh
sundowndev: No `ghcli` was just an alias from my env. I'm wondering if we should add a few lines to install the GH cli if it's not already installed (targeting a fixed version), and put the binary in `bin`.

craigfurman: Personally, I absolutely hate it when development scripts install tools for me 😅. I think we're overthinking this: if the github CLI breaks its interface, let's have a think about version pinning then. In the mean time, how about we just call `gh`?

eliecharra: > I absolutely hate it when development scripts install tools for me

I agree, we should absolutely avoid to do that. 
What about a docker container if we want to provide a consistent environment, maybe this is overkill for that ? 
Can we try not to use gh CLI at all ?

eliecharra:
> A couple of notes, and also would you mind pasting some expected output so that users can review it? I had trouble getting this script running in the first place, so I'm not sure I'm seeing the right thing slightly_smiling_face

Thanks a lot for having a first pass on that @craigfurman I really appreciate since I still haven't find the time to review it 👍🏻 
Would you mind if I assign you on this one ?

scripts/changelog.sh
craigfurman: Requiring docker would certainly be overkill. IMO, making scripts sufficiently parameterisable to be runnable with `docker run` is fine, but we shouldn't need linux containers to run a changelog builder.

sundowndev:
@craigfurman I've force pushed to rebase along with a commit to address your review comments.

- Binary name was changed to `gh`
- Two checks were added to check the GH cli is installed and auth configured
- Base branch is now hard-coded to `origin/HEAD`

craigfurman (approved):
Nice one! The output I got from trying this out looks sensible to me, but I trust you to know best on that.

sundowndev:
> The output I got from trying this out looks sensible to me

Can you elaborate on that, may be something isn't clear enough ?

craigfurman:
Looks sensible == looks fine!

I rebased this locally on https://github.com/snyk/driftctl/pull/1252 and ran the script and it appeared to print the relevant PR summaries. If you're happy, please merge 🙂

Issue #1203, opened 2021-11-04
YAML document in the body attribute of `aws_api_gateway_rest_api` resource shouldn't return an error


wbeuil:
**Description**
One of our users in the community Discord (@val123) came with a problem with an infra that has specific `aws_api_gateway_rest_api` resources that use the `body` attribute to spin up a rest API. Every time he/she tried to scan the infra, there was an error saying:

```bash
DEBU[0086] Starting middleware middleware=middlewares.AwsApiGatewayRestApiExpander
TRAC[0086] Exiting scan cmd
DEBU[0086] Closing providers
DEBU[0086] Closing provider key=aws
DEBU[0086] Closing gRPC client alias=us-east-2
TRAC[0086] 2021/11/02 14:32:31 [TRACE] GRPCProvider: Close
TRAC[0086] [TerraformPlugin] received EOF, stopping recv loop[[err rpc error: code = Unavailable desc = error reading from server: EOF]]
TRAC[0086] [TerraformPlugin] plugin process exited[[path /Users/myname/.driftctl/plugins/darwin_amd64/terraform-provider-aws_v3.19.0_x5 pid 25215]]
TRAC[0086] [TerraformPlugin] plugin exited[[]]
TRAC[0086] Exited
DEBU[0086] Sentry captured error cause_type=*json.SyntaxError error_type=*json.SyntaxError
invalid character '-' in numeric literal
DEBU[0086] Flushed Sentry events success=%!s(bool=true) timeout=1m0s
LOG_LEVEL=trace driftctl scan --from  --send-crash-report  9.95s user 1.96s system 13% cpu 1:26.40 total
```

**How to reproduce**
Apply this terraform

```hcl
resource "aws_api_gateway_rest_api" "foo" {
  name        = "foo"
  body = templatefile("${path.module}/template.tpl", { title = "foo" })
}
```

with a template file like this one:

```yaml
---
openapi: "3.0.1"
info:
  title: ${title}
  description: ComputelessBlog
paths:
  /{path+}:
    get:
      responses:
        200:
          description: "200 response"
          content:
            text/html:
              schema:
                $ref: "#/components/schemas/Empty"
      x-amazon-apigateway-integration:
        type: "mock"
        responses:
          default:
            statusCode: "200"
        passthroughBehavior: "never"
        httpMethod: "POST"
components:
  schemas:
    Empty:
      type: object
      title: Empty Schema
      description: Empty Schema
```

**Possible Solution**

[...]

PR #1204, opened 2021-11-04
Middleware should handle YAML body document now

wbeuil:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | yes
| 🚀 New feature?   | no
| ⚠ Deprecations?   | no
| ❌ BC Break       | no
| 🔗 Related issues | #1203
| ❓ Documentation  | no

## Description
Add the same yaml package as the one used in the kin-openapi package to parse OpenAPI document.

After we try to unmarshall json document, if it fails due to syntax error, let's try to unmarshall with the yaml format.


eliecharra (approved):
Thanks 👍🏻

PR #1205, opened 2021-11-04
Fix unbuffered channels

wbeuil:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | no
| 🚀 New feature?   | no
| ⚠ Deprecations?   | no
| ❌ BC Break       | no
| 🔗 Related issues | none
| ❓ Documentation  | no

## Description

As per this [issue](https://youtrack.jetbrains.com/issue/GO-11092), with Go v1.17 our linter throws errors on unbuffered channel.

What it means for us to use an unbuffered one, I will paste this from the [vet tool](https://tip.golang.org/doc/go1.17#vet):

```
Using an unbuffered channel risks missing signals sent on them as signal.Notify does not block when sending to a channel.
```

The simple answer is to add a capacity of 1 to both of our unbuffered channel.


sundowndev (approved):

moadibfr (approved):
from what https://gobyexample.com/signals show buffer size 1 seems to be ok for sigint/sigterm as there will always be only one signal sent. So lgtm

PR #1207, opened 2021-11-04
Put github.com/ghodss/yaml as direct dependency

sundowndev:
Following #1204, that yaml dependency was mistakenly added as a indirect dependency, but `go mod tidy` disagrees.


eliecharra (approved):
Nice catch

PR #1208, opened 2021-11-04
gen-driftignore input/output UX

craigfurman:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | no
| 🚀 New feature?   | yes
| ⚠ Deprecations?   | no
| ❌ BC Break       | yes
| 🔗 Related issues | #...
| ❓ Documentation  | yes 

## Description

The overall intent behind this PR is to (subjectively, IMO) make the command line UX of `gen-driftignore` smoother. It privileges the use case of piping `driftctl scan` in and generating a `.driftignore` file, while still allowing other files to be used.

The first commit adds a default to the input file, so is additive. The second commit introduces a breaking change (writing to `./.driftignore` instead of stdout). I can seen how this might be controversial, and am happy to remove this change if the break isn't considered worth it.

### changelog

**gen-driftignore defaults to reading from stdin**

Allow the input file to be specified as "-", which traditionally many
programs interpret as standard input. Make this value the default too,
as piping `driftctl scan | driftctl gen-driftignore` appears to be a
documented idiom.

**gen-driftignore defaults to writing to a file**

Users can still print to stdout by passing `-o -`.


pkg/cmd/gen_driftignore.go
```diff
@@ -16,19 +17,34 @@ func NewGenDriftIgnoreCmd() *cobra.Command {
...
+			_, list, err := genDriftIgnore(opts, driftFile)
 			if err != nil {
 				return err
 			}
 
-			fmt.Println(list)
+			ignoreFile := os.Stdout
+			if opts.OutputPath != "-" {
+				var err error
+				ignoreFile, err = os.OpenFile(opts.OutputPath, os.O_CREATE|os.O_TRUNC|os.O_WRONLY, 0644)
```
craigfurman, line 41: Perhaps it's better to _not_ truncate and overwrite existing files, and use `O_EXCL` instead of `O_TRUNC`

craigfurman: Yeah now that we talk about it, I think EXCL is the safest. Make the user explicitly remove the file if they want to overwrite it. I'm not sure why we'd use APPEND?

craigfurman: Ah wait, I think I might see why APPEND could work: if you already have a `.driftignore`, the new drift won't contain the already-ignored resources... is that what you meant?

This is more complicated than I'd thought 😁

moadibfr: > I'm not sure why we'd use APPEND?

You could run it again to update your .driftignore without re running a full scan. From what I know @sjourdan used to >> into the file (and filled a bug because there was no \n at the end of the generated file). But yeah now we talk about this I'm not sure what use case it would fit so exclude would be safe.

craigfurman: That makes sense... I'll defer to your expertise here, I've barely used driftctl yet. So far my thoughts are:

* append would definitely satisfy some use-cases, but I haven't fully considered whether it'd be a footgun for other use-cases.
* excl would force the user to perform a full rescan with no ignore rules (unless they pass `--driftignore` to work around this), which sort of feels like the most conservative approach.
* truncate is probably wrong... we'd be implicitly removing the existing driftignore rules!

eliecharra: I think truncate should be dropped unless we introduce a previous file backup mechanism. If we trunc then we should spawn a `.driftignore.bak` file.

Anyway I think the debate here is to decide if we want to keep the append behavior or not, with the UX implication @craigfurman greatly summarized above ☝🏻 🙏🏻 

>  - append would definitely satisfy some use-cases, but I haven't fully considered whether it'd be a footgun for other use-cases.
>  - excl would force the user to perform a full rescan with no ignore rules (unless they pass --driftignore to work around this), which sort of feels like the most conservative approach.
we could even support the append case by adding an extra optional flag --append

I think the append way is more convenient and more user friendly, let me elaborate.

driftctl is most of the time used in an iterative way. You run, check your result, fix / ignore, and repeat.
I think it fit better with this usage, I mean the UX will be better since you do not need to invalidate your previous ignore policy. In term of performance this is probably also a better move since driftignore rules are taken into account when deciding to make API calls or not (this is not the case with `--filter`). It work fine for a first run too.

I'm seeing only one drawback of this approach. For example if we take the case when you decided to have an inverted ignore file logic, like this:

```
*
!aws_s3_bucket
```

You may see unmanaged bucket in your output after a scan, pipe this to the gen-driftignore cmd and you'll end up with this kind of file:

```
*
!aws_s3_bucket
aws_s3_bucket.my-unmanaged-bucket
```

I don't know how we handle this but I think that the `!aws_s3_bucket` rule is taking the precedence over this one `aws_s3_bucket.my-unmanaged-bucket`. So we need people to reorder their logic, but I'm not seeing a simple way to handle that.

pkg/cmd/gen_driftignore.go
```diff
@@ -16,19 +17,34 @@ func NewGenDriftIgnoreCmd() *cobra.Command {
...
+				if err != nil {
+					return errors.Wrapf(err, "can't open file %s", opts.OutputPath)
```
eliecharra, line 43: We are hiding the root cause of why the file can't be create here, maybe we should just keep the original error since `os.Openfile` errors (`open /that/does/not/exist: no such file or directory
`) are explicit enough WDYT ?

pkg/cmd/gen_driftignore.go
```diff
@@ -16,19 +17,34 @@ func NewGenDriftIgnoreCmd() *cobra.Command {
...
-			if opts.InputPath == "" {
-				return errors.New("Error: you must specify an input to parse JSON from. Use driftctl gen-driftignore -i <drifts.json>\nGenerate a JSON file using the output flag: driftctl scan -o json://path/to/drifts.json")
+			driftFile := os.Stdin
+			if opts.InputPath != "-" {
+				var err error
+				driftFile, err = os.Open(opts.InputPath)
+				if err != nil {
+					return err
+				}
+				defer driftFile.Close()
```
eliecharra, line 30: Why not putting this logic directly in `genDriftIgnore` ? I do not see any reason to keep it here.

For me the `RunE` from a command should be viewed as Controller in an MVP approach, and we want to keep less business as possible here. It will be also easier to unit test `genDriftIgnore`, I know that we are testing the whole cmd run but I like to keep things separated. This logic for me is a part of the business logic and the responsibility of `genDriftIgnore`

I'm wondering if we should move the output logic within too 🤔 For me this is fine to have output logic and formatting stuff in a controller but I hesitate. WDYT @craigfurman ?

pkg/cmd/gen_driftignore_test.go
```diff
@@ -120,7 +103,7 @@ func TestGenDriftIgnoreCmd_ValidFlags(t *testing.T) {
 		{args: []string{"gen-driftignore", "--exclude-missing"}},
 		{args: []string{"gen-driftignore", "--exclude-changed"}},
 		{args: []string{"gen-driftignore", "--exclude-changed=false", "--exclude-missing=false", "--exclude-unmanaged=true"}},
-		{args: []string{"gen-driftignore", "--input", "/dev/stdin"}},
+		{args: []string{"gen-driftignore", "--input", "-"}},
```
eliecharra, line 106: I think it will be more valuable if we replace this with `{args: []string{"gen-driftignore"}},` instead. We'd better want to cover the default value behavior instead if enforcing it explicitly.

pkg/cmd/gen_driftignore_test.go
```diff
@@ -55,53 +54,37 @@ func TestGenDriftIgnoreCmd_Input(t *testing.T) {
 			output: "./testdata/output_stdin_valid_filter2.txt",
 			err:    errors.New("open doesnotexist: no such file or directory"),
 		},
-		{
-			name:   "test error when input flag is not specified",
-			args:   []string{},
-			output: "",
-			err:    errors.New("Error: you must specify an input to parse JSON from. Use driftctl gen-driftignore -i <drifts.json>\nGenerate a JSON file using the output flag: driftctl scan -o json://path/to/drifts.json"),
-		},
```
eliecharra, line 63: This test cannot be simply dropped, we should now ensure that it will not trigger any error if we do not pass any args.

eliecharra (changes requested):
Thanks @craigfurman for this awesome PR 🎉 

I think that actually your feedback and first impression on the tool UX is very relevant for us as a newcomer. Really appreciate that 🙏🏻

pkg/cmd/gen_driftignore.go
craigfurman: Thanks for that @eliecharra, I think I understand now, and agree with you. WDYT of using append, _and_ backing up the output file (`.driftignore` by default) to `${NAME}.bak` (unless the file is stdin), just to be on the safe side?

Even if we default to append, there are a couple of potential rough edges remaining, including the one you call out at the end of your last comment. At the moment, section comments like `# Resources not covered by IaC` are duplicated when we append, but this is also the case when using shell `>>`, or manually merging files. I reckon changing things like that are probably out of scope of this particular PR, WDYT?

Finally, if we're writing to a file, we could print a message to stderr informing the user that we did this, in case they get confused by the silence. I'm not sure whether or not I like that idea in general, but since we're proposing to change the default behaviour, it might make sense. Of course totally up to you.

pkg/cmd/gen_driftignore.go
craigfurman: I agree with the analogy to controllers in web-type-things, and that in general you don't want to mix a ton of logic in there, for sanity's sake 😅 

This is quite subjective, but the yardstick I normally use for what to put "at the edge", in handlers for CLI / web adapters, and what to put deeper in the core, is: map inputs and outputs to whatever technical domain\* we're targeting in the "controller", and put the rest in the core, as long as that core function can be written with a "pure go" signature that has no / very-little dependency on the technical details of how the program does I/O.

But, having said all that - now that I re-look at the code, `genDriftIgnore` is just a function in the same file! I'd be totally happy to move this part.

\*: by "technical domain", I mean things like parsing flags and writing to files for a CLI like driftctl, or parsing HTTP requests and serialising responses for a web handler.

pkg/cmd/gen_driftignore.go
eliecharra: > WDYT of using append, and backing up the output file (.driftignore by default) to ${NAME}.bak (unless the file is stdin), just to be on the safe side?

For me this is as minor safety feature since there is no real risk to blow existing rules. Maybe you can open an issue for it (it could be a good first issue since it is a simple nice to have feat) 😃 

> Even if we default to append, there are a couple of potential rough edges remaining, including the one you call out at the end of your last comment. At the moment, section comments like # Resources not covered by IaC are duplicated when we append, but this is also the case when using shell >>, or manually merging files. I reckon changing things like that are probably out of scope of this particular PR, WDYT?

Maybe we could add a header like `# Generated by gen-driftignore cmd @05/11/2021 10:12` while appending results ? So we could have multiples sequential ignore blocks with `# Resources not covered by IaC` acting as sub-section. In a iterative process it could be useful for used to be able to get the history WDYT ?

> Finally, if we're writing to a file, we could print a message to stderr informing the user that we did this, in case they get confused by the silence. I'm not sure whether or not I like that idea in general, but since we're proposing to change the default behaviour, it might make sense. Of course totally up to you.

If we are writing to a file, why not using stdout instead of stderr ?

pkg/cmd/gen_driftignore_test.go
craigfurman: Good idea, if we cover that default and explicit `-`, that'll satisfy this note and the one below right?

pkg/cmd/gen_driftignore.go
craigfurman: > For me this is as minor safety feature since there is no real risk to blow existing rules. Maybe you can open an issue for it

Sure, so just to be clear you want to leave that out of this PR? I'm fine with that 🙂 

> Maybe we could add a header like # Generated by gen-driftignore cmd @05/11/2021 10:12

That's a great idea, and as you say a good first step towards maybe doing some auto-merging in the future, while not confusing users today.

> If we are writing to a file, why not using stdout instead of stderr ?

Sure, if we're writing a file then we could get away with printing diagnostic messages to stdout, but traditionally that's what stderr is for. I think it confers a minor advantage: if a user _does_ try to pipe the output into something that parses it, or redirects it to a file, e.g. because they're expecting the old behaviour of driftctl, they'll still see that diagnostic message on stderr (as long as they don't pipe that too). A minor point for sure, but I don't think we _lose_ anything by using stderr so it makes sense to me to take the very small win! WDYT?

pkg/cmd/gen_driftignore_test.go
craigfurman: Sorry, I read too quickly, the note I linked isn't as related to this thread as I thought.

If we want to test the default-stdin case, we'll have to do some ceremony around replacing stdin during the test. I don't think that would assist design much, or do much to prevent regressions: it seems like all it would really do is ensure golang stdlib's `os.Stdin` does what it says it does.

Rephrased in code comments:

```go

// This line is covered by tests, but this value is never used.
// This compiles, because `os.Stdin` is a `*os.File`.
// We do not test that os.Stdin behaves like a file, but surely not every program that
// uses go's stdlib needs to re-validate it completely?
driftFile := os.Stdin

// The contents of this branch are covered by tests
if opts.InputPath != "-" {
	var err error
	driftFile, err = os.Open(opts.InputPath)
	if err != nil {
		return err
	}
	defer driftFile.Close()
}
```

This PR already removes similar test setup around pipes and stdout, instead using temporary files passed to `-o` to avoid the need for it.

eliecharra: I think we should keep this test since this will allow us to assert on the results by the same time. Maybe we can drop this case  `{args: []string{"gen-driftignore", "--input", "-"}}` in the other test and keep this one. The other test is only validating that we are not getting any error, but obviously `TestGenDriftIgnoreCmd_Input` supersed it.

We should have to update a little bit this test to be able to send a proper input to stdin

pkg/cmd/gen_driftignore.go
eliecharra: > But, having said all that - now that I re-look at the code, genDriftIgnore is just a function in the same file! I'd be totally happy to move this part.

Oh you are right so no big deal here 😅 I've made expectations on the partial diff output I was reading without looking deeper 🙏🏻

pkg/cmd/gen_driftignore.go
eliecharra: > Sure, so just to be clear you want to leave that out of this PR? I'm fine with that slightly_smiling_face

Yeah ! We tried to leave these kind of stuff to the community. We wanted to encourage contributions and BTW it's a win-win for everyone 😃 

> That's a great idea, and as you say a good first step towards maybe doing some auto-merging in the future, while not confusing users today.

Yeah exactly auto merging could be the ultimate level of this command, we'll be able to handle inverted logic by prepending some rules instead of appending. This one will probably need a `.bak` file BTW 😛 

> Sure, if we're writing a file then we could get away with printing diagnostic messages to stdout, but traditionally that's what stderr is for. I think it confers a minor advantage: if a user does try to pipe the output into something that parses it, or redirects it to a file, e.g. because they're expecting the old behaviour of driftctl, they'll still see that diagnostic message on stderr (as long as they don't pipe that too). A minor point for sure, but I don't think we lose anything by using stderr so it makes sense to me to take the very small win! WDYT?

I think you have a better experience than me on that so I leave that up to you, I'm fine to use both. 

More generally about BC break: for now the project is in a really early beta stage
We do not have shipped a `1.x.x` so we can totally assume BC break in minor versions. Anyway there is no reason to introduce some if we can avoid OFC 😃

craigfurman:
I've implemented a few review suggestions, replied on other threads. I know we're not at merge time yet, but hopefully when we are, let me squash the review stuff in before merge

pkg/cmd/gen_driftignore_test.go
eliecharra: For sure we do not care about validating `os.Stdin` behavior here.
The point is more about to validate that when we do not specify anything we try to read from `os.Stdin`, replacing it during the test will cover that.

The problem I have here is that if we change the default `driftFile := os.Stdin` we are blind in test suite about this, and since it could be a potential big BC break I just wanted to cover that.

Maybe you have a better approach for that ?

craigfurman: That line is arguably more like config than code. It contains no logic. If we were to test it by swapping out `os.Stdin`, the test isn't really proving much: we're basically just asserting that the symbols on each side of the test-prod boundary match, if you see what I mean.

I'm not saying there's no value in such a test, and am also not saying that _not_ covering the stdin case is risk-free: what I am saying is that subjectively, I suspect that such a test is on the wrong side of the cost/benefit calculus, _given the way the test harness is already set up_. I wouldn't be making this point if we were talking about a more "integration test" style harness, against the compiled binary, that we could pipe _real_ stdin into (not just stubbing out `os.Stdin` in go). As far as I know driftctl doesn't have such a harness?

I don't want to be too stubborn though, this is my first contribution to this project! I've made my case, let me know what you decide.

pkg/cmd/gen_driftignore_test.go
```diff
@@ -158,3 +143,10 @@ func TestGenDriftIgnoreCmd_InvalidFlags(t *testing.T) {
 		assert.EqualError(t, err, tt.err.Error())
 	}
 }
+
+// The leading comment, "Generated by gen-driftignore..." contains a timestamp,
```
eliecharra, line 147: I'm good with that for now, but this something I consider hacky.

I consider time like any other dependency in a domain codebase, that should have an interface and be mock-able.
Since the gen-driftignore does not have a clear domain service code let's keep that for the moment. I just wanted to add my 50c here 😅

craigfurman: Sure, injecting a fake clock is something I've done a lot before for similar problems. I'm definitely on board with mock-driven TDD. But, it's a question of how much we want to go about refactoring this part of driftctl for this individual PR. Maybe this can almost be considered technical debt? Up to you

eliecharra: > But, it's a question of how much we want to go about refactoring this part of driftctl for this individual PR. Maybe this can almost be considered technical debt? Up to you

Of course it is tech debt and it has nothing to do with this PR 🙂 Sorry if that wasn't clear enough in my comment.

craigfurman: it was, I think I was just being overly explicit 😅

pkg/cmd/gen_driftignore_test.go
eliecharra: > we're basically just asserting that the symbols on each side of the test-prod boundary match, if you see what I mean.

Yeah totally.

I agree with your point about the value in a such test.
Maybe here we do not want an unit test but something more close to an acceptance test or E2E testing.

> I wouldn't be making this point if we were talking about a more "integration test" style harness, against the compiled binary, that we could pipe real stdin into (not just stubbing out os.Stdin in go). As far as I know driftctl doesn't have such a harness?

Ok ok my bad I read that comment too fast and that was exactly my feelings too in my previous sentence 😅  Actually we do have some integration testing (we call that acceptance testing but whatever). We basically run driftctl cmd in a wrapper that is able to expose an assert API, you can have an example [here](https://github.com/cloudskiff/driftctl/blob/main/pkg/resource/google/google_bigtable_table_test.go#L11)

I want to cover this default behavior case from the end user point of view, so you are right this does not make sense to cover this in a kinda unit scoped test 👍🏻 

Actually we do not have any mechanism to be able to replace the `os.Stdin` in out "micro-acceptance-test-framework" but this is something we can probably do without any trouble. We are capturing driftctl output during the run to be able to assert on it (btw we do not have any acceptance test case that is asserting on stdout as far as I know 😅 ). 

Soo maybe this is something we can achieve in another PR, but I don't think that the fix will be too heavy so as you like. Since it will do not put main branch in an inconsistent state if we split this in another PR, I'd rather prefer that. I'm I'm not teaching you anything but big PR can be really painfull to review

eliecharra (approved):

eliecharra:
@all-contributors please add @craigfurman to code

eliecharra:
@craigfurman Yeah sorry I didn't explained the doc process 🙏🏻 The idea is that when we merge a feature here, we push a doc PR on the doc repo. When the PR gets merged in the doc repo, your changes will be available for "next" version :  https://docs.driftctl.com/next

When we want to release, we just create a new version of the doc based on the main branch, so your changes will be taken into account for the next version. 

Please let me know if it's not clear enough 🙏🏻

craigfurman:
Thanks! I've made a docs PR here: https://github.com/cloudskiff/driftctl-docs/pull/180

Issue #1211, opened 2021-11-10
Progressbar is never showing


moadibfr:
**Description**
progressbar is supposed to show unless user add --quiet or the output is in interpretable format on stdout

**How to reproduce**
Running a scan will never show any progressbar

**Possible Solution**
when we added multiple output capabilities `output.GetOutput(o, opts.Quiet)` was moved to the end of the scan in scan.go. This function is used to set the global printer so if it's not called at the beginning we wont show anything on stdout as the default printer is voidPrinter.
We have to moved the global printer setting to the begining on command and iterate all output tu check which one to use.

PR #1212, opened 2021-11-10
Display progress bar in stdout

sundowndev:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | yes
| 🚀 New feature?   | no
| ⚠ Deprecations?   | no
| ❌ BC Break       | no
| 🔗 Related issues | #1211
| ❓ Documentation  | no 

## Description

As reported by @moadibfr, the progress bar and other global output usage were hidden because as we implemented multiple output (#865), the global output was always assigned to the `voidPrinter` which prints nothing. This PR fixes that by setting the global output **before** the scan.

### DIFF
--- a/pkg/cmd/scan.go
+++ b/pkg/cmd/scan.go
@@ -216,6 +216,12 @@ func scanRun(opts *pkg.ScanOptions) error {
 
 	alerter := alerter.NewAlerter()
 
+	// For now, we only use the global printer to print progress and information about the current scan, so unless one
+	// of the configured output should silence global output we simply use console by default.
+	if output.ShouldPrint(opts.Output, opts.Quiet) {
+		globaloutput.ChangePrinter(globaloutput.NewConsolePrinter())
+	}
+
 	providerLibrary := terraform.NewProviderLibrary()
 	remoteLibrary := common.NewRemoteLibrary()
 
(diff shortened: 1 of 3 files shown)


pkg/cmd/scan.go
```diff
@@ -216,6 +216,9 @@ func scanRun(opts *pkg.ScanOptions) error {
 
 	alerter := alerter.NewAlerter()
 
+	// For now, we only use the global printer to print progress and information about the current scan, so we simply use stdout by default.
+	globaloutput.ChangePrinter(output.GetPrinter(output.OutputConfig{Key: output.ConsoleOutputType}, opts.Quiet))
```
moadibfr, line 220: You would have to iterate on the real resources to see if that conflict. If one of the output is json or plan in stdout we don't want to show the progesssbar as it will break the output.

Also remember that this is not only for progressbar we have some other print that use globalprinter and should not be displayed in these cases

moadibfr (changes requested):

pkg/cmd/scan.go
sundowndev: Yeah I should've think about that, just updated the changes

pkg/cmd/scan/output/output.go
```diff
@@ -64,6 +62,16 @@ func GetOutput(config OutputConfig, quiet bool) Output {
 	}
 }
 
+func ShouldPrint(outputs []OutputConfig, quiet bool) bool {
```
eliecharra, line 65: It can be usefull to add context as comment for this function, maybe we can rename it to `isOutputAllowed` or something like that

eliecharra (changes requested):

pkg/cmd/scan/output/output.go
sundowndev: I added a comment, please tell me if it makes it clearer, I can also rename it.

eliecharra (approved):

Issue #1063, opened 2021-09-14
Support Azure Private DNS Resources  (TF)


sjourdan:
**Description**

All Azure Private DNS Zone-related resources are grouped here for simplicity. 

Add TF Azure support for 

- [x] `azurerm_private_dns_zone`
- [x] `azurerm_private_dns_a_record`
- [x] `azurerm_private_dns_aaaa_record`
- [x] `azurerm_private_dns_cname_record`
- [x] `azurerm_private_dns_mx_record`
- [x] `azurerm_private_dns_ptr_record`
- [x] `azurerm_private_dns_srv_record`
- [x] `azurerm_private_dns_txt_record`

**Sources**

* [Terraform Resource Documentation](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/private_dns_zone)

PR #1213, opened 2021-11-10
add azurerm private dns cname record

moadibfr:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | no
| 🚀 New feature?   | yes
| ⚠ Deprecations?   | no
| ❌ BC Break       | no
| 🔗 Related issues | #1063
| ❓ Documentation  | yes 

## Description
add azurerm private dns cname record


pkg/remote/azurerm/repository/privatedns.go
```diff
@@ -145,6 +146,30 @@ func (s *privateDNSRepository) ListAllAAAARecords(zone *armprivatedns.PrivateZon
...
+
+	records, err := s.listAllRecords(zone)
+	if err != nil {
+		return nil, err
+	}
+	results := make([]*armprivatedns.RecordSet, 0)
+	for _, record := range records {
+		if record.Properties.CnameRecord == nil {
+			continue
+		}
```
eliecharra, line 163: This is not a efficient usage of the cache, we are calling the `listAllRecords` one time per record type with this approach.

I think this logic of record type check should be moved in the provider. In the repository we should have only a method that return all the records for a zone IMHO, so we can cache all results.

eliecharra (changes requested):

pkg/remote/azurerm/repository/privatedns.go
moadibfr: I removed cache for the filtering and leaved it only for the request for all record.
I like having the ListAll<Type> in the repository as we could as well do filtering in the request and so this make the abstraction more transparent.

eliecharra: Yeah sorry, seeing caching logic here put me on the wrong direction 🙏🏻 Thanks to have removed it, it does not add a real performance improvement since it may be most costly to acquire a lock on the cache map than filtering at some point, thanks to have removed it 👍🏻 

> I like having the ListAll{Type} in the repository as we could as well do filtering in the request and so this make the abstraction more transparent.

I agree with that 👍🏻

eliecharra (approved):

Issue #1061, opened 2021-09-14
Support Azure azurerm_lb_rule (TF)


sjourdan:
**Description**

Add Azure support for `azurerm_lb_rule` (TF)

**Sources**

* [Terraform Resource Documentation](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/lb_rule)

PR #1221, opened 2021-11-17
Add azurerm_lb_rule resource

sundowndev:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | no
| 🚀 New feature?   | yes
| ⚠ Deprecations?   | no
| ❌ BC Break       | no
| 🔗 Related issues | #1061 
| ❓ Documentation  | https://github.com/cloudskiff/driftctl-docs/pull/184 

**Note to reviewers**:

- I renamed `loadbalancersClient` to `loadBalancersClient` in the network repository to follow a common naming convention with other properties, so you may notice a small amount of unrelated diffs
- Be aware that this resource has deep mode enabled
- A change has been made to the way we compute filenames for golden files, but enabled only for filenames longer than 255 bytes to avoid backward compatibility issues (I added a comment with more context)


pkg/resource/azurerm/testdata/acc/azurerm_lb_rule/.driftignore
```diff
@@ -0,0 +1,3 @@
+*
+!azurerm_lb
```
moadibfr, line 2: do we really need to list lb here ? Why ?

moadibfr (changes requested):

pkg/resource/azurerm/testdata/acc/azurerm_lb_rule/.driftignore
sundowndev: No, it's not mandatory, I can delete it

eliecharra (approved):

Issue #1021, opened 2021-09-07
Support terraform state backend on GCS


sjourdan:
**Description**

Terraform users on GCP may use GCS (cloud storage) as a storage backend.
Driftctl users may want to easily access those states to scan for unmanaged resources - as S3 users do.

cf: https://www.terraform.io/docs/language/settings/backends/gcs.html

**Example**

S3 users: 

```sh
$ driftctl scan --from tfstate+s3://acme-tfstates/eu-west-3.tfstate
```

GCS users: 

```sh
$ driftctl scan --from tfstate+gcs://acme-tfstates/bigtable.tfstate
```

Users wanting this: 

- B., a DevOps in a consulting company in London. ~200 accounts (2021/12/8)
- "trepx" on Discord (2021/12/9)

PR #1232, opened 2021-11-23
Implement GCS backend

sundowndev:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | no
| 🚀 New feature?   | yes
| ⚠ Deprecations?   | no
| ❌ BC Break       | no
| 🔗 Related issues | #1021 
| ❓ Documentation  | https://github.com/cloudskiff/driftctl-docs/pull/188 

## Description

This PR implements a new backend called GS (Google Storage).

**Notes to reviewers**:
- I choose the alias `gs://` over `gcs://` to be consistent with the gsutil URI (`gs://path/to/file`) shown in GCP console, let me know WDYT about it
- Testing is slightly different than Asset since the Storage API uses HTTP, not gRPC


test/google/storage.go
```diff
@@ -0,0 +1,57 @@
...
+	}()
+
+	// This prevents the client to use HTTPS
+	_ = os.Setenv("STORAGE_EMULATOR_HOST", fakeServerAddr)
+
+	// Create a client.
+	client, err := storage.NewClient(ctx,
+		option.WithoutAuthentication(),
+		option.WithGRPCDialOption(grpc.WithInsecure()),
+	)
```
eliecharra, line 52: Can you have a look here https://github.com/googleapis/google-api-go-client/blob/main/testing.md#testing-http-services-using-fakes.

I do not think that we need `option.WithGRPCDialOption` or `STORAGE_EMULATOR_HOST` things here.

eliecharra (changes requested):
I think we can tune the way we test things to be more straightforward.

pkg/iac/terraform/state/backend/backend.go
```diff
@@ -53,6 +56,12 @@ func GetBackend(config config.SupplierConfig, opts *Options) (Backend, error) {
 		return NewHTTPReader(&http.Client{}, fmt.Sprintf("%s://%s", config.Backend, config.Path), opts)
 	case BackendKeyTFCloud:
 		return NewTFCloudReader(&http.Client{}, config.Path, opts)
+	case BackendKeyGS:
+		client, err := storage.NewClient(context.Background())
+		if err != nil {
+			return nil, err
+		}
```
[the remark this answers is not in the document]
sundowndev: It was for testing purposes, like we do for the HTTP backend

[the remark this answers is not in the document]
eliecharra: I don't think that we need to do that, can you try without I think we can achieve a proper testing without the need to use a constructor

pkg/iac/terraform/state/backend/gs_reader.go
```diff
@@ -0,0 +1,61 @@
...
+		storageClient: storageClient,
+	}, nil
+}
+
+func (s *GSBackend) Read(p []byte) (int, error) {
+	if s.reader == nil {
+		ctx := context.Background()
+		rc, err := s.storageClient.Bucket(s.bucketName).Object(s.path).NewReader(ctx)
+		if err != nil {
+			return 0, errors.WithMessage(err, s.path)
```
moadibfr, line 46: Shouldn't the error message be a little more explicit than just the path ? Or maybe you can just let the original error go ?

eliecharra: I agree, if we forward the original error we got this 
image

I think this is clear enough WDYT ?

moadibfr: :+1: the path is already displayed so I don't think we need to add anything

test/google/storage.go
eliecharra: FYI @sundowndev I removed all options passer to `NewClient`.
We only need to set a proper fake host in `STORAGE_EMULATOR_HOST`

The storage client is very inconsistent, it remain in the gRPC namespace of GCP but under the hoods it uses an http client 🤔

eliecharra (approved):

eliecharra (approved):

eliecharra (approved):

moadibfr (approved):

PR #1236, opened 2021-11-25
Fix hacky way to serve fake TF provider schema in tests

eliecharra:
When you run `go test github.com/cloudskiff/driftctl`,  `runtime.Caller(0)` contains the package name and it makes our relative file read to fail. 

By using embed we can provide a better way to read mock schema during test.


docs/new-remote-provider.md
```diff
@@ -179,6 +179,6 @@ Then, add a call to this function in the `remote/<provider>/init.go` file you cr
 
 You also need to create a test schema for upcoming tests.
 
-Please use `TestCreateNewSchema` located in `test/schemas/schemas_test.go` to generate a schema file that will be used for the mocked provider.
+Please use `TestCreateNewSchema` located in `test/terraform/schemas_test.go` to generate a schema file that will be used for the mocked provider.
```
sundowndev, line 182: Just ouf of curiosity, what's the reason behind this renaming ?

sundowndev (approved):

PR #1237, opened 2021-11-25
Make version check test more dynamic

eliecharra:
In some particular context (when packaging driftctl in release mode, check [nixpkg](https://github.com/NixOS/nixpkgs/commit/ded247ada384ceacc5d0d186cffa7c1cc927d268#diff-f6248b889160cf7652be9255c593b1ded04177a6d33449c69b8f4e4abf28ce57R45-R47)) theses test are failing since the version `dev-dev`is hardcoded.
This PR make sure that they will match the current version regardless of any ldflags.

### DIFF
--- a/pkg/cmd/driftctl_test.go
+++ b/pkg/cmd/driftctl_test.go
@@ -41,23 +41,6 @@ func TestDriftctlCmd_Help(t *testing.T) {
 	}
 }
 
-func TestDriftctlCmd_Version(t *testing.T) {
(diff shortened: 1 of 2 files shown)


sundowndev (commented):
I don't understand the difference between the two tests you've modified, should one be removed?

eliecharra:
> I don't understand the difference between the two tests you've modified, should one be removed?

You are right, I removed one of them 👍🏻

sundowndev (approved):

PR #1238, opened 2021-11-25
Add a global switch to disable third party

eliecharra:
## Description

This PR is adding a more convenient way to disable any third-party network calls.

You can now build an telemetry-free version of driftctl easily with 
```shell
$ go build -ldflags="-X 'github.com/cloudskiff/driftctl/build.enableThirdPartyNetworkCalls=false'" 
```

## References
- https://github.com/NixOS/nixpkgs/commit/ded247ada384ceacc5d0d186cffa7c1cc927d268#diff-f6248b889160cf7652be9255c593b1ded04177a6d33449c69b8f4e4abf28ce57R31-R38

### DIFF
--- a/build/build.go
+++ b/build/build.go
@@ -2,12 +2,23 @@ package build
 
 var env = "dev"
 
+// This flag could be switched to false while building to create a binary without third party network calls
+// That mean that following services will be disabled:
+// - telemetry
+// - version check
+var enableUsageReporting = "true"
+
 type BuildInterface interface {
 	IsRelease() bool
+	IsUsageReportingEnabled() bool
 }
 
 type Build struct{}
 
 func (b Build) IsRelease() bool {
 	return env == "release"
+}
+
+func (b Build) IsUsageReportingEnabled() bool {
+	return enableUsageReporting == "true"
 }
--- a/pkg/telemetry/telemetry.go
+++ b/pkg/telemetry/telemetry.go
@@ -6,6 +6,7 @@ import (
 	"net/http"
 	"runtime"
 
+	"github.com/cloudskiff/driftctl/build"
 	"github.com/cloudskiff/driftctl/pkg/memstore"
 	"github.com/cloudskiff/driftctl/pkg/version"
 	"github.com/sirupsen/logrus"
@@ -21,7 +22,21 @@ type telemetry struct {
 	ProviderName   string `json:"provider_name"`
 }
 
-func SendTelemetry(store memstore.Bucket) {
+type Telemetry struct {
+	build build.BuildInterface
+}
+
+func NewTelemetry(build build.BuildInterface) *Telemetry {
+	return &Telemetry{build: build}
+}
+
+func (te Telemetry) SendTelemetry(store memstore.Bucket) {
+
+	if !te.build.IsUsageReportingEnabled() {
+		logrus.Debug("Usage reporting is disabled on this build, telemetry skipped")
+		return
+	}
+
 	t := &telemetry{
 		Version: version.Current(),
 		Os:      runtime.GOOS,
--- a/main.go
+++ b/main.go
@@ -31,8 +31,14 @@ func run() int {
 
 	config.Init()
 	logger.Init()
+	build := build.Build{}
+	logrus.WithFields(logrus.Fields{
+		"isRelease":               fmt.Sprintf("%t", build.IsRelease()),
+		"isUsageReportingEnabled": fmt.Sprintf("%t", build.IsUsageReportingEnabled()),
+		"version":                 version.Current(),
+	}).Debug("Build info")
 
-	driftctlCmd := cmd.NewDriftctlCmd(build.Build{})
+	driftctlCmd := cmd.NewDriftctlCmd(build)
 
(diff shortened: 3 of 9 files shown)


build/build.go
```diff
@@ -2,12 +2,23 @@ package build
 
 var env = "dev"
 
+// This flag could be switched to false while building to create a binary without third party network calls
+// That mean that following services will be disabled:
+// - telemetry
+// - version check
+var enableThirdPartyNetworkCalls = "true"
```
eliecharra, line 9: Please suggest me another naming, I'm not really convinced by that 😅

craigfurman: The awkward name might be suggesting that we've combined separate concepts into this one flag, as the comment explains. Use-case-wise, are disabling version checks and telemetry really the same thing?

Not that it's really a strong opinion of mine to separate this into 2 build flags, perhaps it's not worth it from a complexity PoV.

If we do want to keep them the same, how about: `enableTelemetry`, to lead with the thing users are more likely to care about disabling.

eliecharra: > The awkward name might be suggesting that we've combined separate concepts into this one flag

exactly 😅 

> Use-case-wise, are disabling version checks and telemetry really the same thing?

Some people reached us about that a couple of month ago, even if telemetry and version check can be opted-out, they wanted to bring a binary in their ecosystem free of that kind of code, so they forked and removed every external calls. 

Today it seems that nix pkg maintainers want to opt-out theses things by default.

So yeah to try to answer your question, I do not think that telemetry and version check are the same think, but both are - by default - leaking data in the internet, and here I think this is our usecase : be able to build a driftctl version free of any data leak.

Since dctl can be seen as a security tool, this should be taken seriously and I think "one flag to rule them all" is a better approach since this ensure maintainers that the version will be free of data leak in the time.

For example if we decide for some reason one day to enable by default crash reporting, we should be able to disabled it from our perspective without the need for the maintainers to check if there is a new data leak feat in every new version. 

> If we do want to keep them the same, how about: enableTelemetry, to lead with the thing users are more likely to care about disabling.

Maybe it's a good idea to consider everything that leak data as "telemetry" related 👍🏻. I am biased since we separated theses two concept in dctl, but maybe for most of the users it should be clear enough ?

moadibfr: Maybe something like enableUsageReport ? It's not 100% accurate but still it would fit with what we actually do with telemetry and version check and be clear enough to people that would use that ?
Third party network call sounds really strange to me.

craigfurman: @eliecharra that all sounds reasonable. Some possible alternative names: `enableTelemetryAndCheckup`, `enableTelemetryAndDiagnostics`? I'm being slightly more generic over the "version check" aspect

eliecharra: > Third party network call sounds really strange to me.

100% agree, whatever we choose this should not be this one 😅 

WDYT of `privateMode` ? It make me think of web browser private mode, I get the idea during the sleep.

sundowndev: May be we should simply use two different build arguments for this ? I've looked at how it's done in other projects and VSCode uses `enableCrashReporter` and `enableTelemetry` to disable those 3rd party calls. Also the user will better understand what they're disabling exactly, instead of being confused by a generic name. WDYT?

pkg/telemetry/telemetry.go
```diff
@@ -21,7 +22,21 @@ type telemetry struct {
...
+}
+
+func NewTelemetry(build build.BuildInterface) *Telemetry {
+	return &Telemetry{build: build}
+}
+
+func (te Telemetry) SendTelemetry(store memstore.Bucket) {
+
+	if !te.build.IsThirdPartyNetworkCallsAllowed() {
+		logrus.Debug("Third party network call are disabled on this build, telemetry skipped")
```
wbeuil, line 36: `Third party network callS are ...`

wbeuil (changes requested):

build/build.go
eliecharra: I already answered to that question. BTW I do not think VSCode (or more generally Microsoft related products) are well designed about this, I would like to avoid these kind of things : https://vscodium.com/

wbeuil: I don't have a clear opinion on whether we should have 2 flags or 1 flag to disable/enable third-party calls. Both what Craig/Raph said to have different build args and what Elie said on "one flag to rule them all" kind of make sense to me.

wbeuil: But if we go the one flag road, I thought of `enableReporters` or both suggested names of Craig that are more transparent of what we disable/enable.

eliecharra: I switched it to `privateMode`, I think it fit well with what we want to achieve here.

craigfurman: Sorry to keep this thread going, and I'll defer to long-term maintainers on this, but: `privateMode` is perhaps a bit of a loaded term? It makes it sound like the alternative is a privacy violation, as opposed to helpful version checks and anonymous telemetry.

eliecharra: It can make sense yeah, what about `enableUsageReport` ?

craigfurman: Slight tweak: `enableUsageReporting`? I still like `enableTelemetry` 😁 

totally up to you though

PR #1240, opened 2021-11-25
Disable checks that cannot make PRs green

eliecharra:


.circleci/config.yml
```diff
@@ -201,33 +201,37 @@ workflows:
...
-          matrix:
-            parameters:
-              pattern:
-                - TestAcc_Aws
-                - TestAcc_Github_
-                - TestAcc_Google
-                - TestAcc_Azure_
-          context:
-            - driftctl-acc
+# Disabled since it fail for contributors because of the snyk token stored in a private context
```
sundowndev, line 204: Instead of disabling them, isn't there a way to skip them when the context cannot be shared ?

eliecharra: I was looking for something like that initially, but I didn't manage to find something. Feel free if you want to have a look, no emergency for that

sundowndev: Yeah too bad CircleCI didn't address this issue yet

sundowndev (approved):

Issue #959, opened 2021-08-27
Support aws_launch_configuration


sjourdan:
**Description**

`aws_launch_configuration` provides a resource to create a new launch configuration, used for autoscaling groups.

**Example**

TF doc: https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/launch_configuration

**Known Users** 

- `medelgover` on discord (2021/10/19) using the `tf_aws_bastion_s3_keys` module
  - also needs https://github.com/cloudskiff/driftctl/issues/925

PR #1242, opened 2021-11-26
Add aws_launch_configuration

sundowndev:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | no
| 🚀 New feature?   | yes
| ⚠ Deprecations?   | no
| ❌ BC Break       | no
| 🔗 Related issues | #959 
| ❓ Documentation  | https://github.com/cloudskiff/driftctl-docs/pull/190


wbeuil (changes requested):
Small comments on your PR.

In regards to the deep mode, by simply looking at the Terraform Read function in their source code I found and tested a special case where someone could add a user-data in 2 different ways:

```
resource "aws_launch_configuration" "foo" {
  name          = "foo"
  image_id      = "ami-0d718c3d715cec4a7"
  instance_type = "t3.micro"
  user_data_base64 = filebase64("script.sh")
}

resource "aws_launch_configuration" "bar" {
  name          = "bar"
  image_id      = "ami-0d718c3d715cec4a7"
  instance_type = "t3.micro"
  user_data = file("script.sh")
}
```

This will end up at the end of the analysis with false positives.

image

I also see that they fetch for root device (e.g. image below):

image

I didn't test to find false positives in that regard but the chances are there are.

Maybe we should release this resource in non deep mode.

sundowndev:
@wbeuil Good catch! 

> Maybe we should release this resource in non deep mode.

You're right, I though it made sense to enable deep mode for this resource because you usually want to keep the template fully sync with your state. According to your tests there are too many edge cases we don't to deal with, I'll disable it.

wbeuil (dismissed):
LGTM

wbeuil (approved):

PR #1245, opened 2021-11-29
Fix acceptance cli arguments

sundowndev:
In acceptance testing environment, usage report has been disabled, this means --disable-telemetry and --no-version-check flags are not supported. This PR removes it to prevent having this error : `unknown flag: --disable-telemetry`


eliecharra (approved):
Wow nice catch thanks 👍🏻

Issue #958, opened 2021-08-27
Support aws_launch_template


sjourdan:
**Description**

`aws_launch_template` provides an EC2 launch template resource. Can be used to create instances or auto scaling groups.

**Example**

TF doc: https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/launch_template

PR #1246, opened 2021-11-29
Add aws_launch_template

sundowndev:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | no
| 🚀 New feature?   | yes
| ⚠ Deprecations?   | no
| ❌ BC Break       | no
| 🔗 Related issues | #958 
| ❓ Documentation  | https://github.com/cloudskiff/driftctl-docs/pull/192


pkg/remote/aws/launch_template_enumerator.go
```diff
@@ -0,0 +1,65 @@
...
+		}
+		ltData := launchTemplateVersions[0].LaunchTemplateData
+
+		results = append(
+			results,
+			e.factory.CreateAbstractResource(
+				string(e.SupportedType()),
+				*tmpl.LaunchTemplateId,
+				map[string]interface{}{
+					"credit_specification": getCreditSpecification(ltData.CreditSpecification),
```
eliecharra, line 48: I'm not sure to understand why this is required here ?

sundowndev: I tested the resource again without it and indeed it is unused, I removed it

eliecharra: Yeah, this is not used by any middleware in non deep mode and not in `ReadResource` either this should be OK to drop it.

eliecharra (approved):

PR #1252, opened 2021-12-03
Add support for aws_apigatewayv2_route

craigfurman:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | no
| 🚀 New feature?   | yes
| ⚠ Deprecations?   | no
| ❌ BC Break       | no
| 🔗 Related issues | https://github.com/cloudskiff/driftctl/issues/1249
| ❓ Documentation  | https://github.com/snyk/driftctl-docs/pull/196

## Description

This is the first resource that I'm trying to add to driftctl, so I may have got various things wrong.

~~I've ticked "no" on the documentation question above, I'm not sure if that's right? Is there a list of resources we support that I should update too?~~

~~I haven't yet written middleware to avoid spurious drift in the case in which the terraform user defines an aws_apigatewayv2_api with an inline openAPI declaration, that will cause AWS to create routes. There is already middleware for this exact case in the v1 API gateway, so I can probably extend that to support v2. Is there value in merging this first, or will the possibility of spurious drift confuse users, in which case it might be best to add middleware here before merge? I'm happy either way, but figured I'd ask for feedback early.~~

Closes https://github.com/cloudskiff/driftctl/issues/1249


sundowndev:
> I've ticked "no" on the documentation question above, I'm not sure if that's right? Is there a list of resources we support that I should update too?

For each provider, we maintain [a list of supported resources](https://docs.driftctl.com/0.17.0/providers/aws/resources) and whether deep mode is available or not, in the documentation. Please make a PR on the relevant repository for that resource 🙏🏻  ([example](https://github.com/cloudskiff/driftctl-docs/pull/193)).

> Is there value in merging this first, or will the possibility of spurious drift confuse users, in which case it might be best to add middleware here before merge?

I think it's useless to add this resource without the relevant middleware because we don't want the user to have false positives anyway. It's better to spend more time on it rather than releasing an incomplete support of the resource IMO.

craigfurman:
Hey @sundowndev, thanks for the docs! I'll be sure to file an update after I add the middleware to this.

> I think it's useless to add this resource without the relevant middleware

Minor semantic point, but I disagree that it's _useless_: it's useful to users who don't use openAPI declarations, but yes now that I think again it arguably offers negative, not neutral experience to users who do. False positives are perhaps worse than no support at all!

This doesn't really matter because the result is the same: I'll take this back to draft and add middleware 🙂

pkg/remote/aws/apigatewayv2_route_enumerator.go
```diff
@@ -0,0 +1,50 @@
...
+}
+
+func (e *ApiGatewayV2RouteEnumerator) SupportedType() resource.ResourceType {
+	return aws.AwsApiGatewayV2RouteResourceType
+}
+
+func (e *ApiGatewayV2RouteEnumerator) Enumerate() ([]*resource.Resource, error) {
+	apis, err := e.repository.ListAllApis()
+	if err != nil {
+		return nil, remoteerror.NewResourceListingError(err, string(e.SupportedType()))
```
wbeuil, line 29: Here we should return an error that means that it is the listing of the Api Gateway V2 Apis that failed and not the listing of the Routes.

`return nil, remoteerror.NewResourceListingErrorWithType(err, string(e.SupportedType()), aws.AwsApiGatewayV2ApiResourceType)`

pkg/remote/aws_apigatewayv2_scanner_test.go
```diff
@@ -100,6 +100,87 @@ func TestApiGatewayV2Api(t *testing.T) {
...
+			},
+		},
+		{
+			test: "cannot list api gateway v2 apis",
+			mocks: func(repository *repository.MockApiGatewayV2Repository, alerter *mocks.AlerterInterface) {
+				repository.On("ListAllApis").Return(nil, dummyError)
+				alerter.On("SendAlert", resourceaws.AwsApiGatewayV2RouteResourceType, alerts.NewRemoteAccessDeniedAlert(common.RemoteAWSTerraform, remoteerr.NewResourceListingErrorWithType(dummyError, resourceaws.AwsApiGatewayV2ApiResourceType, resourceaws.AwsApiGatewayV2RouteResourceType), alerts.EnumerationPhase)).Return()
+			},
+			wantErr: remoteerr.NewResourceListingError(dummyError, resourceaws.AwsApiGatewayV2RouteResourceType),
+		},
```
wbeuil, line 144: You would need to change this test since the returned error will change. Can you add also one test that will return an error if you cannot list API gateway v2 routes

wbeuil (changes requested):
It misses the test inside the `pkg/resource/aws/metadata_test.go` file for non deep-mode flag.

craigfurman:
@wbeuil I think this is ready for review again. Although I'm not sure exactly what you mean by "It misses the test inside the pkg/resource/aws/metadata_test.go file for non deep-mode flag", so if it means something other than what was in the line notes in that same review, I haven't done that yet.

I've implemented the middleware, let me know what you think.

wbeuil (commented):
We probably should rename the middleware file to `pkg/middlewares/aws_api_gateway_api_expander_test.go` since it's not only rest API anymore.

craigfurman:
Thanks @wbeuil, I've made those changes, this should be ready for another review

wbeuil (approved):
Amazing work @craigfurman!

Issue #916, opened 2021-08-06
Do not retrieve details for unmanaged resources.


eliecharra:
**Description**
We do not need details for unmanaged resources in deep mode since we do not have anything to compare.
We can save a lot of time and network calls by avoiding to retrieve details for unmanaged resource.

The new workflow could be

- Enumerate resources from IaC
- Enumerate resources from remote
- Analysis : without diff/changes
- Read resource details only for managed ones
- Analysis : add diff/changes
- Summary

sundowndev:
Reopening since we decided to rollback this feature. Considering the impact of #1253 on new releases, further investigation must be done to ensure the new architecture does not break the way we handle resources and analysis.

CodaBool:
Scan times between only managed and all are the same.
There seems to be something off about the way this was implemented.

PR #1253, opened 2021-12-03
Avoid reading unmanaged resources

sundowndev:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | no
| 🚀 New feature?   | no
| ⚠ Deprecations?   | no
| ❌ BC Break       | no
| 🔗 Related issues | #916 
| ❓ Documentation  | no 

## Description

Currently, the scan fetch details of every resources, even unmanaged ones, which is useless since they don't belong to any kind of intention. This PR split the process in two part : enumeration and resource reading. I also modified the analyzer logic to able to compare attribute differences in a separate function. The resource reading function now takes an argument so we pass the resources for which we want to fetch details. Since we need middlewares to run before beginning the analysis and retrieve unmanaged resources, the resource reading process is now done after running all middlewares.

As expected, the performance improvement is huge since we avoid many API calls.

**Note to reviewers**:

- Keep in mind that this change shouldn't block implementation of a new IaC supplier
- Since this change imply that middlewares will always run in non-deep mode, we should ensure that every single middleware can work in non-deep mode


pkg/driftctl.go
```diff
@@ -149,19 +149,33 @@ func (d DriftCTL) Run() (*analyser.Analysis, error) {
 		}
 	}
 
-	analysis, err := d.analyzer.Analyze(remoteResources, resourcesFromState)
+	analysis := d.analyzer.CompareEnumeration(remoteResources, resourcesFromState)
 	if err != nil {
```
moadibfr, line 153: seems like you removed the error returning here

pkg/analyser/analyzer.go
```diff
@@ -42,18 +42,17 @@ type AnalyzerOptions struct {
 }
 
 type Analyzer struct {
-	alerter *alerter.Alerter
-	options AnalyzerOptions
-	filter  filter.Filter
+	analysis *Analysis
```
moadibfr, line 45: not sure why we need an analysis in here and method return pointer to potentially incomplete analysis

pkg/analyser/analyzer.go
```diff
@@ -42,18 +42,17 @@ type AnalyzerOptions struct {
...
 
 func NewAnalyzer(alerter *alerter.Alerter, options AnalyzerOptions, filter filter.Filter) *Analyzer {
-	return &Analyzer{alerter, options, filter}
+	return &Analyzer{NewAnalysis(options), alerter, options, filter}
 }
 
-func (a Analyzer) Analyze(remoteResources, resourcesFromState []*resource.Resource) (Analysis, error) {
-	analysis := Analysis{options: a.options}
-
+func (a Analyzer) CompareEnumeration(remoteResources, resourcesFromState []*resource.Resource) *Analysis {
```
moadibfr, line 55: Ok I find this new interface quite confusing and it took me some time understanding how that will work and why you needed this.

As it occurs you need 2 separate step to first get resource that need to be read and then read them if in deep mode, I thinks we could make the analyzer interface more comprehensive:
- I don't really like that the 2 calls return the pointer to an internal analyze I think you should create a new one at each step and init the second one with either the internal one or an argument to the second step ?
- It's strange that you need to call a method to get the analysis correct (analyzer.SetAlerts()) and that this work only because you returned a pointer to the internal analysis that will be updated after being returned.

pkg/driftctl.go
```diff
@@ -149,19 +149,33 @@ func (d DriftCTL) Run() (*analyser.Analysis, error) {
...
 
+	if d.opts.Deep {
+		remoteResources, err = d.readResources(analysis)
+		if err != nil {
+			return nil, err
+		}
+
+		d.analyzer.CompareDetails(remoteResources, resourcesFromState)
+	}
+	d.analyzer.SetAlerts()
```
moadibfr, line 165: this looks really strange to me as you use an old return analysis that will be modified by this call...

pkg/analyser/analyzer_test.go
```diff
@@ -1026,7 +1026,10 @@ func TestAnalyze(t *testing.T) {
 				addSchemaToRes(drift.res, repo)
 			}
 
-			result, err := analyzer.Analyze(c.cloud, c.iac)
+			result := analyzer.CompareEnumeration(c.cloud, c.iac)
+			_ = analyzer.CompareDetails(c.cloud, c.iac)
+			analyzer.SetAlerts()
+			result.SortResources()
```
moadibfr, line 1032: I'ts confusing that you use the first call results here and that it works

moadibfr (changes requested):

pkg/driftctl.go
sundowndev: Yes because it was never used, there is no error to return here, should we keep it anyway ?

pkg/analyser/analyzer.go
sundowndev: Because we need to keep the same analysis between CompareEnumeration and CompareDetails

pkg/driftctl.go
moadibfr: no there is no problem with that but you left the error check on the next line ?

pkg/analyser/analyzer.go
moadibfr: we could take the old analysis in arg, if you want to keep a local analysis inside the object you should not return a pointer of it to the outside this is error prone and that's also what led to having an api like
```go
analysis := anaylzer.CompareEnumeration()
// analysis does not have correct alert a this point
anaylzer.SetAlerts()
// Now we didn't do anything to analysis from our pov but the alert are set
```

sundowndev:
@moadibfr I updated the changes as per our 1:1 discussion, does this match what we talked about ? Let me know if it misses anything.

moadibfr (dismissed):
This is not exactly what I had in mind but it's finefor me :+1: ;)

sundowndev:
> This is not exactly what I had in mind but it's finefor me +1 ;)

We can spend more time on this if you want, this is totally fine, do you think the API could've been better ?

moadibfr:
> > This is not exactly what I had in mind but it's finefor me +1 ;)
> 
> We can spend more time on this if you want, this is totally fine, do you think the API could've been better ?

No don't worry I think the api is understandable and doesn't hide anything so I't fine !

sundowndev:
This is a list of middlewares we have to check before merging this PR. Each middleware should run properly in non-deep mode. We can know this if all the required attributes are present at the resource enumeration, or if the middleware doesn't use any resource attributes at all.

- [ ] aws_api_gateway_deployment_expander
- [ ] aws_api_gateway_resource_expander
- [ ] aws_api_gateway_rest_api_expander
- [ ] aws_api_gateway_rest_api_policy_expander
- [ ] aws_bucket_policy_expander
- [ ] aws_console_api_gateway_gateway_response
- [ ] aws_default_igw_route
- [ ] aws_default_internet_gateway
- [ ] aws_default_network_acl
- [ ] aws_default_network_acl_rule
- [ ] aws_default_route
- [ ] aws_default_route_table
- [ ] aws_default_security_group_rule
- [ ] aws_defaults
- [ ] aws_default_sqs_queue_policy
- [ ] aws_default_subnet
- [ ] aws_default_vpc
- [ ] aws_eip_association_expander
- [ ] aws_iam_policy_attachement_transformer
- [ ] aws_instance_block_device
- [ ] aws_instance_eip
- [ ] aws_nat_gateway_eip_assoc
- [ ] aws_network_acl_expander
- [ ] aws_rds_cluster_instance_expander
- [ ] aws_role_managed_policy_expander
- [ ] aws_route_table_expander
- [ ] aws_sns_topic_policy_expander
- [ ] aws_sqs_queue_policy_expander
- [ ] azurerm_route_expander
- [ ] azurerm_subnet_expander
- [ ] google_default_iam_member
- [ ] google_iam_binding_transformer
- [ ] google_iam_policy_transformer
- [ ] google_legacy_bucket_iam_member
- [ ] google_storage_bucket_iam_binding_transformer
- [ ] iam_policy_attachment_expander
- [ ] route53_records
- [ ] route53_records_id_reconcilier
- [ ] s3_bucket_acl
- [ ] tags_all_manager
- [ ] vpc_security_group_default
- [ ] vpc_security_group_rules

moadibfr (approved):

PR #1276, opened 2021-12-13
feat: add support for google_folder

bgdanix:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | no
| 🚀 New feature?   | yes
| ⚠ Deprecations?   | no
| ❌ BC Break       | no
| 🔗 Related issues | N/A
| ❓ Documentation  | yes 

## Description

Added support for google_folder resource that can be created for GCP Organizations. A new env variable is required: `CLOUDSDK_ORGANIZATION`


bgdanix:
Bug: because a new env var is required, driftctl will fail if this variable is not set. Not sure on the best approach to fix this.

Adding a check in the `searchAllOrgResources` for validating first if the `s.config.Organization` is empty or not, before continuing, will break the folder_scanner_test

bgdanix:
I eventually made `CLOUDSDK_ORGANIZATION` optional by checking if it is set and only then scan for resources inside the GCP Org. Also added a setter for `GCPTerraformProvider` in order to mock the test

PR #1277, opened 2021-12-13
add support for apigatewayv2_integration

moadibfr:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | no
| 🚀 New feature?   | yes
| ⚠ Deprecations?   | no
| ❌ BC Break       | no
| 🔗 Related issues | #1258 
| ❓ Documentation  | yes 

## Description

add support for apigatewayv2_integration

PR #1278, opened 2021-12-13
support aws apigatewayv2 integration response

moadibfr:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | no
| 🚀 New feature?   | yes
| ⚠ Deprecations?   | no
| ❌ BC Break       | no
| 🔗 Related issues | #1271
| ❓ Documentation  | yes 

## Description

support aws apigatewayv2 integration response


craigfurman (commented):
I've only glanced at the spec, but it looks like there's a way to specify responses in OpenAPI declarations in the `body` field of the API gateway. If so, we'll need to write middleware in a similar way to what we did for the v2_route.

Other that that, everything looks fine, but this is my first resource review (!), and having only implemented 1 resource myself, I'm finding it hard to distinguish the functionality from the "boilerplate". That's a topic for another thread though - I only mention this to warn that this review risks being rubber-stamp-y, as new resource diffs are quite large.

moadibfr:
> I've only glanced at the spec, but it looks like there's a way to specify responses in OpenAPI declarations in the `body` field of the API gateway. If so, we'll need to write middleware in a similar way to what we did for the v2_route.

Totally right but that's mainly an issue for the first PR (#1277) I'm going to take a look at this :+1:

craigfurman:
Oh I see... I didn't know that other PR existed, this one contains commits from the other, if I'm reading it correctly.

Do you actually want review now, or should we wait until after this is rebased on #1277?

moadibfr:
> Do you actually want review now, or should we wait until after this is rebased on #1277?

Yeah I mentioned it in the PR description 

> warning this is based on #1277 as I need to list integrations so please do not merge and review only the second commit

I had to base this pr on the first one so the repository method to list integrations exist. I may put this one into draft if you think it's more clear.

craigfurman:
I think let's wait for #1277, because that's quite a "mobile platform". Review effort here may be wasted if there are significant changes there.

pkg/resource/resource_types.go
```diff
@@ -143,18 +143,19 @@ var supportedTypes = map[string]ResourceTypeMeta{
...
-	"aws_launch_template":             {},
-	"aws_launch_configuration":        {},
-	"aws_apigatewayv2_api_mapping":    {},
+	"aws_apigatewayv2_model":                {},
+	"aws_apigatewayv2_stage":                {},
+	"aws_apigatewayv2_route_response":       {},
+	"aws_apigatewayv2_deployment":           {},
+	"aws_apigatewayv2_domain_name":          {},
+	"aws_launch_template":                   {},
+	"aws_launch_configuration":              {},
```
wbeuil, line 152: Can you put these two `aws_launch` resources at the end? Let's regroup all v2 types.

wbeuil (changes requested):
There's one file missing (lock file of the terraform file after the acc test) and it misses the test inside `pkg/resource/aws/metadata_test.go` file.

wbeuil (approved):

PR #1282, opened 2021-12-14
Avoid triggering auth in GCS backend tests

sundowndev:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | no
| 🚀 New feature?   | no
| ⚠ Deprecations?   | no
| ❌ BC Break       | no
| 🔗 Related issues | #...
| ❓ Documentation  | no

### DIFF
--- a/pkg/iac/terraform/state/backend/gs_reader_test.go
+++ b/pkg/iac/terraform/state/backend/gs_reader_test.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"os"
 	"reflect"
 	"testing"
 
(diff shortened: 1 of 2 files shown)


wbeuil (approved):

eliecharra:
⚠️ We should not do any operation in a `New` method. If the `storage.NewClient` is performing IO or network operations, it should be moved to the Read method. 

We should not have to set `GOOGLE_APPLICATION_CREDENTIALS` in any test file because we are not supposed to reach GCP auth code in theses tests. That is a very hacky workaround, please open a tech debt issue or PR to handle this properly.

PR #1291, opened 2021-12-16
User large circleci executors

eliecharra:
## Description

This PR goal is to speed-up test check in PRs

- Use larger circleci executors
- Update circleci/go orb: allow us to reuse go binary cache
- Add go mod caching

**Before**
image

**After**
image

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -18,6 +18,7 @@ jobs:
     parameters:
       pattern:
         type: string
+    resource_class: large
     machine:
       image: ubuntu-2004:202010-01
     environment:
@@ -75,6 +76,7 @@ jobs:
       - run: make install-tools
       - run: make lint
   test:
+    resource_class: large
     machine:
       image: ubuntu-2004:202107-02
     steps:
@@ -92,6 +94,7 @@ jobs:
       - store_test_results:
           path: ./
   release:
+    resource_class: large
     docker:
       - image: golang:1.16
     steps:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -54,6 +55,9 @@ jobs:
               -H "Content-Type: application/json" \
               -d "{\"content\": \"❌ Acceptance tests failed\nSuite: ${ACC_PATTERN}\n<${CIRCLE_BUILD_URL}>\" }"\
               ${DISCORD_WEBHOOK}
+      - go/save-cache:
+          key: test
```
wbeuil, line 59: shouldn't it be `test_acc` as the key here?

.circleci/config.yml
```diff
@@ -75,23 +79,28 @@ jobs:
       - run: make install-tools
       - run: make lint
   test:
+    resource_class: large
     machine:
       image: ubuntu-2004:202107-02
     steps:
       - checkout
+      - go/install:
+          version: "1.16.2"
```
sundowndev, line 88: Shouldn't it be 1.17.x ?

eliecharra: We do not have switched to go 1.17 builds so far for driftct, why do we want to run our test not to the target build version not sure to get it ? 🤔

.circleci/config.yml
eliecharra: Nice catch thanks 👍🏻

.circleci/config.yml
sundowndev: I forgot we didn't move to 1.17 yet, nvm

wbeuil (approved):

PR #1294, opened 2021-12-17
Use docker executor for tests

eliecharra:
## Description

Reduce the time of test.
We want to keep the machine executor for acc tests since they are more heavy.

**Before**
image

**After**
image

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -80,21 +80,19 @@ jobs:
       - run: make lint
   test:
     resource_class: large
-    machine:
-      image: ubuntu-2004:202107-02
+    executor:
+        name: go/default
+        tag: '1.16.2'
     steps:
       - checkout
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -80,21 +80,19 @@ jobs:
...
-          key: test
+          key: test-docker
       - run: make install-tools
       - run:
           name: Run tests
           command: make test
       - go/save-cache:
-          key: test
-          path: /home/circleci/.go_workspace/pkg/mod
+          key: test-docker
```
sundowndev, line 95: Isn't the path required ? What's cached in this case ?

eliecharra: The default orb executor is `cimg/go` and thus the default cache path match the good one by default too

moadibfr (approved):

Issue #1284, opened 2021-12-14
Be able to change the GCP scope


eliecharra:
**Description**
Currently driftctl is only able to scan resources under a project scope. 

To be able to have a global view of the org, driftctl should be able to scan a whole organization or folder.
We also need that if we want to add the support for `google_folder` or `google_project` resources.
So we need to find a way to specify the root GCP scope.

**Proposals**

```
$ driftctl scan --to gcp+tf://${SCOPE}
$ driftctl scan --to gcp+tf --gcp-scope ${SCOPE}
```

`${SCOPE}` can follow GCP convention:

```go
	// The allowed values are:
	//
	// * projects/{PROJECT_ID} (e.g., "projects/foo-bar")
	// * projects/{PROJECT_NUMBER} (e.g., "projects/12345678")
	// * folders/{FOLDER_NUMBER} (e.g., "folders/1234567")
	// * organizations/{ORGANIZATION_NUMBER} (e.g., "organizations/123456")
```

So we can end up with two options:

```
$ driftctl scan --to gcp+tf://organizations/123456
$ driftctl scan --to gcp+tf --gcp-scope organizations/123456
```

References:
- This issue was reported by @bgdanix (trepx) thanks a lot for this feedback 🙏🏻 
- https://pkg.go.dev/google.golang.org/genproto/googleapis/cloud/asset/v1#SearchAllResourcesRequest

eliecharra:
> Totally agree with having this implemented and I assume in order to not break the current functionality, the --gcp-scope should accept a list, as in providing a list of projects for example because my understanding is that there's already support for multiple states. So if those TF states target multiple projects, then it makes sens to have a list.

Good point, I've updated the issue description 👍🏻

PR #1296, opened 2021-12-20
Add support for GCP scopes

bgdanix:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | no
| 🚀 New feature?   | yes
| ⚠ Deprecations?   | no
| ❌ BC Break       | yes
| 🔗 Related issues | #1284
| ❓ Documentation  | yes

## Description

Add support for specifying the search scopes when using GCP provider


pkg/remote/google/repository/cloudresourcemanager.go
```diff
@@ -25,26 +28,40 @@ func NewCloudResourceManagerRepository(service *cloudresourcemanager.Service, co
...
 
-	return bindingsByProject, nil
+	if len(errorsByProject) > 0 {
+		for project, errval := range errorsByProject {
+			erorsString = erorsString + "Project: " + project + " had the following error: " + errval.Error() + "; "
+		}
+		return bindingsByProject, errors.New(erorsString)
+	} else {
+		return bindingsByProject, nil
+	}
```
sundowndev, line 66: this else case is useless, you can just return at the end

pkg/remote/google/repository/asset.go
```diff
@@ -75,101 +74,111 @@ func NewAssetRepository(client *asset.Client, config config.GCPTerraformConfig,
...
-		it := s.client.SearchAllResources(context.Background(), req)
-		for {
-			resource, err := it.Next()
-			if err == iterator.Done {
-				break
-			}
-			if err != nil {
-				return nil, err
+	for _, scope := range s.config.Scope {
+		cacheKey := "SearchAllResources" + scope
```
sundowndev, line 133: Could you please use the same pattern as for other repositories ?

```go
cacheKey := fmt.Sprintf("SearchAllResources_%s", scope)
```

sundowndev (changes requested):
Thanks for the PR @bgdanix 👍🏻 

I left some comments about the implementation, I generally agree with the approach, I think code now lack of readability and testing.

Some issues I had while testing your code : 

- The identifier for the organization is wrongly asserted and lead to a rpc error.
- I get an unexpected result when one of the GCP scopes is having permission issues, it ignores listing of some resources for all scopes. This is expected from a technical POV, but not from a user POV. We may have UX concerns about this behavior.

Could you address the comments until we review this again ? It'd be nice to get a review from someone else on this one.

pkg/cmd/scan.go
```diff
@@ -323,6 +345,26 @@ func scanRun(opts *pkg.ScanOptions) error {
 	return nil
 }
 
+func parseScopeFlag(scope []string) ([]string, error) {
+
+	scopeRegex := `projects/\S*$|folders/\d*$|organizations/\S*$`
```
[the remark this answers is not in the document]
bgdanix: I've crafted this regex. The reason I did not put more constraints for organizations was to make future-proof, in case GCP will switch to a different style of ID. But thinking about it, I did put the constraint on the folder scope, I'll add it also to the org

pkg/remote/google/repository/cloudresourcemanager.go
bgdanix: Agree, implemented

pkg/remote/google/config/config.go
```diff
@@ -1,7 +1,5 @@
 package config
 
 type GCPTerraformConfig struct {
-	Project string `cty:"project"`
-	Region  string `cty:"region"`
-	Zone    string `cty:"zone"`
+	Scope []string `cty:"scope"`
```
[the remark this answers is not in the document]
bgdanix: Implemented

pkg/remote/google/repository/cloudresourcemanager.go
```diff
@@ -25,26 +28,40 @@ func NewCloudResourceManagerRepository(service *cloudresourcemanager.Service, co
...
-	}
 
-	request := new(cloudresourcemanager.GetIamPolicyRequest)
-	policy, err := s.service.Projects.GetIamPolicy(s.config.Project, request).Do()
-	if err != nil {
-		return nil, err
-	}
+	bindingsByProject := make(map[string]map[string][]string)
+	errorsByProject := make(map[string]error)
+	var erorsString string
```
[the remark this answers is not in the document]
bgdanix: Implemented

pkg/cmd/scan.go
```diff
@@ -124,6 +140,12 @@ func NewScanCmd(opts *pkg.ScanOptions) *cobra.Command {
 		false,
 		"Do not display anything but scan results",
 	)
+	fl.StringSliceP(
+		"gcp-scope",
+		"s",
```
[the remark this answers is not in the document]
bgdanix: Done

pkg/remote/google/repository/asset.go
bgdanix: Done

bgdanix:
>     * The identifier for the organization is wrongly asserted and lead to a rpc error.
> 
>     * I get an unexpected result when one of the GCP scopes is having permission issues, it ignores listing of some resources for all scopes. This is expected from a technical POV, but not from a user POV. We may have UX concerns about this behavior.

Thank you @sundowndev for taking time to review this. I have addressed all your comments. The above reported issues should also be fixed now.

pkg/cmd/scan.go
```diff
@@ -142,7 +142,7 @@ func NewScanCmd(opts *pkg.ScanOptions) *cobra.Command {
 	)
 	fl.StringSliceP(
 		"gcp-scope",
-		"s",
+		"",
```
sundowndev, line 145: Please use `fl.StringSlice()` instead so you don't need this empty string

pkg/remote/google/repository/asset.go
```diff
@@ -76,9 +78,10 @@ func NewAssetRepository(client *asset.Client, config config.GCPTerraformConfig,
 func (s assetRepository) listAllResources(ty string) ([]*assetpb.Asset, error) {
 
 	filteredResults := []*assetpb.Asset{}
+	var erorString string
```
sundowndev, line 81: errorString*

pkg/remote/google/repository/asset_test.go
```diff
@@ -52,10 +52,10 @@ func Test_assetRepository_searchAllResources_CacheMiss(t *testing.T) {
...
 	}
 	c := &cache.MockCache{}
-	c.On("GetAndLock", "SearchAllResources").Return(nil).Times(1)
-	c.On("Unlock", "SearchAllResources").Times(1)
-	c.On("Put", "SearchAllResources", mock.IsType([]*assetpb.ResourceSearchResult{})).Return(false).Times(1)
-	repo := NewAssetRepository(assetClient, config.GCPTerraformConfig{Scope: []string{""}}, c)
+	c.On("GetAndLock", "SearchAllResources_").Return(nil).Times(1)
+	c.On("Unlock", "SearchAllResources_").Times(1)
+	c.On("Put", "SearchAllResources_", mock.IsType([]*assetpb.ResourceSearchResult{})).Return(false).Times(1)
+	repo := NewAssetRepository(assetClient, config.GCPTerraformConfig{Scopes: []string{""}}, c)
```
sundowndev, line 58: Is there a particular reason to use an empty scope here ? I find it confusing to read

pkg/remote/google/repository/cloudresourcemanager.go
```diff
@@ -58,10 +59,10 @@ func (s *cloudResourceManagerRepository) ListProjectsBindings() (map[string]map[
 
 	if len(errorsByProject) > 0 {
 		for project, errval := range errorsByProject {
-			erorsString = erorsString + "Project: " + project + " had the following error: " + errval.Error() + "; "
+			erorString = erorString + fmt.Sprintf("Project: %s had the following error: %s; ", project, errval.Error())
```
sundowndev, line 62: Not sure to get the end of the template `%s; `, the semicolon is unnecessary here

sundowndev (changes requested):

pkg/cmd/scan.go
bgdanix: Ah, yeah, makes sense. Done

pkg/remote/google/repository/asset.go
bgdanix: Missed it. Now should be ok

pkg/remote/google/repository/asset_test.go
bgdanix: The only reason I kept it empty was because previously it also had empty project, region, etc. Now I've added a dummy scope

pkg/remote/google/repository/cloudresourcemanager.go
bgdanix: The reason I've put the semicolon is to separate between multiple errors. E.g. 
`For scope projects/123456 on resource TheResource got error: rpc error: code = 123 desc = description; For scope projects/123456 on resource AnotherResource got error: abc`

pkg/remote/google/repository/asset.go
```diff
@@ -75,102 +76,131 @@ func NewAssetRepository(client *asset.Client, config config.GCPTerraformConfig,
...
+	filteredResults := []*assetpb.Asset{}
+	var errorString string
+
+	for _, scope := range s.config.Scopes {
+		cacheKey := fmt.Sprintf("listAllResources_%s", scope)
+		cachedResults := s.cache.GetAndLock(cacheKey)
+		defer s.cache.Unlock(cacheKey)
+
+		req := &assetpb.ListAssetsRequest{
+			Parent:      scope,
```
eliecharra, line 92: I'm not sure that a folder is supported for `listAssets` call according to the documentation here : https://cloud.google.com/asset-inventory/docs/reference/rpc/google.cloud.asset.v1#google.cloud.asset.v1.ListAssetsRequest

Can you double check that ?

pkg/remote/google/provider.go
```diff
@@ -55,10 +53,8 @@ func (p *GCPTerraformProvider) Version() string {
 	return p.version
 }
 
-func (p *GCPTerraformProvider) GetConfig() config.GCPTerraformConfig {
+func (p *GCPTerraformProvider) SetConfig(scopes []string) config.GCPTerraformConfig {
 	return config.GCPTerraformConfig{
-		Project: os.Getenv("CLOUDSDK_CORE_PROJECT"),
-		Region:  os.Getenv("CLOUDSDK_COMPUTE_REGION"),
-		Zone:    os.Getenv("CLOUDSDK_COMPUTE_ZONE"),
```
eliecharra, line 62: Why do you removed Region and Zone parameters ?

pkg/remote/google/provider.go
```diff
@@ -55,10 +53,8 @@ func (p *GCPTerraformProvider) Version() string {
...
+		Scopes: scopes,
```
eliecharra, line 58: I think you missed the point, scopes are not what we are expecting here in this struct https://registry.terraform.io/providers/hashicorp/google/latest/docs/guides/provider_reference#scopes

I think this PR is more complicated than you think, it seems that it is not possible to initialize a `google` terraform provider with a folder or organization scope (not to be confused with OAuth scopes). You can only configure a project.

We probably need to create another kind of detail fetcher that will lazy instantiate one terraform providers per project. We are doing that for S3 buckets, but it's a really tricky thing.

Did you try to run this PR in deep mode ? I think this is broken

bgdanix:
@sundowndev I've changed the code a bit to cater for cases when multiple scopes are used and one (or all) of them has permission error. Might not be the best way, but couldn't think of a better implementation for this case.

pkg/remote/google/provider.go
```diff
@@ -55,10 +53,8 @@ func (p *GCPTerraformProvider) Version() string {
 	return p.version
 }
 
-func (p *GCPTerraformProvider) GetConfig() config.GCPTerraformConfig {
+func (p *GCPTerraformProvider) SetConfig(scopes []string) config.GCPTerraformConfig {
```
eliecharra, line 56: `GetConfig` was ok, we are not setting anything in this function

pkg/cmd/scan.go
```diff
@@ -62,6 +62,22 @@ func NewScanCmd(opts *pkg.ScanOptions) *cobra.Command {
...
+			if to == common.RemoteGoogleTerraform && len(GCPScope) > 0 {
+				limitScope, err = parseScopeFlag(GCPScope)
+				if err != nil {
+					return err
+				}
+			} else if to != common.RemoteGoogleTerraform && len(GCPScope) > 0 {
+				return errors.New("gcp-scope can only be utilized when using " + common.RemoteGoogleTerraform + " flag")
+			} else if to == common.RemoteGoogleTerraform && len(GCPScope) == 0 {
+				return errors.New("gcp-scope must be specified when using " + common.RemoteGoogleTerraform + " flag")
+			}
```
eliecharra, line 77: Plea try to avoid else-if statements, there is not reason do use them here

eliecharra (changes requested):
This PR cannot work as it, terraform provider should be reworked. 
I will do another review pass later

pkg/remote/google/repository/asset.go
bgdanix: Indeed. Ideally we should not use listAssets but only searchAssets as the main difference is that list takes a snapshot of how the assets were at a particular time, which could result in a diff from what it's actually present.

pkg/remote/google/provider.go
bgdanix: Because they are not needed since we use scopes.

pkg/remote/google/provider.go
bgdanix: The entire point I had in mind was to support multiple or boarder scopes for scanning. I did not run it in deep mode since it's not recommended, but for a simple mode targeting various projects or an entire GCP org it works really good for me:

```json
	"summary": {
		"total_resources": 400,
		"total_changed": 0,
		"total_unmanaged": 398,
		"total_missing": 1,
		"total_managed": 1
	}
```

pkg/remote/google/provider.go
bgdanix: Well there's also nothing to get as during automated tests we have to set the config (scope)

pkg/remote/google/repository/asset.go
eliecharra: Where did you get this info ? I asked this question [here](https://stackoverflow.com/questions/69648122/difference-between-listassets-and-searchallresources) a couple of month ago and I didn't get any answer.

bgdanix: From [here](https://cloud.google.com/asset-inventory/docs/listing-assets) there are a couple of hints about it:

```
Listing assets might not meet the performance requirements for large scale customers. If this is the case for you and calling list API encounters timeout, it's recommended to use export API instead.
```

Also in [the API](https://cloud.google.com/asset-inventory/docs/reference/rest/v1/assets) it's mentioned the purpose is to list resources from a specific timeframe

```
Lists assets with time and resource types and returns paged results in response.
readTime:  Timestamp to take an asset snapshot.  If not specified, the current time will be used. Due to delays in resource data collection and indexing, there is a volatile window during which running the same query may get different results.
```

While the search is more appropriate for our intent:

```
The Cloud Asset API allows you to use a custom query language to query resource metadata on a project, folder, or organization.
Method: searchAllResources. Searches all Cloud resources within the specified scope, such as a project, folder, or organization. 
```

PR #1297, opened 2021-12-20
Document meaninglessness of test tfstate values

craigfurman:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | no
| 🚀 New feature?   | no
| ⚠ Deprecations?   | no
| ❌ BC Break       | no
| 🔗 Related issues | 
| ❓ Documentation  | no

## Description

This repo contains many tfstate files for testing purposes. Resource
names and other values in here are just opaque strings, and no running
code attempts to access such resources.

Documenting this might prevent bug bounty hunters from filing security
reports about things like unclaimed S3 bucket names.

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,3 +76,12 @@ Build with ❤️️ from 🇫🇷 🇬🇧 🇯🇵 🇬🇷 🇸🇪 🇺🇸 
 We have 30 min slots for 1:1 [tech discussions and overviews](https://calendly.com/stephanejourdan/30min) twice a day to help you get started and collect users feedback.
 
 Feel free to book a slot.
+
+---
+
+## Security notice
+
+All Terraform state and Terraform files in this repository are for unit test
+purposes only. No running code attempts to access these resources (except to
+create and destroy them, in the case of acceptance tests). They are just opaque
+strings.


Tomer-Avraham (approved):

README.md
```diff
@@ -76,3 +76,12 @@ Build with ❤️️ from 🇫🇷 🇬🇧 🇯🇵 🇬🇷 🇸🇪 🇺🇸
...
 Feel free to book a slot.
+
+---
+
+## Security notice
+
+All Terraform state files (`*.tfstate`) in this repository are for unit test
+purposes only. They do not reference resources in any particular cloud
+environment, and no running code attempts to access them. They are just opaque
+strings.
```
eliecharra, line 87: We also have a bunch of `.tf` files that are never applied but that can be used to recreate some testing context for local dev. I think we should we reference theses files here, not only the states.

I'm also wondering if we can extend the scope of this notice to the whole repo for both `.tfstate` and `.tf` files.
We do not plug any real cloud env on them and the only one that are applied are the ones for acceptance testing.
It still does not make sense to trigger security issues on theses files. BTW we may also want to have assumed vulnerable files here for testing purposes.

eliecharra (commented):
I'm fine with that, but I think we can scope the whole repo instead of only test folders

README.md
craigfurman: Thanks @eliecharra, good idea. I've removed the deeply-nested readme, and updated the notice in the top-level readme to apply to TF (not just state) and made the message broader and shorter. WDYT?

eliecharra (approved):
Thanks a lot for this quick fix 👍🏻

PR #1298, opened 2021-12-21
Configure github-cli in circleci

sundowndev:
## Description

This PR installs the github-cli in CI pipelines and edit the changelog script to authenticate using the GITHUB_TOKEN env variable.

### References

- https://circleci.com/developer/orbs/orb/circleci/github-cli
- https://cli.github.com/manual/

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -13,6 +13,7 @@ orbs:
   go: circleci/go@1.7.0
   codecov: codecov/codecov@3.1.0
   snyk: snyk/snyk@1.1.2
+  gh: circleci/github-cli@1.1.0
 jobs:
   test_acc:
     parameters:
@@ -103,6 +104,8 @@ jobs:
       - image: golang:1.16
     steps:
       - checkout
+      - gh/setup:
+            version: 2.2.0
       - run:
           name: "Ensure GnuPG is available"
           command: gpg --version


scripts/changelog.sh
```diff
@@ -9,6 +9,10 @@ LATEST_TAG=$(git describe --abbrev=0) # Get the least created tag
 DEFAULT_BRANCH=origin/HEAD
 BASE=$DEFAULT_BRANCH # Change this if you don't want to use the default branch as base
 
+if [ -n "$GITHUB_TOKEN" ]; then
+    echo -n $GITHUB_TOKEN | gh auth login --with-token &> /dev/null
+fi
+
```
eliecharra, line 15: I don't think the changelog script should handle that setup.

I think it make more sense to have a working env before invoking the script, maybe you can reuse the `setup` command of the GH orb to do the login directly in CircleCI, the script is asserting that we are logged-in with a `gh auth status` and this is the only thing that it should do IMO

https://circleci.com/developer/orbs/orb/circleci/github-cli#usage-setup

eliecharra (changes requested):

moadibfr (dismissed):

scripts/changelog.sh
moadibfr: Hmm yeah now that you said it it may make more sense to fail instead of setuping here

sundowndev: You're right, gh/setup does the job as intended

moadibfr (approved):

eliecharra (approved):

PR #1302, opened 2021-12-22
Use second most recent tag in changelog script

sundowndev:
## Description

Since the tag is already created when releasing, we need to compare commits based on the second most recent tag, not the default branch. For now we just get an empty changelog so goreleaser fallback to its default changelog generator.


craigfurman (approved):
Good catch

PR #1306, opened 2022-01-04
Update readme

sundowndev:
## Description

Slight update to our readme since we now support more cloud providers. Also removed markdown separators as I find them useless, if it's an issue I can undo those changes.


karniwl (approved):
💅

Issue #1311, opened 2022-01-06
GCP: panic: interface conversion: interface {} is nil, not []interface {}


nie-kita2:
**Description**

Running driftctl against GCP infrastructure, I get the panic message.

The Go trace:
```
panic: interface conversion: interface {} is nil, not []interface {}

goroutine 1 [running]:
github.com/snyk/driftctl/pkg/middlewares.(*GoogleIAMBindingTransformer).Execute(0xc00025c408, 0xc00012f1b8, 0xc00012f218, 0x1, 0x1)
	/root/project/pkg/middlewares/google_iam_binding_transformer.go:44 +0xa27
github.com/snyk/driftctl/pkg/middlewares.Chain.Execute(0xc0007fc500, 0x29, 0x50, 0xc00012f1b8, 0xc00012f218, 0xc0007fc500, 0x26)
	/root/project/pkg/middlewares/chain_middleware.go:21 +0x24c
github.com/snyk/driftctl/pkg.DriftCTL.Run(0x29aff80, 0xc0008f8a40, 0x29afe20, 0xc00083ca80, 0x29afd60, 0xc0002843a8, 0xc00083caa0, 0x29b03a0, 0xc0005d86f0, 0x29ee0e0, ...)
	/root/project/pkg/driftctl.go:135 +0xf4a
github.com/snyk/driftctl/pkg/cmd.scanRun(0xc000644320, 0x0, 0x0)
	/root/project/pkg/cmd/scan.go:283 +0x9d8
github.com/snyk/driftctl/pkg/cmd.NewScanCmd.func2(0xc0000bcb00, 0xc000114f00, 0x0, 0x6, 0x0, 0x0)
	/root/project/pkg/cmd/scan.go:115 +0x2a
github.com/spf13/cobra.(*Command).execute(0xc0000bcb00, 0xc000114ea0, 0x6, 0x6, 0xc0000bcb00, 0xc000114ea0)
	/go/pkg/mod/github.com/spf13/cobra@v1.0.0/command.go:842 +0x472
github.com/spf13/cobra.(*Command).ExecuteC(0xc000349b80, 0x268aa98, 0xc0002c46c0, 0x0)
	/go/pkg/mod/github.com/spf13/cobra@v1.0.0/command.go:950 +0x375
main.run(0x0)
	/root/project/main.go:66 +0x578
main.main()
	/root/project/main.go:27 +0x25
```

**Environment**
 - OS: Ubuntu
 - driftctl version: 0.18.3
[...]

PR #1312, opened 2022-01-06
Fix crash in GCP iam binding middleware

eliecharra:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | yes
| 🚀 New feature?   | no
| ⚠ Deprecations?   | no
| ❌ BC Break       | no
| 🔗 Related issues | #1311 
| ❓ Documentation  | no

## Description

Fix a weakness in the GCP IAM binding transformer

### DIFF
--- a/pkg/middlewares/google_iam_binding_transformer.go
+++ b/pkg/middlewares/google_iam_binding_transformer.go
@@ -39,7 +39,11 @@ func (m *GoogleIAMBindingTransformer) Execute(_, resourcesFromState *[]*resource
 
 		resName := *stateRes.Attrs.GetString(resField)
 		roleName := *stateRes.Attrs.GetString("role")
-		members, _ := stateRes.Attrs.Get("members")
+		members, exist := stateRes.Attrs.Get("members")
+
+		if !exist || members == nil {
+			continue
+		}
 
 		for _, member := range members.([]interface{}) {
 			id := fmt.Sprintf("%s/%s/%s", resName, roleName, member)
--- a/pkg/middlewares/google_iam_binding_tranformer_test.go
+++ b/pkg/middlewares/google_iam_binding_tranformer_test.go
@@ -19,6 +19,29 @@ func TestGoogleProjectIAMBindingTransformer_Execute(t *testing.T) {
 		mock               func(factory *terraform.MockResourceFactory)
 	}{
 		{
+			name: "Test that bindings with nil members does not cause any crash",
+			resourcesFromState: []*resource.Resource{
+				{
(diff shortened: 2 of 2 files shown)


craigfurman (approved):

pkg/middlewares/google_iam_binding_transformer.go
```diff
@@ -39,7 +39,11 @@ func (m *GoogleIAMBindingTransformer) Execute(_, resourcesFromState *[]*resource
 
 		resName := *stateRes.Attrs.GetString(resField)
 		roleName := *stateRes.Attrs.GetString("role")
-		members, _ := stateRes.Attrs.Get("members")
+		members, exist := stateRes.Attrs.Get("members")
```
wbeuil, line 42: Can the role be nil?

moadibfr: good question I don't think so

eliecharra: I have really no idea, I've not worked on this one. I've not found the root cause of why do we have an empty members key here in the state but I was supposing that it was either a `default [] vs nil` issue or a state refresh problem.

I would say that without a role this resource does not make sense so I don't think that it can be empty at some point, @moadibfr any idea ?

moadibfr: okm from tf doc members AND role are required so maybe it would be safe to test both ?

eliecharra: I asked the reporter for details, maybe it will help let's see if we can get the full resource definition in the state

eliecharra: Anyway, I'm wondering if we should introduce a mechanism to recover from theses panics in middleware pass and send alerts WDYT ?

craigfurman: I thought we already had existing sentry functionality wrapped around the whole invocation for exactly this? (if the user opts into crash reports)

Are you talking about a different alerting mechanism?

eliecharra: Ho yeah sorry for the poor context @craigfurman 🙏🏻 we have another internal mechanism for alerting  the user that something went wrong during the scan without interrupting it. This mechanism also allows us to ignore a resource or a whole complete resource type from the computation too (to avoid noise for resource types that encountered some errors). But good point about sentry, if we recover here we wont see theses issues in sentry, maybe it can be nice to manually send them in sentry for theses case event if we catch the panic, because it's still something unexpected.

Issue #1257, opened 2021-12-06
aws_apigatewayv2_domain_name


sundowndev:
https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/apigatewayv2_domain_name

wbeuil:
To keep an explanation of why, as a user, you will only see `aws_api_gateway_domain_name` (v1) resources as unmanaged and never `aws_apigatewayv2_domain_name` (v2), I invite you to check the PR I wrote [here](https://github.com/snyk/driftctl/pull/1320#issue-1102006740).

TL;DR, both (v1|v2) AWS SDK endpoints to list domain names return v1 AND v2 resources with no way to find which one is which. We decided to find managed resources by looping over matched remote and state ones, and leave only v1 unmanaged ones in the output.

PR #1320, opened 2022-01-13
Add aws_apigatewayv2_domain_name resource

wbeuil:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | no
| 🚀 New feature?   | yes
| ⚠ Deprecations?   | no
| ❌ BC Break       | no
| 🔗 Related issues | #1257 
| ❓ Documentation  | https://github.com/snyk/driftctl-docs/pull/211

## Description

Needed to add a new middleware to handle the reconciliation between `aws_api_gateway_domain_name` (v1) and `aws_apigatewayv2_domain_name` (v2) resources.

The all explanation can be easily sum up by a small example:

```hcl
resource "aws_api_gateway_domain_name" "foo" {
  domain_name              = "foo.example.com"
  regional_certificate_arn = aws_acm_certificate.foo.arn

  endpoint_configuration {
    types = ["REGIONAL"]
  }
}

resource "aws_apigatewayv2_domain_name" "bar" {
  domain_name = "bar.example.com"

  domain_name_configuration {
    certificate_arn = aws_acm_certificate.bar.arn
    endpoint_type   = "REGIONAL"
  }
}
```

If you try to apply these 2 resources and use the AWS SDK to list all domain names with the v1 and v2 endpoints you will get both resources as result of each endpoint as per below images.

image
image

Unfortunately there are no ways to distinct a v1 from a v2 by looking at the structs... So we decided to go the middleware way with one that will find matching domain names between a remote and a state resource and only leave v1 format as unmanaged domain names.

### DIFF
--- a/pkg/middlewares/aws_api_gateway_domain_names_reconciler.go
+++ b/pkg/middlewares/aws_api_gateway_domain_names_reconciler.go
@@ -1,0 +1,79 @@
+package middlewares
+
+import (
+	"github.com/snyk/driftctl/pkg/resource"
+	"github.com/snyk/driftctl/pkg/resource/aws"
+)
+
+// Used to reconcile API Gateway domain names (v1 and v2) from both remote
+// and state resources because v1|v2 AWS SDK list endpoints return all domain
+// names without distinction
+type AwsApiGatewayDomainNamesReconciler struct{}
+
+func NewAwsApiGatewayDomainNamesReconciler() AwsApiGatewayDomainNamesReconciler {
+	return AwsApiGatewayDomainNamesReconciler{}
+}
+
+func (m AwsApiGatewayDomainNamesReconciler) Execute(remoteResources, resourcesFromState *[]*resource.Resource) error {
+	newRemoteResources := make([]*resource.Resource, 0)
+	managedDomainNames := make([]*resource.Resource, 0)
+	unmanagedDomainNames := make([]*resource.Resource, 0)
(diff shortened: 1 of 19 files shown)


pkg/middlewares/aws_api_gateway_domain_names_reconciler.go
```diff
@@ -0,0 +1,81 @@
...
+			i--
+			continue
+		}
+
+		// Now remove only v2 domain names
+		if unmanaged.ResourceType() == aws.AwsApiGatewayV2DomainNameResourceType {
+			unmanagedDomainNames = append(unmanagedDomainNames[:i], unmanagedDomainNames[i+1:]...)
+			i--
+		}
+	}
```
eliecharra, line 73: I always feel uncomfortable by mutating array that we are currently iterating on.
I don't really know how golang handle this under the hoods but I suppose that it creates a copy of the slice 🤔

WDYT of explicitly creating a new `deduplicatedUnmanagedDomains` slice instead and appending on it while iterating on `unmanagedDomainNames` ?

I think it will lead to a code that will be easier to maintain.

@craigfurman Nevermind of this comment if your schedule is full, but I would love to have your goland expert eye on this remark 🙏🏻 I'm kinda new in the goland world and would be very happy to know more 😃
I could probably have found the answer by myself 😅

eliecharra (changes requested):

eliecharra (approved):

PR #1326, opened 2022-01-20
Add sleep-backoff to GCP ACC tests [CFG-1305]

sundowndev:
## Description

As per a discussion with @craigfurman on [CFG-1305](https://snyksec.atlassian.net/browse/CFG-1305), we agreed that retrying the scan in a tight loop was not ideal and we should use some linear or exponential sleep-backoff instead. The common problem that seem to affect all GCP acc tests is that the asset API sometimes return cached result, making new resource appear as missing ([example](https://app.circleci.com/pipelines/github/snyk/driftctl/4118/workflows/5ab1d12f-eb0b-40c2-bdf1-cad9a47dd1b6/jobs/8763)). The [GCP docs](https://cloud.google.com/asset-inventory/docs/faq#result-stale) says : 

> Data freshness in the Cloud Asset API is on a best-effort basis. While almost all asset updates will be available to clients in minutes, in rare cases it's possible the result of the Cloud Asset API methods won't include the most recent asset updates.

What we were doing before basically is 
- `retry the scan if it didn't succeed and we've been scanning for less than a minute`.

What this change suggests is 
- `retry the scan if it didn't succeed, after 3 minutes, once`.

Retrying in a loop for a minute appear to be less efficient than retrying once after some time. I think 3 minutes is a reasonable time frame for an nightly run.

### DIFF
--- a/pkg/resource/google/google_bigquery_dataset_test.go
+++ b/pkg/resource/google/google_bigquery_dataset_test.go
@@ -18,10 +18,15 @@ func TestAcc_Google_BigqueryDataset(t *testing.T) {
 		},
 		Checks: []acceptance.AccCheck{
 			{
-				// New resources are not visible immediately on GCP api after an apply operation
-				// Logic below retry driftctl scan until we can retrieve the results (infra will be in sync) and for maximum 60 seconds
+				// New resources are not visible immediately through GCP API after an apply operation.
+				// Logic below retries driftctl scan using a back-off strategy of retrying 'n' times
+				// and doubling the amount of time waited after each one.
 				ShouldRetry: func(result *test.ScanResult, retryDuration time.Duration, retryCount uint8) bool {
-					return !result.IsSync() && retryDuration < time.Minute
+					if result.IsSync() || retryDuration > 10*time.Minute {
+						return false
+					}
+					time.Sleep((2 * time.Duration(retryCount)) * time.Minute)
+					return true
 				},
 				Check: func(result *test.ScanResult, stdout string, err error) {
 					if err != nil {
--- a/pkg/resource/google/google_bigquery_table_test.go
+++ b/pkg/resource/google/google_bigquery_table_test.go
@@ -18,10 +18,15 @@ func TestAcc_Google_BigqueryTable(t *testing.T) {
 		},
 		Checks: []acceptance.AccCheck{
 			{
-				// New resources are not visible immediately on GCP api after an apply operation
-				// Logic below retry driftctl scan until we can retrieve the results (infra will be in sync) and for maximum 60 seconds
+				// New resources are not visible immediately through GCP API after an apply operation.
+				// Logic below retries driftctl scan using a back-off strategy of retrying 'n' times
+				// and doubling the amount of time waited after each one.
 				ShouldRetry: func(result *test.ScanResult, retryDuration time.Duration, retryCount uint8) bool {
-					return !result.IsSync() && retryDuration < time.Minute
+					if result.IsSync() || retryDuration > 10*time.Minute {
+						return false
+					}
+					time.Sleep((2 * time.Duration(retryCount)) * time.Minute)
+					return true
 				},
 				Check: func(result *test.ScanResult, stdout string, err error) {
 					if err != nil {
--- a/pkg/resource/google/google_bigtable_instance_test.go
+++ b/pkg/resource/google/google_bigtable_instance_test.go
@@ -18,10 +18,15 @@ func TestAcc_Google_BigtableInstance(t *testing.T) {
 		},
 		Checks: []acceptance.AccCheck{
 			{
-				// New resources are not visible immediately on GCP api after an apply
-				// Logic below retry driftctl scan until we can retrieve the results (infra will be in sync) and for maximum 60 seconds
+				// New resources are not visible immediately through GCP API after an apply operation.
+				// Logic below retries driftctl scan using a back-off strategy of retrying 'n' times
+				// and doubling the amount of time waited after each one.
 				ShouldRetry: func(result *test.ScanResult, retryDuration time.Duration, retryCount uint8) bool {
-					return !result.IsSync() && retryDuration < time.Minute
+					if result.IsSync() || retryDuration > 10*time.Minute {
+						return false
+					}
+					time.Sleep((2 * time.Duration(retryCount)) * time.Minute)
+					return true
 				},
 				Check: func(result *test.ScanResult, stdout string, err error) {
 					if err != nil {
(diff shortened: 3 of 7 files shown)


craigfurman (commented):
Why retry only once? Why not use something like https://github.com/sethvargo/go-retry to retry repeatedly with some linear or exponential backoff (I'd recommend exponential if in doubt, starting at a few seconds sleep) up to some time limit, e.g. 5 or 10 mins?

sundowndev:
> Why retry only once? Why not use something like https://github.com/sethvargo/go-retry to retry repeatedly with some linear or exponential backoff (I'd recommend exponential if in doubt, starting at a few seconds sleep) up to some time limit, e.g. 5 or 10 mins?

I was thinking that we may not want to use an exponential backoff since it'd take too much time, but having a limit of 5-10 min sounds reasonable to me.

craigfurman:
I'm not particularly opinionated on whether we use linear or exponential backoff, but we should retry >1 time, up to some time limit. I'd really recommend taking a look at that library I linked, it's a very simple retry wrapper function that uses context cancellations.

sundowndev:
> I'm not particularly opinionated on whether we use linear or exponential backoff, but we should retry >1 time, up to some time limit. I'd really recommend taking a look at that library I linked, it's a very simple retry wrapper function that uses context cancellations.

We already use a [very similar library](http://github.com/eapache/go-resiliency) for retry strategies, but it doesn't fit well with the current acceptance api. Here's a snippet that nearly does the same as go-retry : 

```go
ShouldRetry: func(result *test.ScanResult, retryDuration time.Duration, retryCount uint8) bool {
	if result.IsSync() || retryDuration > 10*time.Minute {
		return false
	}
	time.Sleep((2 * time.Duration(retryCount)) * time.Minute)
	return true
},
```

I updated the changes with this example.

pkg/resource/google/google_bigquery_dataset_test.go
```diff
@@ -18,10 +18,15 @@ func TestAcc_Google_BigqueryDataset(t *testing.T) {
...
-				// Logic below retry driftctl scan until we can retrieve the results (infra will be in sync) and for maximum 60 seconds
+				// New resources are not visible immediately through GCP API after an apply operation.
+				// Logic below retries driftctl scan using a back-off strategy of retrying 'n' times
+				// and doubling the amount of time waited after each one.
 				ShouldRetry: func(result *test.ScanResult, retryDuration time.Duration, retryCount uint8) bool {
-					return !result.IsSync() && retryDuration < time.Minute
+					if result.IsSync() || retryDuration > 10*time.Minute {
+						return false
+					}
+					time.Sleep((2 * time.Duration(retryCount)) * time.Minute)
```
craigfurman, line 28: Basing the exponential backoff on the retry count seems really weird to me? You mentioned earlier you thought exponential backoff might not be optimal, and FWIW I'm totally fine with that 😁 

Also every time I said linear backoff earlier, I actually meant constant backoff - sorry!

WDYT of a simple hardcoded constant backoff to start with, and we can re-evaluate based on the flake rate after that?

```suggestion
					time.Sleep(10 * time.Second)
```

pkg/resource/google/google_bigquery_dataset_test.go
```diff
@@ -18,10 +18,15 @@ func TestAcc_Google_BigqueryDataset(t *testing.T) {
...
 		Checks: []acceptance.AccCheck{
 			{
-				// New resources are not visible immediately on GCP api after an apply operation
-				// Logic below retry driftctl scan until we can retrieve the results (infra will be in sync) and for maximum 60 seconds
+				// New resources are not visible immediately through GCP API after an apply operation.
+				// Logic below retries driftctl scan using a back-off strategy of retrying 'n' times
+				// and doubling the amount of time waited after each one.
 				ShouldRetry: func(result *test.ScanResult, retryDuration time.Duration, retryCount uint8) bool {
-					return !result.IsSync() && retryDuration < time.Minute
+					if result.IsSync() || retryDuration > 10*time.Minute {
```
craigfurman, line 25:
```suggestion
					if result.IsSync() || retryDuration < 10*time.Minute {
```

I think the direction of the inequality was correct before, right? We want to abort the retry loop if we've been retrying for longer than retryDuration, if I understand the code that calls this function correctly: https://github.com/snyk/driftctl/blob/b751cd50aefb2af820d58c589211ff0371780107/test/acceptance/testing.go#L436-L440

pkg/resource/google/google_bigquery_dataset_test.go
```diff
@@ -18,10 +18,15 @@ func TestAcc_Google_BigqueryDataset(t *testing.T) {
 		},
 		Checks: []acceptance.AccCheck{
 			{
-				// New resources are not visible immediately on GCP api after an apply operation
-				// Logic below retry driftctl scan until we can retrieve the results (infra will be in sync) and for maximum 60 seconds
+				// New resources are not visible immediately through GCP API after an apply operation.
+				// Logic below retries driftctl scan using a back-off strategy of retrying 'n' times
+				// and doubling the amount of time waited after each one.
 				ShouldRetry: func(result *test.ScanResult, retryDuration time.Duration, retryCount uint8) bool {
```
craigfurman, line 24: if we stop using retryCount, and instead retry however many times it takes to fill the max duration (10m in this PR), should we eliminate the parameter and follow the compiler errors?

```suggestion
				ShouldRetry: func(result *test.ScanResult, retryDuration time.Duration) bool {
```

Is there any reason we can't do this?

craigfurman (changes requested):
Only one of these notes is blocking, the one about the direction of the inequality boolean that can break out of the retry loop. Apologies if I've misunderstood this.

I've left a few other ideas, let me know what you think.

In all cases the notes are generalisable to all test files in this PR, I just didn't make them on all of them 😉

pkg/resource/google/google_bigquery_dataset_test.go
craigfurman: Even if we do keep the backoff as a multiple of retryCount as you originally suggest, `count * 2m` seems a bit long for the initial wait. WDYT?

sundowndev: > You mentioned earlier you thought exponential backoff might not be optimal, and FWIW I'm totally fine with that grin

I did but I also think exponential backoff can be good with a reasonable time limit. Your initial suggestion was good actually! A constant backoff can lead to more friction as it's impossible to know what's the right time frame to choose in this case. Exponential suggests we keep using a best-effort approach : retrying immediately, then after just 2 minutes, then after a bit longer... 

If after all we stay on a hardcoded backoff, I think 3 minutes per retry for a maximum of 3 attempts (9min) is enough for now. WDYT?

craigfurman: Yeah let's start with this actually, I'll approve and we can always re-evaluate if the flakes continue.

FWIW I think this is linear backoff, not exponential (2,4,6.., not 2,4,8,16...), but whatever 😉

craigfurman (approved):
Thanks for this @sundowndev, hopefully this will drive down the flakes 🎊

Issue #1217, opened 2021-11-15
Default `aws_api_gateway_account` should be ignored 


wbeuil:
**Description**

We all have an `aws_api_gateway_account` resource inside each region of AWS that will by default output as unmanaged. We should ignore this one unless you manage it by Terraform.

**How to reproduce**

Simply test with the most recent version of driftctl `driftctl scan`, you should have one unmanaged`aws_api_gateway_account` resource.

**Possible Solution**

New middleware to ignore the default resource.

sundowndev:
Released in [v0.19.0](https://github.com/snyk/driftctl/releases/tag/v0.19.0).

PR #1332, opened 2022-01-24
Ignore default aws_api_gateway_account resources

sundowndev:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | yes
| 🚀 New feature?   | no
| ⚠ Deprecations?   | no
| ❌ BC Break       | no
| 🔗 Related issues | #1217
| ❓ Documentation  | yes 

## Description

> We all have an aws_api_gateway_account resource inside each region of AWS that will by default output as unmanaged. We should ignore this one unless you manage it by Terraform.

This PR adds a simple middleware to ignore remote `aws_api_gateway_account` resources when it's not managed by Terraform.

**Notes to reviewers**:
- I didn't find a way to differentiate default and managed resources, so beware of the impact of this change
- This middleware will not run in strict mode

### DIFF
--- a/pkg/middlewares/aws_default_api_gateway_account.go
+++ b/pkg/middlewares/aws_default_api_gateway_account.go
@@ -1,0 +1,52 @@
+package middlewares
+
+import (
+	"github.com/sirupsen/logrus"
+	"github.com/snyk/driftctl/pkg/resource"
+	"github.com/snyk/driftctl/pkg/resource/aws"
+)
+
+// AwsDefaultApiGatewayAccount is a middleware that ignores default API gateway account resources in the current region
+type AwsDefaultApiGatewayAccount struct{}
+
+func NewAwsDefaultApiGatewayAccount() AwsDefaultApiGatewayAccount {
+	return AwsDefaultApiGatewayAccount{}
+}
+
+func (m AwsDefaultApiGatewayAccount) Execute(remoteResources, resourcesFromState *[]*resource.Resource) error {
+
+	newRemoteResources := make([]*resource.Resource, 0)
+
+	for _, remoteResource := range *remoteResources {
+		// Ignore all resources other than API gateway account
+		if remoteResource.ResourceType() != aws.AwsApiGatewayAccountResourceType {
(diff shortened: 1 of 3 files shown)


pkg/middlewares/aws_default_api_gateway_account.go
```diff
@@ -0,0 +1,52 @@
+package middlewares
+
+import (
+	"github.com/sirupsen/logrus"
+	"github.com/snyk/driftctl/pkg/resource"
+	"github.com/snyk/driftctl/pkg/resource/aws"
+)
+
+// AwsDefaultApiGatewayAccount is a middleware that ignores default API gateway account resources in the current region
```
wbeuil, line 9: We can only have 1 api gateway account, can you change to `that ignores the default API Gateway account resource in the current region`

pkg/middlewares/aws_default_api_gateway_account_test.go
```diff
@@ -0,0 +1,99 @@
...
+func TestAwsDefaultApiGatewayAccount_Execute(t *testing.T) {
+
+	tests := []struct {
+		name               string
+		remoteResources    []*resource.Resource
+		resourcesFromState []*resource.Resource
+		expected           []*resource.Resource
+	}{
+		{
+			"test that default account are not ignored when managed by IaC",
```
wbeuil, line 22: `is not ignored`

pkg/middlewares/aws_default_api_gateway_account_test.go
```diff
@@ -0,0 +1,99 @@
...
+				},
+				{
+					Id:    "default-managed-by-IaC",
+					Type:  aws.AwsApiGatewayAccountResourceType,
+					Attrs: &resource.Attributes{},
+				},
+			},
+		},
+		{
+			"test that default account are ignored when not managed by IaC",
```
wbeuil, line 57: `is ignored`

wbeuil (changes requested):
simple grammar comments, LGTM overall

wbeuil (approved):

Issue #1317, opened 2022-01-11
Add number of processed IaC Sources to telemetry


sjourdan:
**Description**

TLDR: let's add the number of processed states to the telemetry

Background: 

- when working on improving user journeys, a source is to look at the telemetry
- telemetry sometimes reports difficult to understand behavior, as simultaneous runs with hugely varying number of resources (total/managed)
- we suspect it could be a signal of users using different combinations of states, or gradually onboarding
- a way to know more with full respect for the users' privacy is to add a new telemetry field

**Example**

```json 
{
    "version": "0.7.1",
    "os": "linux",
    "arch": "amd64",
    "iac_source_count": 1
    "total_resources": 100,
    "total_managed": 10,
    "duration": 300,
    "provider_name": "aws"
}
```

sundowndev:
I get the use case and agree with it, I just have a remark on the wording here. I think we should avoid using Terraform wording as much as possible, so the logic/wording stays agnostic and fit with any new IaC supplier we might add in the future (Pulumi, CFN...). Why not name this field with something like `iac_source_count` instead ?

sjourdan:
great idea! let's do this. I updated the main example.

sundowndev:
Released in [v0.20.0](https://github.com/snyk/driftctl/releases/tag/v0.20.0).

PR #1336, opened 2022-01-26
Add iac source count to telemetry

sundowndev:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | no
| 🚀 New feature?   | no
| ⚠ Deprecations?   | no
| ❌ BC Break       | no
| 🔗 Related issues | #1317
| ❓ Documentation  | yes 

## Description

Please include a summary of the change and which issue is fixed.
Please also include relevant motivation and context.


pkg/resource/supplier.go
```diff
@@ -1,8 +1,9 @@
 package resource
 
-// Resource Supplier supply the list of resource.Resource, its the front to retrieve remote resources
+// Supplier supply the list of resource.Resource, it's the main interface to retrieve state or remote resources
 type Supplier interface {
 	Resources() ([]*Resource, error)
+	SourceCount() int
```
eliecharra, line 6: I don't think this make sense to add this to this interface. For remote ones we should be able to retrieve resources without implementing any `SourceCount` method.

Maybe we should introduce a new interface like `IacSupplier` and compose the `TerraformStateReader` with both `Supplier` and `IacSupplier` WDYT ?

pkg/iac/terraform/state/terraform_state_reader.go
```diff
@@ -41,6 +41,7 @@ type TerraformStateReader struct {
 	progress       output.Progress
 	filter         filter.Filter
 	alerter        *alerter.Alerter
+	keys           []string
```
eliecharra, line 44: I think we should find a better naming for that, `keys` sounds too generic for me.

Since we only use theses "keys" to return the length, why not explicitly storing the number of states here and returning it directly in `SourceCount` ? In `retrieveForState` you'll only have to increment a counter.

So we can rename the unexported `keys` attribute to `stateCount` WDYT ?

eliecharra (changes requested):
I'm fine with the wording `SourceCount`

pkg/telemetry/telemetry.go
```diff
@@ -20,6 +20,7 @@ type telemetry struct {
 	TotalManaged   int    `json:"total_managed"`
 	Duration       uint   `json:"duration"`
 	ProviderName   string `json:"provider_name"`
+	IaCSourceCount int    `json:"iac_source_count"`
```
eliecharra, line 23: Should be `uint` everywhere

eliecharra (approved):

PR #1347, opened 2022-02-14
Add retry policy for remaining GCP acc tests

sundowndev:
## Description

This PR adds a retry policy for some GCP tests that failed recently without retrying. It also creates a new common `LinearBackoff` function so we don't duplicate the retry policy for each test.

Tests where I added the retry policy : 
- `TestAcc_Google_ComputeInstance`
- `TestAcc_Google_ComputeNetwork`

**References**

- https://app.circleci.com/pipelines/github/snyk/driftctl/4195/workflows/5e665d3d-db9e-485b-bc29-1b106fd9846a/jobs/9071

### DIFF
--- a/pkg/resource/google/google_compute_instance_test.go
+++ b/pkg/resource/google/google_compute_instance_test.go
@@ -2,6 +2,7 @@ package google_test
 
 import (
 	"testing"
+	"time"
 
 	"github.com/snyk/driftctl/test"
 	"github.com/snyk/driftctl/test/acceptance"
@@ -17,6 +18,16 @@ func TestAcc_Google_ComputeInstance(t *testing.T) {
 		},
 		Checks: []acceptance.AccCheck{
 			{
+				// New resources are not visible immediately through GCP API after an apply operation.
+				// Logic below retries driftctl scan using a back-off strategy of retrying 'n' times
+				// and doubling the amount of time waited after each one.
+				ShouldRetry: func(result *test.ScanResult, retryDuration time.Duration, retryCount uint8) bool {
+					if result.IsSync() || retryDuration > 15*time.Minute {
+						return false
+					}
+					time.Sleep((2 * time.Duration(retryCount)) * time.Minute)
+					return true
+				},
 				Check: func(result *test.ScanResult, stdout string, err error) {
 					if err != nil {
 						t.Fatal(err)
--- a/pkg/resource/google/google_compute_instance_test.go
+++ b/pkg/resource/google/google_compute_instance_test.go
@@ -19,15 +19,7 @@ func TestAcc_Google_ComputeInstance(t *testing.T) {
 		Checks: []acceptance.AccCheck{
 			{
 				// New resources are not visible immediately through GCP API after an apply operation.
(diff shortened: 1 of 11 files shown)


pkg/resource/google/google_compute_instance_test.go
```diff
@@ -17,6 +18,16 @@ func TestAcc_Google_ComputeInstance(t *testing.T) {
 		},
 		Checks: []acceptance.AccCheck{
 			{
+				// New resources are not visible immediately through GCP API after an apply operation.
+				// Logic below retries driftctl scan using a back-off strategy of retrying 'n' times
+				// and doubling the amount of time waited after each one.
+				ShouldRetry: func(result *test.ScanResult, retryDuration time.Duration, retryCount uint8) bool {
```
craigfurman, line 24: Can we extract this func some where, to avoid the duplication of its logic and the explanatory comment everywhere?

test/acceptance/testing.go
```diff
@@ -499,3 +501,16 @@ func RetryFor(timeout time.Duration, f func(c chan struct{}) error) error {
 		return ctx.Err()
 	}
 }
+
+// ExponentialBackoff returns a function that retries using
+// a back-off strategy of retrying 'n' times and doubling the
+// amount of time waited after each one.
+func ExponentialBackoff(limit time.Duration) ShouldRetryFunc {
```
craigfurman, line 508: This function name and comment describe exponential backoff, but I'm pretty sure it's actually linear? We should either change the name to reflect reality, or change the impl to actually do exponential backoff. The comment "doubling every time" makes me think you intended to do exponential backoff, but before we go and change it - what has the introduction of backoff done to the flake rate of the tests so far? If the current, linear impl has been fine, we could just change the function name and comment, WDYT?

```suggestion
func LinearBackoff(limit time.Duration) ShouldRetryFunc {
```

sundowndev: > what has the introduction of backoff done to the flake rate of the tests so far?

AFAIK, we experienced less regression in our nightly tests, but I guess it's something that takes longer to observe.

image

My bad you're right, it's linear

craigfurman: Great, the overall regression is down - but I meant, do we observe _individual_ tests that have had retry logic added to them, flaking still?

craigfurman (approved):

test/acceptance/testing.go
sundowndev: > Great, the overall regression is down - but I meant, do we observe individual tests that have had retry logic added to them, flaking still?

Some of them yeah, but less often. I looked at circleci insights and according to them we don't have any flacky test anymore. The retry policy definitely decrease regression for those tests.

image

wbeuil: I got one in my mind (AWS_EIP) that has retry logic but still sometimes (last night) failed. But the rest seems to be good for now.

craigfurman: nice, that's an improvement! Maybe we want to try a switch to longer timeout, maybe with exponential backoff? Let's merge this for sure though.

Issue #1325, opened 2022-01-19
Add an identifier for `unmanaged` AWS security group rules 


optplx:
**Description**
For unmanaged security group rules, the json output currently shows only a hash generated based on the security group rule properties (such as from/to ports, protocol, sg ID). As far as I can tell, this makes it impossible to identify which rule is listed using the json report, and we have to run the "raw text" report to do so.

**Example**
Here is an example of the json output for an unmanaged security group rule:
```
{
    "id": "sgrule-123456789",
    "type": "aws_security_group_rule"
}
```
Here is an example of a raw report for the same rule:
```
- sgrule-123456789
    Ports: 442, Protocol: tcp, SecurityGroup: sg-1fsdf2345dsfsdf2s, Source: 10.0.0.0/16, Type: ingress
```

Using the json one, it is not possible to identify on AWS which rule is actually behind the `sgrule-123456789`, making the json report hard to leverage here. 

**Proposition**
Would it be possible to print the real security group rule ID instead for example?

eliecharra:
I flagged this one as a good first issue, if someone from the community want to take it, it shouldn't be too hard to handle.
Here is a few hints:

In `pkg/resource/resource.go`:
You can add new field to `SerializableResource`, something like `identityAttributes` or `humanReadableAttributes`. 
Then you need to set this field in `NewSerializableResource` based on `res.Schema().HumanReadableAttributesFunc(res)`.

An example of what is done on the console side can be found here : https://github.com/snyk/driftctl/blob/0a1ed7d01e2370fd3fed45e0120d813bdf14e7d8/pkg/cmd/scan/output/console.go#L279-L301

We may work on this on our side if nobody from the community has taken it in a couple of week since we'll need this for internal needs at some point.

Please go to discord #support channel if you need more help on this one

wakeful:
👋 @eliecharra I created a PR for this issue

moadibfr:
Ok I see this is the usecase for the opened PR then. Sorry I missed that.

patch-the-dog:
This issue has been referenced in the latest release v0.21.0.

PR #1354, opened 2022-02-17
adding `humanReadableAttributes` to JSON output

wakeful:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | no
| 🚀 New feature?   | yes
| ⚠ Deprecations?   | no
| ❌ BC Break       | no
| 🔗 Related issues | #1325
| ❓ Documentation  | no

## Description

added `human_readable_attributes` field to json output e.q.
```json
{
    "id": "sgrule-1180062852",
    "type": "aws_security_group_rule",
    "human_readable_attributes": {
        "Destination": "0.0.0.0/0",
        "Ports": "All",
        "Protocol": "All",
        "SecurityGroup": "sg-08e4f811ab524c960",
        "Type": "egress"
    }
}
```

```json
{
    "id": "nacl-4081969105",
    "type": "aws_network_acl_rule",
    "human_readable_attributes": {
        "CIDR": "0.0.0.0/0",
        "Egress": "true",
        "Network": "acl-4f2eda2b",
        "Protocol": "All",
        "Rule number": "100"
    }
}
```

### DIFF
--- a/pkg/resource/resource.go
+++ b/pkg/resource/resource.go
@@ -99,9 +99,10 @@ type ResourceFactory interface {
 }
 
 type SerializableResource struct {
-	Id     string              `json:"id"`
-	Type   string              `json:"type"`
-	Source *SerializableSource `json:"source,omitempty"`
+	Id                 string              `json:"id"`
+	Type               string              `json:"type"`
+	ReadableAttributes map[string]string   `json:"human_readable_attributes,omitempty"`
+	Source             *SerializableSource `json:"source,omitempty"`
 }
 
 func NewSerializableResource(res *Resource) *SerializableResource {
@@ -114,10 +115,18 @@ func NewSerializableResource(res *Resource) *SerializableResource {
 		}
 	}
 	return &SerializableResource{
-		Id:     res.ResourceId(),
-		Type:   res.ResourceType(),
-		Source: src,
+		Id:                 res.ResourceId(),
+		Type:               res.ResourceType(),
+		ReadableAttributes: formatReadableAttributes(res),
+		Source:             src,
 	}
+}
+
+func formatReadableAttributes(res *Resource) map[string]string {
+	if res.Schema() == nil || res.Schema().HumanReadableAttributesFunc == nil {
+		return map[string]string{}
+	}
+	return res.Schema().HumanReadableAttributesFunc(res)
 }
 
 type NormalizedResource interface {


sundowndev (changes requested):
Hi @wakeful thanks a lot for this PR 🙏🏻 

As per the example you provided and looking at the code, I can notice that the attributes are not parsed as JSON but returned in a single string instead. I don't think that's the right way to handle this, `humanReadableAttributes` should be an object instead. Also make sure to use snake-cased JSON keys (e.g. `human_readable_attributes`). Keep in mind that JSON output is intended to be used by other softwares for processing purposes.

I'll review further when comments will be addressed.

wakeful:
thanks for looking at this @sundowndev, I updated the PR base on your suggestions.

moadibfr (commented):
Hi @wakeful. Thanks a lot for this PR !
I agree with @sundowndev and I think the revised version is better (especially I did not like having empty string when there was no human readable attributes).
I'm just wondering why do you need to change the case of the keys ?

wakeful:
hi @moadibfr, I started changing the key(s) name to keep it in the same `snake_case` form as the `partent` obj, also some of the extra attr are containing spaces, some are using `camelCase`.

e.q. outout without chaning the `case` on the key(s)
```json
{
"id": "sgrule-1180062852",
"type": "aws_security_group_rule",
"human_readable_attributes": {
	"Destination": "0.0.0.0/0",
	"Ports": "All",
	"Protocol": "All",
	"SecurityGroup": "sg-08e4f811ab524c960",
	"Type": "egress"
}},
{
"id": "rtbassoc-02bc8918dfcc806bd",
"type": "aws_route_table_association",
"human_readable_attributes": {
	"Table": "rtb-065a1aaa9b45900cf"
},
{
"id": "nacl-1588147895",
"type": "aws_network_acl_rule",
"human_readable_attributes": {
	"CIDR": "0.0.0.0/0",
	"Egress": "true",
	"Network": "acl-01d26c307d3b96d3e",
	"Protocol": "All",
	"Rule number": "100"
}}
```

eliecharra:
Thanks @wakeful for this very valuable PR and thanks a lot for @sundowndev and @moadibfr to handle that review 🙏🏻 
I agree with @moadibfr we probably do not want to change the case of the attributes, snake_case is the rule for the fixed fields in the JSON output but it's totally fine if we do not enforce anything for this dynamic part of the structure.

Can't wait to get it merged awesome work everyone ❤️

moadibfr:
> hi @moadibfr, I started changing the key(s) name to keep it in the same `snake_case` form as the `partent` obj (normalize), also some of the extra attr are containing spaces, some are using `camelCase`.

Yeah the point is that the key AND the value are supposed to be human readable and not really made to be processed that's why I was asking your use case. I think if at some point we'd like to have formatted key we should fixed the key instead of reformatting at the end.

wakeful:
@moadibfr, @eliecharra I reverted the dynamic change on the key, hopeflully this is ok now.

if you guys want I can `squash` the commits to reduce the amount of irrelevant `history`.

moadibfr:
> I figure maybe worth making sure we can parse the extra attr the same way, but I'm happy to revert that change :)

Yeah it sounds strange but it could make sens to have human readable attribute in the json output so a script/soft using this output could just display them to a human.

> if you guys want I can squash the commits to reduce the amount of irrelevant history.

I feel like 5 commit is a lot for this so I would be grateful if you could squash them yeah 🙇

moadibfr:
@all-contributors please add @wakeful for code

PR #1359, opened 2022-02-24
New --only-managed / --only-unmanaged flags

wbeuil:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | no
| 🚀 New feature?   | yes
| ⚠ Deprecations?   | no
| ❌ BC Break       | no
| 🔗 Related issues | none
| ❓ Documentation  | https://github.com/snyk/driftctl-docs/pull/223

## Description

This PR adds 2 new flags that will output a different summary at the end of the scan. 

The `--only-managed` flag should sum up what's under control only in your states:

```
Found missing resources:
  - vol-xxxx (aws_ebs_volume)
  From tfstate://terraform.tfstate
    - i-xxxxx (aws_instance.bar)
Found changed resources:
  From tfstate://terraform.tfstate
    - i-xxxxxx (aws_instance.foo):
        ~ instance_type: "t2.nano" => "t2.micro" (computed)
Found 4 resource(s)
 - 50% coverage
 - 2 resource(s) managed by Terraform
     - 1/2 resource(s) out of sync with Terraform state
 - 2 resource(s) found in a Terraform state but missing on the cloud provider
```

The `--only-unmanaged` flag should sum up what's NOT under control in your cloud provider:

```
Found resources not covered by IaC:
  aws_iam_policy_attachment:
    ....
  aws_iam_role:
    ....
  aws_iam_role_policy:
    ....
  aws_network_acl_rule:
    ....
Found 18 resource(s)
 - 11% coverage
 - 2 resource(s) managed by Terraform
 - 16 resource(s) not managed by Terraform
```

WARNING: I wonder if we should enforce to have only one of the three flags at a time `--only-managed` `--only-unmanaged` and `--deep`, once we agree on what I've coded.

### DIFF
--- a/pkg/analyser/analysis.go
+++ b/pkg/analyser/analysis.go
@@ -204,6 +204,10 @@ func (a *Analysis) SetAlerts(alerts alerter.Alerts) {
 	a.alerts = alerts
 }
 
+func (a *Analysis) SetOptions(options AnalyzerOptions) {
+	a.options = options
+}
+
 func (a *Analysis) Coverage() int {
 	if a.summary.TotalResources > 0 {
 		return int((float32(a.summary.TotalManaged) / float32(a.summary.TotalResources)) * 100.0)
--- a/pkg/analyser/analyzer.go
+++ b/pkg/analyser/analyzer.go
@@ -38,7 +38,9 @@ func (c *ComputedDiffAlert) ShouldIgnoreResource() bool {
 }
 
 type AnalyzerOptions struct {
-	Deep bool
+	Deep          bool
+	OnlyManaged   bool
+	OnlyUnmanaged bool
 }
 
 type Analyzer struct {
@@ -72,7 +74,9 @@ func (a Analyzer) Analyze(remoteResources, resourcesFromState []*resource.Resour
 		}
 
 		if !found {
-			analysis.AddDeleted(stateRes)
+			if !analysis.Options().OnlyUnmanaged {
+				analysis.AddDeleted(stateRes)
+			}
 			continue
 		}
 
@@ -80,8 +84,8 @@ func (a Analyzer) Analyze(remoteResources, resourcesFromState []*resource.Resour
 		filteredRemoteResource = removeResourceByIndex(i, filteredRemoteResource)
 		analysis.AddManaged(stateRes)
 
-		// Stop there if we are not in deep mode, we do not want to compute diffs
-		if !a.options.Deep {
+		// Stop there if we are not in deep mode or in only-managed mode, we do not want to compute diffs
+		if !a.options.Deep && !a.options.OnlyManaged {
 			continue
 		}
 
@@ -130,7 +134,9 @@ func (a Analyzer) Analyze(remoteResources, resourcesFromState []*resource.Resour
 	}
 
 	// Add remaining unmanaged resources
-	analysis.AddUnmanaged(filteredRemoteResource...)
+	if !analysis.Options().OnlyManaged {
+		analysis.AddUnmanaged(filteredRemoteResource...)
+	}
 
 	// Sort resources by Terraform Id
 	// The purpose is to have a predictable output
--- a/pkg/analyser/analyzer.go
+++ b/pkg/analyser/analyzer.go
@@ -84,8 +84,8 @@ func (a Analyzer) Analyze(remoteResources, resourcesFromState []*resource.Resour
 		filteredRemoteResource = removeResourceByIndex(i, filteredRemoteResource)
 		analysis.AddManaged(stateRes)
 
-		// Stop there if we are not in deep mode or in only-managed mode, we do not want to compute diffs
-		if !a.options.Deep && !a.options.OnlyManaged {
+		// Stop there if we are not in deep mode, we do not want to compute diffs
+		if !a.options.Deep {
 			continue
 		}
 
--- a/pkg/cmd/scan.go
+++ b/pkg/cmd/scan.go
@@ -228,6 +228,16 @@ func NewScanCmd(opts *pkg.ScanOptions) *cobra.Command {
 		configDir,
 		"Directory path that driftctl uses for configuration.\n",
 	)
+	fl.BoolVar(&opts.OnlyManaged,
+		"only-managed",
+		false,
+		"Report only what's managed by your IaC\n",
+	)
+	fl.BoolVar(&opts.OnlyUnmanaged,
+		"only-unmanaged",
+		false,
+		"Report only what's not managed by your IaC\n",
+	)
 
 	return cmd
 }
@@ -271,7 +281,7 @@ func scanRun(opts *pkg.ScanOptions) error {
 	logrus.Debug("Checking for driftignore")
 	driftIgnore := filter.NewDriftIgnore(opts.DriftignorePath, opts.Driftignores...)
 
-	scanner := remote.NewScanner(remoteLibrary, alerter, remote.ScannerOptions{Deep: opts.Deep}, driftIgnore)
+	scanner := remote.NewScanner(remoteLibrary, alerter, remote.ScannerOptions{Deep: opts.Deep, OnlyManaged: opts.OnlyManaged}, driftIgnore)
 
 	iacSupplier, err := supplier.GetIACSupplier(opts.From, providerLibrary, opts.BackendOptions, iacProgress, alerter, resFactory, driftIgnore)
 	if err != nil {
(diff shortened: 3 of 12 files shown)


eliecharra:
> WARNING: I wonder if we should enforce to have only one of the three flags at a time --only-managed --only-unmanaged and --deep, once we agree on what I've coded.

I still haven't read the PR yet, but here are my first thoughts

I think that it's pretty clear that `--only-managed` and `--only-unmanaged` should be mutually exclusives.

But I don't see why we are talking about deep mode here ? 
This is something else and it has nothing to do with manage or unmanaged notion. 
For sure in a perfect world we do not want to enable deep mode and detail fetching for unmanaged resources, but it's another performance topic that is [more complicated](https://github.com/snyk/driftctl/issues/916#issuecomment-999433328) that we thought initally.

wbeuil:
For the deep mode it's more about the fact that the `--only-managed`, from @sjourdan product point of view, should output drifted attributes. Thus, it does not make sense to do `driftctl scan --deep --only-managed`.

But I'm good having a simple `--only-managed` where deleted/managed resources are shown, and the `deep` mode will enable us to show drifted attributes.

eliecharra:
> Thus, it does not make sense to do driftctl scan --deep --only-managed.

I strongly disagree with that assumption.
People should be able to see missing resource without drift, exactly like they want to have only unmanaged.

We should keep those small notions separated, but if we want to address that kind of feature, a `--profile` flag may be more appropriate, and a profile may contains multiples switch to be able to run driftctl in a certain configuration: in our case both `--only-unmanaged` and `--deep`.

With `--only-managed` we are targeting a specific switch in the product (by the naming of the flag itself), we are not expecting it to do some magic under the hoods and activate other flags like deep mode, it will add a lot of mess IMO.

WDYT @sjourdan ?

pkg/remote/scanner.go
```diff
@@ -94,7 +95,7 @@ func (s *Scanner) scan() ([]*resource.Resource, error) {
 		return nil, err
 	}
 
-	if !s.options.Deep {
+	if !s.options.Deep && !s.options.OnlyManaged {
```
eliecharra, line 98: See https://github.com/snyk/driftctl/pull/1359#issuecomment-1050886424

pkg/analyser/analyzer.go
```diff
@@ -72,16 +75,18 @@ func (a Analyzer) Analyze(remoteResources, resourcesFromState []*resource.Resour
...
 		}
 
 		// Remove managed resources, so it will remain only unmanaged ones
 		filteredRemoteResource = removeResourceByIndex(i, filteredRemoteResource)
 		analysis.AddManaged(stateRes)
 
-		// Stop there if we are not in deep mode, we do not want to compute diffs
-		if !a.options.Deep {
+		// Stop there if we are not in deep mode or in only-managed mode, we do not want to compute diffs
+		if !a.options.Deep && !a.options.OnlyManaged {
```
eliecharra, line 89: See https://github.com/snyk/driftctl/pull/1359#issuecomment-1050886424

pkg/analyser/analyzer.go
```diff
@@ -130,7 +135,9 @@ func (a Analyzer) Analyze(remoteResources, resourcesFromState []*resource.Resour
 	}
 
 	// Add remaining unmanaged resources
-	analysis.AddUnmanaged(filteredRemoteResource...)
+	if analysis.Options().OnlyUnmanaged || !hasOneOnlyFlag {
```
eliecharra, line 138: `if !analysis.Options().OnlyManaged {` should be enough nope ?

pkg/analyser/analyzer.go
```diff
@@ -72,16 +75,18 @@ func (a Analyzer) Analyze(remoteResources, resourcesFromState []*resource.Resour
 		}
 
 		if !found {
-			analysis.AddDeleted(stateRes)
+			if analysis.Options().OnlyManaged || !hasOneOnlyFlag {
```
eliecharra, line 78: `if !analysis.Options().OnlyUnmanaged {` should be enough here nope ? 

I don't think we need `hasOneOnlyFlag` do we ?

pkg/remote/scanner.go
wbeuil: I'd love to wait for @sjourdan answer about that.

pkg/analyser/analysis.go
```diff
@@ -192,6 +192,10 @@ func (a *Analysis) SetAlerts(alerts alerter.Alerts) {
 	a.alerts = alerts
 }
 
+func (a *Analysis) SetOptions(options AnalyzerOptions) {
+	a.options = options
+}
+
```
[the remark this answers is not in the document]
wbeuil: `Unexported field 'options' usage` is the error, I could make options exportable just like we do with `ProviderName` or  `ProviderVersion`

pkg/analyser/analyzer.go
wbeuil: Absolutely right, good catch !

pkg/cmd/scan.go
```diff
@@ -218,6 +218,16 @@ func NewScanCmd(opts *pkg.ScanOptions) *cobra.Command {
 		configDir,
 		"Directory path that driftctl uses for configuration.\n",
 	)
+	fl.BoolVar(&opts.OnlyManaged,
+		"only-managed",
+		false,
+		fmt.Sprintf("%s Report only what's under control in your states\n", warn("EXPERIMENTAL:")),
```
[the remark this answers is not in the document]
wbeuil: That maybe to vague indeed, what about `Report only what's managed by your IaC`?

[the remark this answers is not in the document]
eliecharra: 👍🏻

sjourdan:
Thanks everyone for raising the points. 

Let me summarise the current state of discussion:

1. What's the use-case for `--only-managed`? Do users expect support for deleted resources only, or for changed resources also? That can be confirmed through more user testing if we want to double check user data already collected (current suggestion is based on feedback). `--only-managed` is supposed to target users who want to know if anything changed in any resource is managed by Terraform using many states, with the lowest access rights, outside of deployment pipelines, using a lightweight binary. 
2. Yes, if users want to only know about managed but deleted resources, then forcing them to use `--deep` is definitely stupid. AFAIK nobody reported such a need and I currently don't see a lot of value in just displaying deleted resources for managed resources.
3. Regardless of user expectations, what do we think is a good default for users in this situation: when `--only-unmanaged` does not need any more flag to be useful, we may offer a false sense of satisfaction for users thinking `--only-managed` is enough, when actually `--only-managed --deep` is needed to be really useful in this context. That said, both answers are acceptable!
4. A future `--profile` flag with combined configuration (one named profile expands to `--quiet --output="json://console" --only-managed --deep`). I love that idea.
 
Did I get the various threads right?

eliecharra:
> Regardless of user expectations, what do we think is a good default for users in this situation: when `--only-unmanaged` does not need any more flag to be useful, we may offer a false sense of satisfaction for users thinking `--only-managed` is enough, when actually `--only-managed --deep` is needed to be really useful in this context. That said, both answers are acceptable!

The problem is that enforcing `--deep` when using `--only-managed` can cause rate limit issues. 
At that point the flag `--only-managed` will not prevent use to enumerate the whole remote cloud account. It's only filtering results in the analysis step. 
By enabling `--deep` flag under the hood I'm afraid that it will bring a lot of frustration for users that may expect us only to read only resources from remote that are only defined in IaC (like tfplan). They may get some rate limits issues without really understanding why.

But we can document that and start gathering user feedback, but we have to keep in mind that the actual `--only-managed` flag is not really the same behavior than a Terraform plan, and may not work at all for users with a lot of mess in their remote (because of rate limits). In a more complete version we may want to avoid enumerating everything when `--only-managed` is used, but this will need more work and maybe starting gathering user feedback before is safer.

Anyway, if we decide to go with that, we need to switch the deep mode flag on in options parsing level when the `--only-managed` is passed instead than adding new conditions on existing `if deepModeEnabled` statements.

wbeuil:
> At that point the flag --only-managed will not prevent use to enumerate the whole remote cloud account. It's only filtering results in the analysis step.

This is indeed what's next in our todo list for this specific flag. The first iteration was on the output of it. And you're absolutely right about that we need to only read details of resources in our states and not everything.

> Anyway, if we decide to go with that, we need to switch the deep mode flag on in options parsing level when the --only-managed is passed instead than adding new conditions on existing if deepModeEnabled statements.

Good catch, didn't think it that way (probably because I didn't think it was possible) but if we all agree to go down this road, I'm going to change it real quick.

sundowndev:
I'm a bit confused, isn't this directly related to https://github.com/snyk/driftctl/discussions/272 ? Looks like this PR doesn't reflect the discussion in question. Or may be I missed another discussion somewhere else?

sjourdan:
> rate limit issues

Correct! Internal issue CFG-1599 was created to address this asap. 

> user feedback, documentation 

Agreed! Limitations need to be clear, roadmap to remediate too.

eliecharra (approved):

Queenbih87 (approved):
Approve

Issue #1358, opened 2022-02-23
html output on stdout shows progress and all the non html output


moadibfr:
**Description**
Output to stdout is not supposed to display progress and all the others warning/info (like provider version scan duration and everything).
This does work with `json://stdout`but does not with `html://stdout` thus you cannot pipe the output to a html file

**Environment**
 - OS: darwin
 - driftctl version: v0.19.0

**How to reproduce**
run a scan with ` scan --output html://stdout`. The output will show progress and other non html output.

patch-the-dog:
This issue has been referenced in the latest release v0.21.0.

PR #1360, opened 2022-02-28
Use voidprinter when sending HTML output to stdout

sundowndev:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | yes
| 🚀 New feature?   | no
| ⚠ Deprecations?   | no
| ❌ BC Break       | no
| 🔗 Related issues | #1358
| ❓ Documentation  | no 

## Description

We should not print anything but the scan output when we send output to stdout.


moadibfr (approved):
Thanks @sundowndev for fixing that. I made the issue for us to remember but the fix is quick so 👍

PR #1361, opened 2022-02-28
chore: include uncategorised changes in changelog

craigfurman:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | no
| 🚀 New feature?   | no
| ⚠ Deprecations?   | no
| ❌ BC Break       | no
| 🔗 Related issues | none
| ❓ Documentation  | internal maintainer docs - maybe

### Changelog

**chore: include uncategorised changes in changelog**

Ensure all PRs, even ones missing a "kind" label, are present in the
generated changelog. During the manually-initiated release process,
maintainers can edit the changelog to categorise them if they wish.

If we move to an automated, continuous release, it will still be better
to include uncategorised PRs than to omit them altogether.

### Notes

I can't find any maintainer docs around the manually-initiated parts of our release process, do we have any? ~~Either way, this PR as it stands now would require us to manually promote prereleases to releases after reviewing the changelog. Perhaps we don't want this part of the PR?~~ We can remove that commit if so.

**Update**: that commit was removed

### DIFF
--- a/scripts/changelog.sh
+++ b/scripts/changelog.sh
@@ -4,6 +4,23 @@
 # Please note that this script only work with Github repositories.
 # Prerequisites: git, github cli
 
+format_change() {
+    # First sed: extract title
+    # 2nd sed: hack to remove ugly `[]` from PRs with no labels
+    echo "$1" | sed "s/\[map\[$PARTITION_COLUMN.*//" | sed 's/ \[\]$//'
+}
+
+print_changelist() {
+  title=$1
+  shift
+  list=("$@")
+
+  echo -e "$title"
+  for change in "${list[@]}"; do
+    echo "$change"
+  done
+}
+
 GHCLI_BIN="gh"
 REPO="snyk/driftctl"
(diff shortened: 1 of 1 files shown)


sundowndev:
> Promotion from pre-release to release would be manual, so that watchers
would only get an email containing already-reviewed release notes,
giving maintainers an opportunity to re-categorise PRs if they wish.

I'm not familiar with GitHub's pre-release behavior, does it send an email when we put an existing release to ready for production ?

craigfurman:
Good question, I'm not actually sure. I was mostly trying to avoid making a premature announcement, while we're still shuffling release notes around. I'm happy to remove that commit if we'd rather consider the 2 commits separately?

sundowndev:
Well I don't think it's worth but curious to have more thoughts about this - WDYT @eliecharra ?

craigfurman:
@eliecharra that's fair, I've chopped off that commit. Would you mind re-reviewing?

eliecharra (approved):

PR #1362, opened 2022-02-28
Fix issue tagging script

sundowndev:
## Description

This PR fixes the issue tagging script. This script is intended to make users notice that a new release is available with a new patch or feature, as well as grouping tickets in milestones automatically. This script was tested locally and works as expected.

Note it passed all tests with shellcheck.

**fix: issue tagging script**

This commit fixes the grep expression that parses the related issue ID on pull requests. It is intended to ignore pull requests that don't mention any issue as well. The `|| true` forces exit code 0 when grep has no match, for some reason the `-q` option would not achieve this. It also fixes the milestone API endpoint missing a 's'. Finally, it uses $LATEST_TAG instead of  $BASE_TAG variable as the new version.


craigfurman (approved):
🤞

Issue #552, opened 2021-05-25
Automatically create a GH discussion along with a release


sjourdan:
**Description**

a) we want to collect as many user feedback as possible
b) we want to release value more often 
c) @mwarkentin suggested repeatedly (and rightly) to create a GH discussion linked to a release (it's new GH feature).

**Example**

0.8.0 example (manual): https://github.com/cloudskiff/driftctl/discussions/520

![image](https://user-images.githubusercontent.com/419121/119558109-4f22d800-bda1-11eb-80aa-a61754bee26b.png)

sjourdan:
didn't work for 0.10, let's find a solution for 0.11 then

sundowndev:
Goreleaser (#355) should allow us to do this since it can create draft releases

sjourdan:
We'll create the discussions manually until further notice.
+ we wonder if it's relevant with a high release cadence.

sundowndev:
Reopening as Goreleaser now supports automatic [release discussions](https://goreleaser.com/customization/release/?h=release+discussion#github).

craigfurman:
@sjourdan WDYT of a potential downside to doing this: opening an unfocussed discussion for every release would be for multiple issues to be discussed in one thread. It seems to me that the benefit (a), "collect as many user feedback as possible", is achievable through github discussions, issues, PRs, and discord.

Benefit (b), "release value more often", seems totally orthogonal to opening release discussions?

I'm not saying I'm necessarily against the idea, and I'll defer to longer-term maintainers in any case, but I do want to probe the "why" a little.

sjourdan:
to summarize the current state of discussions around this: 

- it was a great idea when we were releasing once every few weeks with a large scope of new features
- now we decided to release as soon as we merge something, it may create too much noise

thank you @craigfurman for raising the point!

PR #1365, opened 2022-03-01
Enable automatic release discussions

sundowndev:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | no
| 🚀 New feature?   | no
| ⚠ Deprecations?   | no
| ❌ BC Break       | no
| 🔗 Related issues | https://github.com/snyk/driftctl/issues/552
| ❓ Documentation  | no 

## Description

This PR adds a new option to goreleaser so it creates a new GitHub discussion for each new release.

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -6,6 +6,7 @@ release:
         name: driftctl
     draft: false
     prerelease: auto
+    discussion_category_name: Q&A
 before:
     hooks:
         - go generate ./...


craigfurman:
I'll of course defer to longer-term maintainers, but I think we should start with: why?

Specific issues, PRs, and discussions can already be raised. A potential downside of opening an unfocussed discussion for every release would be for multiple issues to be discussed in one thread. What would the benefit be?

PR #1368, opened 2022-03-03
Add div block for brand logo in HTML output

sundowndev:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | no
| 🚀 New feature?   | no
| ⚠ Deprecations?   | no
| ❌ BC Break       | no
| 🔗 Related issues | #...
| ❓ Documentation  | no 

## Description

[Related ticket (internal)](https://snyksec.atlassian.net/browse/CFG-1587)

**refactor(html): add div block for brand logo**

Add a div block that embeds the brand logo (here, driftctl logo) in the HTML output. The purpose here is to make the Snyk CLI able to easily replace that logo with their own branding. Also edit the brand logo style so the style doesn't resides in the SVG file anymore. The SVG now takes the width and height available, and the parent div has a static size of 100x81px. So when we will change the SVG file, it will not break the style.


eliecharra:
@sundowndev I don't have the full context of that, but if it's just changing stuff in the HTML report when launched under the snyk CLI, I am working on setting a `IS_SNYK` env var when launched from the snyk CLI.

This is a WIP, I pushed it here if you wanna see : https://github.com/snyk/driftctl/pull/1377/files#diff-2873f79a86c0d8b3335cd7731b0ecf7dd4301eb19a82ef7a1cba7589b5252261R37-R38

BTW If we want to do a snyk brand related modification, maybe we may want to avoid to do that in driftctl engine and maybe it can make more sense to implement a snyk branded html report in the snyk CLI directly ?

eliecharra:
> It was my initial thought, but @wbeuil pointed out that we might use the fmt command directly since it's only used by Snyk CLI. It's less complicated and doesn't bring too much complexity (vs. doing it in the CLI). WDYT?

IDK to be honest. But from product perspective I know that we want to avoid to do anything snyk-branded within driftctl, to be confirmed with @sjourdan

wbeuil:
Let's wait for @sjourdan point of view here, but my opinion was since we're pretty sure that this command will not last in the future, let's try to keep it simple to generate the snyk branded html output.

pkg/cmd/scan/output/assets/index.tmpl
```diff
@@ -11,7 +11,7 @@
 <div class="container">
     <header>
         <div class="div-left">
-            {{ .LogoSvg }}
+            <div id="brand_logo">{{ .LogoSvg }}</div>
```
wbeuil, line 14: Not a huge fan of using id just for css purposes. Any other reasons you are using an id here?

sundowndev-snyk: Yes, the purpose is to make this block unique so we can replace it in other tools.

wbeuil (approved):
LGTM

PR #1372, opened 2022-03-04
Add `driftctl fmt` command

eliecharra:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | no
| 🚀 New feature?   | yes
| ⚠ Deprecations?   | no
| ❌ BC Break       | no
| 🔗 Related issues | N/A
| ❓ Documentation  | We do not want to document that command

## Description

Add `driftctl fmt` command.
I moved output flag parsing method outside `scan.go` since it's now also shared with fmt, **nothing was changed in that diff**.


pkg/cmd/fmt.go
```diff
@@ -0,0 +1,68 @@
...
+
+func runFmt(opts *pkg.FmtOptions, reader io.Reader) error {
+
+	var analysisText []byte
+	scanner := bufio.NewScanner(reader)
+	for scanner.Scan() {
+		analysisText = append(analysisText, scanner.Bytes()...)
+	}
+
+	analysis := analyser.NewAnalysis(analyser.AnalyzerOptions{})
```
sundowndev, line 61: We don't use the Deep option here, this is going to cause the `--deep` flag being ignored, isn't it ?

eliecharra: We are only un-marshalling stdin to the struct here.
Whether if we set any option it will not change anything since we are not  running the analyzer, options will never be used.

sundowndev: Right, but for example we use the deep option in the console output to alter the result. This will not work with the fmt command since we never use it. For now the impact is very low, but someday we could have inconsistency between the scan and fmt command since it lacks of all the options. Dunno if it makes sense?

eliecharra: Hmm good point, we should then serialize also analysis options to be consistent.

I need to add that serialization in that PR, thanks for raising that 🙏🏻

sundowndev (approved):
Great job 👍🏻

Issue #1369, opened 2022-03-03
Add `scan_duration` in JSON output


eliecharra:
**Description**
We want to be able to retrieve the scan duration from the analysis and output it in the JSON output.

**Example**
```javascript
{
	"summary": {
		"total_resources": 6,
		"total_changed": 2,
		"total_unmanaged": 2,
		"total_missing": 2,
		"total_managed": 2
	},
        // Add this new field
	"scan_duration": 300 // In seconds
}
```

patch-the-dog:
This issue has been referenced in the latest release v0.23.0.

PR #1373, opened 2022-03-04
Add scan_duration json field

sundowndev:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | no
| 🚀 New feature?   | yes
| ⚠ Deprecations?   | no
| ❌ BC Break       | no
| 🔗 Related issues | https://github.com/snyk/driftctl/issues/1369
| ❓ Documentation  | yes 

## Description

>  We want to be able to retrieve the scan duration from the analysis and output it in the JSON output.

This PR adds a new JSON field to the scan output.

### DIFF
--- a/pkg/analyser/analysis.go
+++ b/pkg/analyser/analysis.go
@@ -64,6 +64,7 @@ type serializableAnalysis struct {
 	Alerts          map[string][]alerter.SerializableAlert `json:"alerts"`
 	ProviderName    string                                 `json:"provider_name"`
 	ProviderVersion string                                 `json:"provider_version"`
+	Duration        int                                    `json:"scan_duration"`
 }
 
 type GenDriftIgnoreOptions struct {
@@ -107,6 +108,7 @@ func (a Analysis) MarshalJSON() ([]byte, error) {
 	bla.Coverage = a.Coverage()
 	bla.ProviderName = a.ProviderName
 	bla.ProviderVersion = a.ProviderVersion
+	bla.Duration = int(a.Duration.Seconds())
 
 	return json.Marshal(bla)
 }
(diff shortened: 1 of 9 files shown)


pkg/analyser/analysis.go
```diff
@@ -64,6 +64,7 @@ type serializableAnalysis struct {
 	Alerts          map[string][]alerter.SerializableAlert `json:"alerts"`
 	ProviderName    string                                 `json:"provider_name"`
 	ProviderVersion string                                 `json:"provider_version"`
+	Duration        int                                    `json:"scan_duration"`
```
eliecharra, line 67: Can you add `omitempty` and name the field `ScanDuration` for consistency with the json field name ?

It's also an `uint`

eliecharra (changes requested):

go.mod
```diff
@@ -14,6 +14,7 @@ require (
 	github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/privatedns/armprivatedns v0.2.0
 	github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/resources/armresources v0.2.0
 	github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/storage/armstorage v0.2.0
+	github.com/Azure/azure-sdk-for-go/sdk/storage/azblob v0.2.0
```
eliecharra, line 17: Chan you revert thoses diff on the mod, there is no reason to change that here

eliecharra (changes requested):

go.mod
sundowndev: It's the result of a go mod tidy, it's probably that the go module file is still formatted with go1.16, not go1.17. I created a single commit for it but can revert it.

PR #1377, opened 2022-03-07
Improve snyk piping

eliecharra:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | no
| 🚀 New feature?   | no
| ⚠ Deprecations?   | no
| ❌ BC Break       | no
| 🔗 Related issues | none
| ❓ Documentation  | no

## Description

- Add `IS_SNYK` parameter to know whether we are used in snyk CLI
- Enable colorization when piped in snyk CLI
- Log diagnostic output to stderr (allow to get the progressbar when piped with snyk)

### DIFF
--- a/main.go
+++ b/main.go
@@ -17,6 +17,7 @@ import (
 	"github.com/snyk/driftctl/pkg/config"
 	"github.com/snyk/driftctl/pkg/version"
 	"github.com/snyk/driftctl/sentry"
+	"github.com/spf13/viper"
 )
 
 func init() {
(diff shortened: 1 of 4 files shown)


pkg/output/printer.go
```diff
@@ -23,7 +26,7 @@ func NewConsolePrinter() *ConsolePrinter {
 }
 
 func (c *ConsolePrinter) Printf(format string, args ...interface{}) {
-	fmt.Printf(format, args...)
+	_, _ = fmt.Fprintf(os.Stderr, format, args...)
```
sundowndev, line 29: Not sure to understand why we're using stderr here

main.go
```diff
@@ -33,12 +34,20 @@ func run() int {
 	config.Init()
 	logger.Init()
 	build := build.Build{}
+	// Check whether driftCTL is run under Snyk CLI
+	isSnyk := viper.GetBool("IS_SNYK")
 	logrus.WithFields(logrus.Fields{
 		"isRelease":               fmt.Sprintf("%t", build.IsRelease()),
 		"isUsageReportingEnabled": fmt.Sprintf("%t", build.IsUsageReportingEnabled()),
 		"version":                 version.Current(),
+		"is_snyk":                 fmt.Sprintf("%t", isSnyk),
```
sundowndev, line 43: You may keep the camel case format here, or rename other fields to snake case

pkg/output/printer.go
eliecharra: That allows to have progressbar in the snyk CLI.

stderr should be used to print diagnostic output, by doing that we can have progressbar and other diagnostic messages displayed to the user while maintaining a clean stdout with only drift results.

sundowndev (approved):

PR #1380, opened 2022-03-10
Fix panic in apigatewayv2_api_mapping enumerator

sundowndev:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | yes
| 🚀 New feature?   | no
| ⚠ Deprecations?   | no
| ❌ BC Break       | no
| 🔗 Related issues | #...
| ❓ Documentation  | no 

## Description

User on Discord recently reported a crash happening while scanning cloud resources. [Stack trace in crash report](https://sentry.io/organizations/snyk/issues/3079840329/?project=5568568&query=is%3Aunresolved) showed that it's the `apigatewayv2_mapping` enumerator using `nil` as values for `stage` and `api_id`.

### Stack trace

```
*errors.fundamental: A runner routine paniced: runtime error: invalid memory address or nil pointer dereference
  File "/root/project/pkg/parallel/parallel_runner.go", line 94, in (*ParallelRunner).Run.func1.1
  File "/root/project/pkg/remote/aws/apigatewayv2_mapping_enumerator.go", line 47, in (*ApiGatewayV2MappingEnumerator).Enumerate
  File "/root/project/pkg/remote/scanner.go", line 71, in (*Scanner).scan.func1
  File "/root/project/pkg/parallel/parallel_runner.go", line 97, in (*ParallelRunner).Run.func1
```

### DIFF
--- a/pkg/remote/aws/apigatewayv2_mapping_enumerator.go
+++ b/pkg/remote/aws/apigatewayv2_mapping_enumerator.go
@@ -38,15 +38,21 @@ func (e *ApiGatewayV2MappingEnumerator) Enumerate() ([]*resource.Resource, error
 			return nil, remoteerror.NewResourceListingError(err, string(e.SupportedType()))
 		}
 		for _, mapping := range mappings {
+			attrs := make(map[string]interface{})
+
+			if mapping.ApiId != nil {
+				attrs["api_id"] = *mapping.ApiId
+			}
+			if mapping.Stage != nil {
+				attrs["stage"] = *mapping.Stage
+			}
+
 			results = append(
 				results,
 				e.factory.CreateAbstractResource(
 					string(e.SupportedType()),
 					*mapping.ApiMappingId,
-					map[string]interface{}{
-						"stage":  *mapping.Stage,
-						"api_id": *mapping.ApiId,
-					},
+					attrs,
 				),
 			)
 		}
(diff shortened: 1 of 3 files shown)


pkg/remote/aws/apigatewayv2_mapping_enumerator.go
```diff
@@ -38,6 +39,15 @@ func (e *ApiGatewayV2MappingEnumerator) Enumerate() ([]*resource.Resource, error
 			return nil, remoteerror.NewResourceListingError(err, string(e.SupportedType()))
 		}
 		for _, mapping := range mappings {
+			if mapping.Stage == nil || mapping.ApiId == nil {
```
eliecharra, line 42: Are those attributes only used for displaying better human-readable attributes or are they also used in a middleware logic somewhere ? 

My point is that we should avoid skipping resources if we are only unable to retrieve "optional" attributes

wbeuil: Used only for human-readable attributes

eliecharra: Thanks @wbeuil super cool to have your API gateway knowledge here 🙏🏻

If it's not mandatory I think we should do a fix to ignore those attributes when we don't have them, but we should avoid to skip the resource, that will lead to inconsistency and thus false positive drifts

eliecharra (approved):

PR #1444, opened 2022-03-21
Skip invalid issues in issue tagging script

sundowndev-snyk:
## Description

This script recently failed because of failing to parse an issue ID in [this PR](https://github.com/snyk/driftctl/pull/1443). This change aims to better handle cases like this. Ignore the invalid id, print details in logs and exit a non-zero code at the end.


craigfurman:
@sundowndev-snyk I'm not understanding what's invalid about the referenced PR? Would a more correct and robust fix not be to fix our "related issues" logic to not pick false positives, as seems to have happened?

craigfurman:
Sure, that makes sense

craigfurman (approved):

PR #1452, opened 2022-03-30
Disable usage reporting for non release builds

eliecharra:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | kinda
| 🚀 New feature?   | no
| ⚠ Deprecations?   | no
| ❌ BC Break       | no
| 🔗 Related issues | none
| ❓ Documentation  | no

## Description

We should remove usage reporting when we are not in a release build.
That will prevent us to send fake telemetry while working on driftctl

### DIFF
--- a/build/build.go
+++ b/build/build.go
@@ -20,5 +20,5 @@ func (b Build) IsRelease() bool {
 }
 
 func (b Build) IsUsageReportingEnabled() bool {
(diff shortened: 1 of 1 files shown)


craigfurman (approved):
Smart. We probably want something similar for snyk.

Issue #1389, opened 2022-03-15
Support AWS resource: aws_s3_bucket_public_access_block


sjourdan:
**Description**

Add AWS support for `aws_s3_bucket_public_access_block` (TF)

patch-the-dog:
This issue has been referenced in the [v0.28.0 release](https://github.com/snyk/driftctl/releases/tag/v0.28.0).

PR #1455, opened 2022-03-31
feat: add `aws_s3_bucket_public_access_block`

eliecharra:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | no
| 🚀 New feature?   | yes
| ⚠ Deprecations?   | no
| ❌ BC Break       | no
| 🔗 Related issues | #1389 
| ❓ Documentation  | https://github.com/snyk/driftctl-docs/pull/231


pkg/middlewares/aws_s3_bucket_public_access_block_reconcilier.go
```diff
@@ -0,0 +1,80 @@
...
+//   - One resource should be unmanaged: the bucket (expected behavior)
+// - Go to the console and update public access block for that bucket
+// - Run the scan again
+//   - We should now have a new public access block resource unmanaged (expected)
+// - Now uncheck back all things in the public block you just updated
+// - Run the scan again
+//   - We still have the public block as unmanaged, this is NOT expected since all values are back to default
+//
+// This simple middleware is handling that edge case by removing resource that have every attribute set to false both
+// from resources from state and resources from remote.
```
wbeuil, line 32: I have a problem with this middleware.

Imagine I want to use S3 as a static website, as per [AWS documentation](https://docs.aws.amazon.com/AmazonS3/latest/userguide/website-hosting-custom-domain-walkthrough.html#root-domain-walkthrough-configure-bucket-permissions) I will set all Block Public Access settings to false just like below:

```
resource "aws_s3_bucket_public_access_block" "block" {
  bucket = aws_s3_bucket.test.id
  block_public_acls   = false
  block_public_policy = false
  ignore_public_acls = false
  restrict_public_buckets = false
}
```

With this middleware, each driftctl scan will by design ignore my managed resource. And thus, if someone tries to change a setting in the console, instead of having a drifted managed resource I will have a new unmanaged resource which I'm no expecting.

```
# change manually in the console and run driftctl scan again
Found resources not covered by IaC:
  aws_s3_bucket_public_access_block:
    - test-bucket-driftctl-snyk
```

wbeuil (changes requested):
Comments mostly on the middleware, the rest LGTM

pkg/middlewares/aws_s3_bucket_public_access_block_reconcilier.go
eliecharra: That's true, good catch I missed that one.
Maybe we can try to find an according block in IaC when iterating on remote ones, and we can remove it only if we get a match ? That could handle that case WDYT ?

wbeuil: I don't understand your thinking here, can you try with an example?

wbeuil (changes requested):

wbeuil (approved):
LGTM

PR #1464, opened 2022-04-05
Fix aws_ebs_encryption_by_default_reconciler middleware

sundowndev-snyk:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | yes
| 🚀 New feature?   | no
| ⚠ Deprecations?   | no
| ❌ BC Break       | no
| 🔗 Related issues | #...
| ❓ Documentation  | no 

## Description

A simple break statement in that middleware was preventing many remote resources to be included in the scan, resulting in many missing resources. Bug introduced in #1448 but not released yet. The reason why we didn't notice it is because we didn't test it with many remote resources, not even in unit tests.


wbeuil (approved):
Really good catch, note to myself always check with a bunch a resources without a driftignore file :)

Issue #1425, opened 2022-03-15
Support AWS resource: aws_ecr_repository_policy


sjourdan:
**Description**

Add AWS support for `aws_ecr_repository_policy` (TF)

patch-the-dog:
This issue has been referenced in the [v0.33.0 release](https://github.com/snyk/driftctl/releases/tag/v0.33.0).

PR #1494, opened 2022-05-10
feat: add aws `ecr_repository_policy` resource

almog27:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | no
| 🚀 New feature?   | yes
| ⚠ Deprecations?   | no
| ❌ BC Break       | no
| 🔗 Related issues | #1425 
| ❓ Documentation  | https://github.com/snyk/driftctl-docs/pull/240

## Description

This PR adds `ecr_repository_policy` resource as part of the AWS coverage


pkg/remote/aws/repository/ecr_repository_test.go
```diff
@@ -88,3 +88,54 @@ func Test_ecrRepository_ListAllRepositories(t *testing.T) {
...
+				).Return(&ecr.GetRepositoryPolicyOutput{
+					RegistryId:     aws.String("1"),
+					RepositoryName: aws.String("2"),
+				}, nil).Once()
+			},
+			want: &ecr.GetRepositoryPolicyOutput{
+				RegistryId:     aws.String("1"),
+				RepositoryName: aws.String("2"),
+			},
+		},
```
sundowndev-snyk, line 114: Looks like we're not testing errors here

pkg/remote/aws/repository/ecr_repository_test.go
```diff
@@ -88,3 +88,54 @@ func Test_ecrRepository_ListAllRepositories(t *testing.T) {
...
+		name    string
+		mocks   func(client *awstest.MockFakeECR)
+		want    *ecr.GetRepositoryPolicyOutput
+		wantErr error
+	}{
+		{
+			name: "Get repository policy",
+			mocks: func(client *awstest.MockFakeECR) {
+				client.On("GetRepositoryPolicy",
+					mock.Anything,
```
sundowndev-snyk, line 104: You can define a common fake repository above and use it here, instead of `mock.Anything`

sundowndev-snyk (changes requested):

eliecharra:
@almog27 This one seems to be blocked in review for a couple of days, is there something blocking us there ?

pkg/remote/aws/ecr_repository_policy_enumerator.go
```diff
@@ -0,0 +1,51 @@
...
+}
+
+func (e *ECRRepositoryPolicyEnumerator) SupportedType() resource.ResourceType {
+	return aws.AwsEcrRepositoryPolicyResourceType
+}
+
+func (e *ECRRepositoryPolicyEnumerator) Enumerate() ([]*resource.Resource, error) {
+	repos, err := e.repository.ListAllRepositories()
+	if err != nil {
+		return nil, remoteerror.NewResourceListingErrorWithType(err, string(e.SupportedType()), aws.AwsEcrRepositoryPolicyResourceType)
```
moadibfr, line 29:
```suggestion
		return nil, remoteerror.NewResourceListingErrorWithType(err, string(e.SupportedType()), aws.AwsEcrRepositoryResourceType)
```
You fail to read the ecr repository policies because it's impossible to list ecr repositories. The error constructeur first argument is the enumerator type and the second one is the one unable to list, in this case they are different.

moadibfr (changes requested):
one of the error is incorrect but everything else LGTM. You could fix that and rebase so we can merge this pr.
Also, we should probably write a ticket to support deepmode for this one as policy support without policy content is not that useful (cc @sjourdan )

moadibfr (dismissed):
👍

pkg/remote/aws/repository/ecr_repository.go
```diff
@@ -41,3 +42,12 @@ func (r *ecrRepository) ListAllRepositories() ([]*ecr.Repository, error) {
...
 }
+
+func (r *ecrRepository) GetRepositoryPolicy(repo *ecr.Repository) (*ecr.GetRepositoryPolicyOutput, error) {
+	var repositoryPolicyInput *ecr.GetRepositoryPolicyInput = &ecr.GetRepositoryPolicyInput{
+		RegistryId:     repo.RegistryId,
+		RepositoryName: repo.RepositoryName,
+	}
+
+	return r.client.GetRepositoryPolicy(repositoryPolicyInput)
+}
```
sundowndev-snyk, line 53: This method misses a cache hit, is this intentional ?

pkg/remote/aws/repository/ecr_repository_test.go
```diff
@@ -88,3 +91,73 @@ func Test_ecrRepository_ListAllRepositories(t *testing.T) {
...
+		{
+			name: "Get repository policy error",
+			mocks: func(client *awstest.MockFakeECR) {
+				client.On("GetRepositoryPolicy",
+					&ecr.GetRepositoryPolicyInput{
+						RegistryId:     aws.String("1"),
+						RepositoryName: aws.String("2"),
+					},
+				).Return(nil,
+					remoteerr.NewResourceListingError(dummyError, resourceaws.AwsEcrRepositoryPolicyResourceType),
```
sundowndev-snyk, line 131: The repository will never return such errors, this can be confusing. Can you simply use `dummyError` as is ?

pkg/remote/aws/repository/ecr_repository.go
sundowndev-snyk: Also I think we should use a more common return type like `[]*ecr.RepositoryPolicy` if such exists

pkg/remote/aws_ecr_scanner_test.go
```diff
@@ -125,3 +125,69 @@ func TestECRRepository(t *testing.T) {
...
+					RepositoryName: awssdk.String("test_ecr_repo_policy"),
+				}, nil)
+			},
+			assertExpected: func(t *testing.T, got []*resource.Resource) {
+				assert.Len(t, got, 1)
+
+				assert.Equal(t, got[0].ResourceId(), "test_ecr_repo_policy")
+			},
+			err: nil,
+		},
```
sundowndev-snyk, line 155: We should have more tests cases here to test the resource listing errors of the enumerator

sundowndev-snyk (changes requested):

pkg/remote/aws/repository/ecr_repository.go
moadibfr: IIRC it does not exist so out of pure laziness I suggested @almog27 we just leave the output type from the request 😅

pkg/remote/aws_ecr_scanner_test.go
moadibfr: Yeah the error test would make more sense here. But @almog27 is going to make another PR to support deepmode so I think it's better not spending too much time on this test that will be rewritten with goldenfile

sundowndev-snyk (dismissed):
LGTM :+1:

moadibfr (approved):

sundowndev-snyk (approved):

PR #1496, opened 2022-05-12
fix: weakness in eip enumerator

eliecharra:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | yes
| 🚀 New feature?   | no
| ⚠ Deprecations?   | no
| ❌ BC Break       | no
| 🔗 Related issues | none
| ❓ Documentation  | no

## Description

Sentry Issue: [DRIFTCTL-1S](https://sentry.io/organizations/snyk/issues/3266908553/?referrer=github_integration)

```
*errors.fundamental: A runner routine paniced: runtime error: invalid memory address or nil pointer dereference
  File "github.com/snyk/driftctl/pkg/parallel/parallel_runner.go", line 94, in (*ParallelRunner).Run.func1.1
  File "github.com/snyk/driftctl/pkg/remote/aws/ec2_eip_enumerator.go", line 39, in (*EC2EipEnumerator).Enumerate
  File "github.com/snyk/driftctl/pkg/remote/scanner.go", line 71, in (*Scanner).scan.func1
  File "github.com/snyk/driftctl/pkg/parallel/parallel_runner.go", line 97, in (*ParallelRunner).Run.func1
```

### DIFF
--- a/pkg/remote/aws_ec2_scanner_test.go
+++ b/pkg/remote/aws_ec2_scanner_test.go
@@ -232,7 +232,9 @@ func TestEC2Eip(t *testing.T) {
 			test:    "no eips",
 			dirName: "aws_ec2_eip_empty",
 			mocks: func(repository *repository.MockEC2Repository, alerter *mocks.AlerterInterface) {
-				repository.On("ListAllAddresses").Return([]*ec2.Address{}, nil)
+				repository.On("ListAllAddresses").Return([]*ec2.Address{
+					{}, // Test Eip without AllocationId because it can happen (seen in sentry)
+				}, nil)
 			},
 		},
 		{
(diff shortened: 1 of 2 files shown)


pkg/remote/aws_ec2_scanner_test.go
```diff
@@ -232,7 +232,9 @@ func TestEC2Eip(t *testing.T) {
 			test:    "no eips",
 			dirName: "aws_ec2_eip_empty",
 			mocks: func(repository *repository.MockEC2Repository, alerter *mocks.AlerterInterface) {
```
sundowndev-snyk, line 234: This test is for empty result, may be we should create a new test instead? Or at least rename this one.

sundowndev-snyk (approved):
Left 1 comment but still approving to unblock merge

pkg/remote/aws_ec2_scanner_test.go
eliecharra: I think it's OK because this is an empty case we're handling here, just another empty scenario we do not have considered.

Issue #1499, opened 2022-05-24
driftignore type ignore logic broken


eliecharra:
**Description**
Some type are ignored from the scan when they shouldn't

**Environment**
 - OS: any
 - driftctl version: **0.29.0**
 - terraform version: not relevant
 - terraform providers versions: not relevant

**How to reproduce**

Apply that terraform code:

```hcl
provider "aws" {
    region = "us-east-1"
}

terraform {
    required_providers {
        aws = "3.19.0"
    }
}

resource "aws_s3_bucket" "test" {
    bucket = "test-bucket-driftctl"
}
```

Then setup the following `.driftignore`:

```ignore
*
!aws_s3_bucket.test-*
```

Then run `driftctl scan`

We got that output:

```
Found 0 resource(s)
 - 0% coverage
Congrats! Your infrastructure is fully in sync.
```

We expect one resource to be returned here. The problem here is that the engine seems to be ignoring all s3 bucket.
The logic that check if we should fully ignore a type seems to be broken. We can confirm that by looking at the logs

```
DEBU[0002] Found module in state module= resourceCount=1
DEBU[0002] Ignored resource from state since it is ignored in filter name=test type=aws_s3_bucket
Scanned states (1) 
INFO[0002] Start scanning cloud provider
DEBU[0002] Ignored enumeration of resources since it is ignored in filter type=aws_s3_bucket
```

**Possible Solution**
We should not use the match method to determine if a type should be ignored

**Additional context**
That new test case should be passing

```diff
diff --git a/pkg/filter/driftignore_test.go b/pkg/filter/driftignore_test.go
index 61002cfb..8672fd7e 100644
--- a/pkg/filter/driftignore_test.go
+++ b/pkg/filter/driftignore_test.go
@@ -712,6 +712,20 @@ func TestDriftIgnore_IsTypeIgnored(t *testing.T) {
                        path:    "testdata/drift_ignore_all/.driftignore",
                        ignores: []string{"*", "!aws_s3*", "!aws_route53*"},
                },
+               {
+                       name: "do not ignore type when wildcard in resource ID",
+                       resources: []*resource.Resource{
+                               {
+                                       Type: "aws_iam_policy_attachment",
+                                       Id:   "foobar",
+                               },
+                       },
+                       want: []bool{
+                               false,
+                       },
+                       path:    "",
+                       ignores: []string{"*", "!aws_iam_policy_attachment.foo*"},
+               },
        }
        for _, tt := range tests {
[...]

PR #1501, opened 2022-05-24
fix: driftignore that ignore type

eliecharra:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | yes
| 🚀 New feature?   | no
| ⚠ Deprecations?   | no
| ❌ BC Break       | no
| 🔗 Related issues | #1499 
| ❓ Documentation  | no

## Description

In some case driftignore filtering engine was ignoring
some types event if some inclusion rules are defined.
This is because we do not handle the case where we use an
inclusion rule on a resource with an ID, the git matching logic
is not exactly what we want here.

⚠️ **Attention required by the reviewer** ⚠️
- Did I missed other case where the actual match logic is not appropriate ?
- Do you see any edge case from which that fix may introduce false positive inclusions ?

### DIFF
--- a/pkg/filter/driftignore.go
+++ b/pkg/filter/driftignore.go
@@ -90,7 +90,7 @@ func (r *DriftIgnore) parseIgnorePattern(line string, patterns *[]gitignore.Patt
 func (r *DriftIgnore) isAnyOfChildrenTypesNotIgnored(ty resource.ResourceType) bool {
 	childrenTypes := resource.GetMeta(ty).GetChildrenTypes()
 	for _, childrenType := range childrenTypes {
-		if !r.match(fmt.Sprintf("%s.*", childrenType)) {
+		if !r.shouldIgnoreType(childrenType) {
 			return true
 		}
 		if r.isAnyOfChildrenTypesNotIgnored(childrenType) {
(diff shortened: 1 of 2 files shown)


pkg/filter/driftignore.go
```diff
@@ -107,6 +107,18 @@ func (r *DriftIgnore) IsTypeIgnored(ty resource.ResourceType) bool {
...
+	return r.shouldIgnoreType(ty)
+}
+
+func (r *DriftIgnore) shouldIgnoreType(ty resource.ResourceType) bool {
+	for _, pattern := range r.ignorePatterns {
+		// If a line start with a `!` and if the type match, we should not ignore it
+		if strings.HasPrefix(pattern, "!") &&
+			strings.HasPrefix(pattern[1:], fmt.Sprintf("%s.", ty)) {
+			return false
+		}
```
sundowndev-snyk, line 119: Why not just do `if strings.HasPrefix(pattern, fmt.Sprintf("!%s.", ty)) {` here ? May be I'm missing something

sundowndev-snyk (changes requested):
Nothing much to say about this one, it just works :+1:  and it doesn't break my own usage.

pkg/filter/driftignore.go
eliecharra: True, let me fix that 👍🏻

PR #1514, opened 2022-06-02
fix: google_compute_global_forwarding_rule acc test

sundowndev-snyk:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | no
| 🚀 New feature?   | no
| ⚠ Deprecations?   | no
| ❌ BC Break       | no
| 🔗 Related issues | #...
| ❓ Documentation  | no 

## Description

This is an attempt to fix [a failure in our acceptance test](https://app.circleci.com/jobs/github/snyk/driftctl/10699).

### DIFF
--- a/pkg/resource/google/google_compute_global_forwarding_rule_test.go
+++ b/pkg/resource/google/google_compute_global_forwarding_rule_test.go
@@ -2,6 +2,7 @@ package google_test
 
 import (
 	"testing"
+	"time"
 
 	"github.com/snyk/driftctl/test"
 	"github.com/snyk/driftctl/test/acceptance"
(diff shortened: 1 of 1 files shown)


pkg/resource/google/google_compute_global_forwarding_rule_test.go
```diff
@@ -17,6 +18,8 @@ func TestAcc_Google_ComputeGlobalForwardingRule(t *testing.T) {
 		},
 		Checks: []acceptance.AccCheck{
 			{
+				// New resources are not visible immediately through GCP API after an apply operation.
+				ShouldRetry: acceptance.LinearBackoff(10 * time.Minute),
```
karniwl, line 22: is that 10 minutes? quite long isn't it? from your experience it makes sense?

karniwl (approved):

pkg/resource/google/google_compute_global_forwarding_rule_test.go
eliecharra: We could probably switch it to 5min, but it's not that important because if in real life if it took less we'll get out of this without waiting 10 mins.
It's more like if we can do more we can do less in that case.

karniwl: cool sounds great 👍

PR #1545, opened 2022-06-15
Fix false-positive on timeouts attribute in aws_route_table and aws_default_route_table

chdorner-snyk:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | yes
| 🚀 New feature?   |no
| ⚠ Deprecations?   |no
| ❌ BC Break       |no
| 🔗 Related issues | CFG-1974
| ❓ Documentation  | no

## Description

Remove `timeout` attributes for `aws_route_table` and `aws_default_route_table` resources to prevent false-positive drift.
Note that the `timeout` block for these two Terraform resources have only been added in AWS provider version 3.62.0


pkg/iac/terraform/state/test/aws_default_route_table/terraform.tfstate
```diff
@@ -0,0 +1,108 @@
...
+          "private": "eyJlMmJmYjczMC1lY2FhLTExZTYtOGY4OC0zNDM2M2JjN2M0YzAiOnsiY3JlYXRlIjozNjAwMDAwMDAwMDAsInVwZGF0ZSI6MTgwMDAwMDAwMDAwfX0=",
+          "dependencies": [
+            "aws_default_vpc.default"
+          ]
+        }
+      ]
+    },
+    {
+      "mode": "managed",
+      "type": "aws_default_vpc",
```
eliecharra, line 41: You can remove everything but aws_route_table resource from that state file. It will be an inconsistent one regarding Terraform logic but we do not care here it's only to ensure unmarshalling is going well !
We try to minimise test maintenance by removing side effects 😉 

You can also remove the lock file from that folder, it is not used.

You'll also have to update the result file.

pkg/resource/aws/aws_route_table_test.go
```diff
@@ -9,7 +9,7 @@ import (
 
 func TestAcc_Aws_RouteTable(t *testing.T) {
 	acceptance.Run(t, acceptance.AccTestCase{
-		TerraformVersion: "0.15.5",
+		TerraformVersion: "1.0.2",
```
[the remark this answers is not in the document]
eliecharra: I think it's perfectly fine if you update test as you reach them for M1 support 👍🏻 
We do not want a specific version of Terraform here, it's just to pin a version and prevent flackyness in tests due to Terraform changes.

I'm totally happy with that no worries 🙏🏻

eliecharra (changes requested):
Thanks a lot for that PR that looks awesome 🎉 

I only have small remark on IaC test 

Super good job !

pkg/iac/terraform/state/test/aws_default_route_table/terraform.tfstate
eliecharra: > You can also remove the lock file from that folder, it is not used.

☝🏻 `pkg/iac/terraform/state/test/aws_default_route_table/.terraform.lock.hcl` need to be removed too

eliecharra (approved):
Sounds great 🎉  ! Thanks a lot @chdorner-snyk 🙏🏻

PR #1546, opened 2022-06-16
chore: Stop failing issue tagging on invalid issue links

chdorner-snyk:
## Description

The script was detecting JIRA issue links as GH issues and subsequently failing when checking that a GH issue with the same ID exists (example of a failure [here](https://app.circleci.com/pipelines/github/snyk/driftctl/4640/workflows/02c81b9b-96d1-4e0a-af16-a0b84da03316/jobs/10958?invite=true#step-105-5)).

Given we seem to have a mix of GH and JIRA issue links, we can just not fail the job if no GH issue can be found for a certain ID.

### DIFF
--- a/scripts/issue-tagging.sh
+++ b/scripts/issue-tagging.sh
@@ -35,7 +35,6 @@ echo "Finding merged pull requests between $BASE_TAG and $LATEST_TAG..."
 PRs=$(git log --pretty=oneline "$BASE_TAG"..."$LATEST_TAG" | grep 'Merge pull request #' | grep -oE '#[0-9]+' | sed 's/#//')
 
 # Find fixed issues from $BASE_TAG to $LATEST_TAG
-EXIT_CODE=0
 ISSUES=()
 for pr in $PRs; do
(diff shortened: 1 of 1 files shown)


scripts/issue-tagging.sh
```diff
@@ -38,7 +38,7 @@ PRs=$(git log --pretty=oneline "$BASE_TAG"..."$LATEST_TAG" | grep 'Merge pull re
 EXIT_CODE=0
 ISSUES=()
 for pr in $PRs; do
-    id=$($GHCLI_BIN pr view "$pr" --json body | grep -oE '(Related issues \| )(.*)?[0-9]+(.*|\r|\n)?(\|)' | sed 's/[^[:digit:]]//g' | sed -z 's/\n//g' || true)
+    id=$($GHCLI_BIN pr view "$pr" --json body | grep -oE '(Related issues \| )(.*)?#[0-9]+(.*|\r|\n)?(\|)' | sed 's/[^[:digit:]]//g' | sed -z 's/\n//g' || true)
```
sundowndev-snyk, line 41: We used to match github issues that way, but it turns out sometimes we put the complete URL instead of just the ID. To match both we don't look for the `#`. I'd suggest we just don't make the script fail when the parsed ID isn't a Github issue. WDYT?

sundowndev-snyk (approved):
LGTM, thanks for spending some time on this :pray:

PR #1554, opened 2022-06-29
Extract enumeration to it's own module

moadibfr:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | no
| 🚀 New feature?   | no
| ⚠ Deprecations?   | no
| ❌ BC Break       | no
| 🔗 Related issues |
| ❓ Documentation  | no

## Description

First step to extract enumeration as a new library. Enumeration is now in a package with no dependencies to driftctl package.
- Acc test for aws are passing I did not tested for gcp and azure due to setup issues.
- Alerter is now inside enumeration which is wrong but when we rework the api/arch we should be able to get it back inside driftctl.
- Due to the fact we do not normalize resources attribute, some scanner test goldenfile changed name or content.
- Also I had to change enumerator for ec2_network_rules and vpc_security_group_rules to give them an ID at the enumeration step instead of waiting for the normalisation. It seems that the ID did not change thought I'm not sure if in some case it could change...


craigfurman:
Great that we're doing this, this is going to be super useful. I do have a few questions though.

1. What is the actual API to use here? Do clients instantiate NewScanner()?
   2. If so, can we put a facade in this package that looks somewhat like `Enumerate(resourceType string) ([]Resource, error)`, that encapsulates any no-op Alerters and whatever else a Scanner needs for the driftctl-scan use-case, that may be less useful in an enumeration library?
3. Can we filter resources by type?
   4. Preferably preventing them from being enumerated at all, rather than discarded in-memory. More like driftignore than `--filter`, in driftctl-scan terms.

pkg/resource/aws/aws_route.go
```diff
@@ -44,38 +25,4 @@ func initAwsRouteMetaData(resourceSchemaRepository resource.SchemaRepositoryInte
...
-		}
-		if prefix := val.GetString("destination_prefix_list_id"); prefix != nil && *prefix != "" {
-			attrs["Destination"] = *prefix
-		}
-		return attrs
-	})
-	resourceSchemaRepository.SetFlags(AwsRouteResourceType, resource.FlagDeepMode)
-}
-
-func CalculateRouteID(tableId, CidrBlock, Ipv6CidrBlock, PrefixListId *string) string {
```
eliecharra, line 67: I think this is used in a route expander middleware nope ? In my mind in the future the enum lib will mostly be under an `internal/` pkg and will only expose List and Refresh methods. We should avoid to add too many coupling, it's IMO OK to have duplicated code here for that kind of stuff.

pkg/resource/init_metadatas.go
```diff
@@ -0,0 +1,29 @@
...
+	"github.com/pkg/errors"
+	"github.com/snyk/driftctl/enumeration/remote/common"
+	"github.com/snyk/driftctl/enumeration/resource"
+	"github.com/snyk/driftctl/pkg/resource/aws"
+	"github.com/snyk/driftctl/pkg/resource/azurerm"
+	"github.com/snyk/driftctl/pkg/resource/github"
+	"github.com/snyk/driftctl/pkg/resource/google"
+)
+
+func InitMetadatas(remote string,
```
eliecharra, line 13: This is already done in `remote.Activate()`, is that really usefull to call that manually ?

eliecharra (dismissed):
Thanks for that awesome job 🎉 

I have on an aside note a lot of things to be fixed, but that could be done in other PR.
That one is bloated by tons of changes in pkg names, I tried to fix some but this still could lead to merge conflicts (I already rebased it an fixed two of them)

I've also let some comment about metadata, but I'm happy if we decide to merge that one and continue cleaning up the extraction in separated PRs.

TL;DR : nothing critical in my comments to the point that prevent a merge, so feel free to merge it ASAP if you feel comfortable with it @moadibfr

pkg/resource/aws/aws_appautoscaling_target.go
```diff
@@ -1,25 +1,12 @@
...
-			"scalable_dimension": *res.Attributes().GetString("scalable_dimension"),
-		}
-	})
-	resourceSchemaRepository.SetHumanReadableAttributesFunc(AwsAppAutoscalingTargetResourceType, func(res *resource.Resource) map[string]string {
-		attrs := make(map[string]string)
-		if v := res.Attributes().GetString("scalable_dimension"); v != nil && *v != "" {
-			attrs["Scalable dimension"] = *v
-		}
-		return attrs
-	})
```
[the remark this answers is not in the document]
moadibfr: yeah we already had a chat about this. I was wondering where it fitted the best and decided not to move them (cause yeah how I did was move the whole package and then move back what should not be in the enum,eration package)
I think in the end I agree with that and we can move that back to driftctl. I think in the end enumeration will have no schema and metadata anyway...

pkg/resource/aws/aws_appautoscaling_target.go
```diff
@@ -1,25 +1,12 @@
...
+	"github.com/snyk/driftctl/enumeration/resource/aws"
+)
 
 func initAwsAppAutoscalingTargetMetaData(resourceSchemaRepository resource.SchemaRepositoryInterface) {
-	resourceSchemaRepository.SetResolveReadAttributesFunc(AwsAppAutoscalingTargetResourceType, func(res *resource.Resource) map[string]string {
-		return map[string]string{
-			"service_namespace":  *res.Attributes().GetString("service_namespace"),
-			"scalable_dimension": *res.Attributes().GetString("scalable_dimension"),
-		}
-	})
```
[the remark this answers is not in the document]
moadibfr: Yeah We should get rid of this but I did not wanted to make to many change in this PR and just move things around

pkg/resource/init_metadatas.go
moadibfr: this is not the same init. This one will add normalize function that only exists in driftctl so it cannot be call from the enumeration function

pkg/resource/aws/aws_route.go
moadibfr: not sure about this one but there are some place where we compute ID and that may be needed for driftctl too when middleware create it too.
I think I checked that it was used in the enum before moving it but I might be wrong for this one...

PR #1557, opened 2022-07-05
fix: Better error messages for AWS auth

eliecharra:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | yes and no 😅 
| 🚀 New feature?   | yes and no 😅 
| ⚠ Deprecations?   | no
| ❌ BC Break       | no
| 🔗 Related issues | CFG-2002
| ❓ Documentation  | no

## Description

I disabled cred validation on the provider side, and instead I added a call to `sts:GetCallerIdentity` on our side to be able to handle auth failure errors properly

**Before**
Screenshot from 2022-07-05 12-00-38

**After**
Screenshot from 2022-07-05 11-59-26

### DIFF
--- a/pkg/remote/aws/provider.go
+++ b/pkg/remote/aws/provider.go
@@ -3,7 +3,9 @@ package aws
 import (
 	"github.com/aws/aws-sdk-go/aws/credentials"
 	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/sts"
 	"github.com/pkg/errors"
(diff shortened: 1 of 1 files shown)


moadibfr (approved):

wbeuil (approved):
LGTM, I think this is the only/best way to validate AWS creds even for CLI calls

`aws sts get-caller-identity`

eliecharra:
Thanks a lot @wbeuil 🙏🏻

PR #1562, opened 2022-07-11
Support tfstate discovering within workspaces

sundowndev-snyk:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | yes
| 🚀 New feature?   | no
| ⚠ Deprecations?   | no
| ❌ BC Break       | no
| 🔗 Related issues | [CTX-145](https://snyksec.atlassian.net/browse/CTX-145)
| ❓ Documentation  | yes 

## Description

The current feature that discovers state from HCL files does not support non-default workspaces. This PR intend to fix that.


pkg/terraform/hcl/backend_test.go
```diff
@@ -1,78 +1,125 @@
...
+func TestHCL_getCurrentWorkspaceName(t *testing.T) {
+	cases := []struct {
+		name string
+		dir  string
+		want string
+	}{
+		{
+			name: "test with default workspace",
+			dir:  "testdata/default_workspace",
+			want: "default",
```
eliecharra, line 20: That test case overlaps with `test with non-default workspace`. It test exactly the same code. Does it really make sense to maintain it ?

pkg/terraform/hcl/backend.go
```diff
@@ -11,25 +11,26 @@ import (
...
 }
 
-func (b BackendBlock) SupplierConfig() *config.SupplierConfig {
+func (b BackendBlock) SupplierConfig(workspace string) *config.SupplierConfig {
 	switch b.Name {
 	case "local":
 		return b.parseLocalBackend()
 	case "s3":
-		return b.parseS3Backend()
+		return b.parseS3Backend(workspace)
```
eliecharra, line 31: Using the workspace in the path of the state is used only for S3 and GCS, not azure ?

pkg/terraform/hcl/backend.go
```diff
@@ -47,25 +48,29 @@ func (b BackendBlock) parseLocalBackend() *config.SupplierConfig {
...
 
-func (b BackendBlock) parseS3Backend() *config.SupplierConfig {
+func (b BackendBlock) parseS3Backend(ws string) *config.SupplierConfig {
 	if b.Bucket == "" || b.Key == "" {
 		return nil
 	}
+	keyPrefix := ws
+	if keyPrefix == DefaultStateName {
+		keyPrefix = ""
+	}
```
eliecharra, line 58: We only have this behavior for S3 ?

eliecharra (changes requested):
I miss some context so I asked a few questions.

And a testcase seems useless to me

pkg/terraform/hcl/backend.go
sundowndev-snyk: I tested it and indeed, it is also used by azurerm!

pkg/terraform/hcl/backend_test.go
sundowndev-snyk: It's not the same code since the workspace is different

eliecharra: Yes I know, but does that really cover more code than the `test with non-default workspace` case ?
We can put anything in the content of the environment file, the covered code is always the same. 
I'm not sure to see what statements this case additionally covers.

PR #1565, opened 2022-07-19
add a facade to implement enum library api

moadibfr:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | no
| 🚀 New feature?   | yes
| ⚠ Deprecations?   | no
| ❌ BC Break       | no
| 🔗 Related issues | #...
| ❓ Documentation  | yes 

## Description

Add a facade (not fully implemented for the diagnostics mainly) so we can begin to use correct api for enumerating resources.

### DIFF
--- a/enumeration/enumerator/cloud_enumerator.go
+++ b/enumeration/enumerator/cloud_enumerator.go
@@ -1,0 +1,269 @@
+package enumerator
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"sync"
+
+	"github.com/snyk/driftctl/enumeration"
+
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+	"github.com/snyk/driftctl/enumeration/alerter"
+	"github.com/snyk/driftctl/enumeration/parallel"
+	"github.com/snyk/driftctl/enumeration/remote"
+	"github.com/snyk/driftctl/enumeration/remote/common"
+	"github.com/snyk/driftctl/enumeration/resource"
+	"github.com/snyk/driftctl/enumeration/terraform"
+)
+
+type CloudEnumerator struct {
+	alerter              *sliceAlerter
+	progress             enumeration.ProgressCounter
+	remoteLibrary        *common.RemoteLibrary
+	providerLibrary      *terraform.ProviderLibrary
+	enumeratorRunner     *parallel.ParallelRunner
+	detailsFetcherRunner *parallel.ParallelRunner
+	to                   string
+}
+
+type cloudEnumeratorBuilder struct {
+	cloud           string
+	providerVersion string
+	configDirectory string
+}
+
+// WithCloud Choose which cloud to use for enumeration and refresh
+// TODO could be inferred with types listed
+func (b *cloudEnumeratorBuilder) WithCloud(cloud string) *cloudEnumeratorBuilder {
+	b.cloud = cloud
+	return b
+}
+
+// WithProviderVersion optionally choose the provider version used for refresh
+func (b *cloudEnumeratorBuilder) WithProviderVersion(providerVersion string) *cloudEnumeratorBuilder {
+	b.providerVersion = providerVersion
+	return b
+}
+
+// WithConfigDirectory optionally choose the directory used to download terraform provider used for refresh
+func (b *cloudEnumeratorBuilder) WithConfigDirectory(configDir string) *cloudEnumeratorBuilder {
+	b.configDirectory = configDir
+	return b
+}
+
+func (b *cloudEnumeratorBuilder) Build() (*CloudEnumerator, error) {
+	enumerator := &CloudEnumerator{
+		enumeratorRunner:     parallel.NewParallelRunner(context.TODO(), 10),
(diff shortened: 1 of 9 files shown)


enumeration/enumerator/resource_enumerator.go
```diff
@@ -0,0 +1,229 @@
...
+}
+
+func NewResourceEnumerator() *ResourceEnumerator {
+	return &ResourceEnumerator{
+		enumeratorRunner:     parallel.NewParallelRunner(context.TODO(), 10),
+		detailsFetcherRunner: parallel.NewParallelRunner(context.TODO(), 10),
+	}
+}
+
+func (e *ResourceEnumerator) Init(to, providerVersion, configDirectory string) error {
```
craigfurman, line 36: If I understand correctly, for resource enumerator use cases (as opposed to drift detection), we still have some un-necessary parameters here. Suggestions:

`to`: can we infer this from the resource type? Or possibly as a stepping stone to avoid adding scope to this PR: remove the config type. "aws+tf" looks weird for resource enumeration, so how about just "aws"?

`providerVersion`: always "" for enumeration-only cases, we don't care about matching provider versions to a config

`configDirectory`: can we make "" automatically find a tempdir using `os.TempDir`?

We could do all of this with a sort of builder pattern:

```go
enumerator, err := NewResourceEnumerator().
  WithConfigDir(). // optional, "" finds a temp dir automatically
  WithCloud("aws").
  WithConfigType("tf"). // not for enumerator use-cases, but maybe for 
  Build()
```

In the driftctl-scan flow, meaningful parameters would be assigned, e.g. providerVersion from a flag if requested.

This is a pretty hasty sketch of mine, I haven't put a ton of thought into it, so don't take it as gospel!

enumeration/enumerator/resource_enumerator.go
```diff
@@ -0,0 +1,229 @@
...
+				}
+			}
+		case <-runner.DoneChan():
+			break loop
+		}
+	}
+	return results, runner.Err()
+}
+
+func (e *ResourceEnumerator) List(typ string) ([]*resource.Resource, error) {
```
craigfurman, line 167: Perfect! :chefs-kiss-emoji:

craigfurman (commented):
I didn't do a deep code review, but interface-wise, this is extremely close to what I was hoping for.

I would approve it, but want to wait someone closer to the work to do a proper code review, if that's ok.

I left a note about some potential facade improvements to `Init()`, removing parameters that are not necessary for enumeration-only flows (as opposed to driftctl-scan), but this should not be considered blocking for the PR. We can always change it later (even in a breaking way).

eliecharra:
> I would approve it, but want to wait someone closer to the work to do a proper code review, if that's ok.

Yeah that makes a lot of sense. I was supposed to look at this on Friday, but laptop issues ... I'm gonna check that today 
Thanks for your interface review anyway 🙏🏻

eliecharra (changes requested):

enumeration/enumerator/cloud_enumerator.go
```diff
@@ -0,0 +1,266 @@
...
+func (b *cloudEnumeratorBuilder) WithCloud(cloud string) *cloudEnumeratorBuilder {
+	b.cloud = cloud
+	return b
+}
+
+// WithProviderVersion optionally choose the provider version used for refresh
+func (b *cloudEnumeratorBuilder) WithProviderVersion(providerVersion string) *cloudEnumeratorBuilder {
+	b.providerVersion = providerVersion
+	return b
+}
```
eliecharra, line 47: Those two parameters are required (we could make the cloud one optional if we can detect it from the resource type list)

Having them in the builder pattern allow use to create an invalid struct if we do that:

```go
NewCloudEnumerator().Build()
```

Even for the enumeration without refresh usecase it will fail since we always call the `init()` func in the `Build` method.

I think we should add a check in Build method to avoid calling `init()` if we don't have both cloud and version parameters set.

Or maybe also create a `NewCloudEnumerator().WithRefresh("aws", "4.4.0").Build()` could be more explicit ?

Let's also add a check in the refresh to prevent the execution if we instanciate a CloudEnumerator without refresh capability.

Happy to pair on that BTW

eliecharra (changes requested):

enumeration/enumerator/cloud_enumerator.go
moadibfr: > Those two parameters are required (we could make the cloud one optional if we can detect it from the resource type list)

version is not required. the provider init will take a default one when it's not provided (https://github.com/snyk/driftctl/blob/cf9dba0712c76a593857684fa9d7b355140c29c3/enumeration/remote/aws/provider.go#L51)
I could check that the cloud is not empty but it should already return an error from init when the cloud is empty or not supported.

>Let's also add a check in the refresh to prevent the execution if we instanciate a CloudEnumerator without refresh capability.

For now there is no real mean to instantiate that as I need to call activate (and it create provider instance) to instantiate the enumerators. Hence I'm not sure we need such a check until we change that ?

eliecharra (approved):

PR #1572, opened 2022-07-28
Extract metadata from enumeration pkg

eliecharra:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | no
| 🚀 New feature?   | no
| ⚠ Deprecations?   | no
| ❌ BC Break       | no
| 🔗 Related issues | CTX-323
| ❓ Documentation  | no

## Description

Metadata are now in driftctl pkg.
I removed completely func `SetResolveReadAttributesFunc`, attributes set in the enumerator are now used as it byt the detail fetcher. 
The main issue was that the detail fetcher was only using string attributes in ReadResource, so I added some casting logic in it (see latest commits on tests fix).

This had some side effects on following resources:
- `aws_network_acl_rule` (and by heritance we also had impact on `aws_network_acl` too)
- `aws_cloudformation_stack`
- `aws_security_group_rule`

I added a normalization pass on driftctl side after the remote scan to apply normalization. No now we retrieve resources then normalize them instead of doing the normalization directly while retrieving resources. That's why now there is two resource factories, one in enum pkg that only create the struct but without any modification. We got another factory in drifctl that is applying normalization.

All golden files rename can be ommited in the review, it's a side effect to have more attributes fields. Because now we are not only using fields from `SetResolveReadAttributesFunc`. That should not cause any issues to retrieve details if we send additional (useless) fields to the `ReadResource` call.


enumeration/remote/aws/s3_bucket_analytic_enumerator.go
```diff
@@ -69,7 +70,7 @@ func (e *S3BucketAnalyticEnumerator) Enumerate() ([]*resource.Resource, error) {
 					string(e.SupportedType()),
 					id,
 					map[string]interface{}{
-						"region": region,
+						"alias": region,
```
moadibfr, line 73: no sure I understand this change ˆˆ

moadibfr (changes requested):

elie.tf
```diff
@@ -0,0 +1,23 @@
+provider "aws" {
+    region = "us-east-1"
```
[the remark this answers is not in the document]
eliecharra: Oh yeah good catch thanks a lot

enumeration/remote/aws/s3_bucket_analytic_enumerator.go
eliecharra: If you read the previous func that defined read attributes for those s3 related resources, we were retrieviing the regions attribute and setting it to the "alias" field. Since we do not have this method anymore, I needed to change it there directly.

image

Does it answer your question ?

moadibfr (approved):

PR #1576, opened 2022-08-04
fix: do not panic on aws provider init

moadibfr:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | yes
| 🚀 New feature?   | no
| ⚠ Deprecations?   | no
| ❌ BC Break       | no
| 🔗 Related issues | #...
| ❓ Documentation  | no 

## Description

Remove the must so we do not panic on aws session creation and forward the error.

### DIFF
--- a/enumeration/remote/aws/provider.go
+++ b/enumeration/remote/aws/provider.go
@@ -63,9 +63,14 @@ func NewAWSTerraformProvider(version string, progress enumeration.ProgressCounte
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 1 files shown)


eliecharra (dismissed):
Good move 👍🏻 Did you also tested the driftctl behavior with this change ?

craigfurman (approved):

moadibfr:
> Good move 👍🏻 Did you also tested the driftctl behavior with this change ?

yeah there should already be an explicit test that you added. But anyway if it's not handled it will just bubble up to the command error printing so it might still be better than a panic

PR #1579, opened 2022-08-05
fea: transform alert from enum lib into diagnostic

moadibfr:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | no
| 🚀 New feature?   | no
| ⚠ Deprecations?   | no
| ❌ BC Break       | no
| 🔗 Related issues | 
| ❓ Documentation  | no 

## Description

transform alert from enum lib into diagnostic when using the facade implementing the new api


enumeration/alerter/alert.go
```diff
@@ -22,6 +27,10 @@ func (f *FakeAlert) ShouldIgnoreResource() bool {
 	return f.IgnoreResource
 }
 
+func (f *FakeAlert) Resource() *resource.Resource {
```
craigfurman, line 30: Is this FakeAlert still used anywhere?

enumeration/diagnostic/diagnostic.go
```diff
@@ -0,0 +1,54 @@
...
+
+type diagnosticImpl struct {
+	alert alerter.Alert
+}
+
+func (d *diagnosticImpl) Code() string {
+	if _, ok := d.alert.(*alerts.RemoteAccessDeniedAlert); ok {
+		return "ACCESS_DENIED"
+	}
+	return "NOT_IMPLEMENTED"
```
craigfurman, line 24: Is this user-facing? If so, wouldn't something like UNKNOWN_ERROR be more appropriate?

craigfurman (dismissed):
2 non-blocking notes, I'll approve in advance 👍 

Before we merge this, I/you/we could test this branch out in snyk-iac-test and see if we can spike out bubbling up these non-fatal errors from diagnostics. WDYT?

enumeration/alerter/alert.go
eliecharra: Yeah in some tests, it's a manual mock if you see what I mean.

eliecharra:
> Before we merge this, I/you/we could test this branch out in snyk-iac-test and see if we can spike out bubbling up these non-fatal errors from diagnostics. WDYT?

100% in sync with that, let's do a small pairing session on that, shoot me an invite 🙏🏻

enumeration/diagnostic/diagnostic.go
eliecharra: Yeah that's user facing here, I didn't want to return an empty value but `UNKNOWN_ERROR` you proposed is probably better 👍🏻 Will change that thanks

enumeration/alerter/alert.go
craigfurman: Shouldn't we move it into the test files then?

eliecharra: Yeah agreed, that's typically tech debt here.

craigfurman: you're right, it's out of scope for this PR! I forgot I was commenting on the new interface method, not a whole new struct!

craigfurman (dismissed):
Nice!

PR #1582, opened 2022-08-24
Add support for google-beta provider

mdietzer-fn:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | no
| 🚀 New feature?   | yes (sort of)
| ⚠ Deprecations?   | no
| ❌ BC Break       | no
| 🔗 Related issues | N/A
| ❓ Documentation  | yes (the new `gcp-beta+tf` remote)

## Description

This PR adds support for the `google-beta` provider via the `gcp-beta+tf` remote.

Reasoning behind supporting `google-beta` is that several features in GCP require use of the beta provider, and in order to validate IaC infrastructure using that provider, this code is necessary, or driftctl asserts 0% coverage.

As the two providers, beyond their name and version, are basically the same when it comes to driftctl's usage, this is a quite small change


mdietzer-fn:
Test failures seem (to me) unrelated to my change, are these expected/flaky?

wyardley:
This would be great. Just curious: how would this work if you're using both regular and beta provider in the same state?

mdietzer-fn:
> This would be great. Just curious: how would this work if you're using both regular and beta provider in the same state?

Likely the same way  driftctl works with two different providers entirely, like trying to use AWS and GCP providers at the same time (which is to say it does not support that at all)

wyardley:
> Likely the same way driftctl works with two different providers entirely, like trying to use AWS and GCP providers at the same time (which is to say it does not support that at all)

Got it.
In my experience, it's very common with GCP in terraform to commingle regular and beta provider within the same state, though I imagine using beta only would probably still support the superset of things that are in GA as well most of the time....

Issue #1590, opened 2022-10-13
Cannot run a scan when you dont have any s3_account_public_access_block


moadibfr:
**Description**
Running a scan without that resource produce:
```
NoSuchPublicAccessBlockConfiguration: The public access block configuration was not found
	status code: 404, request id: R30QBEAJTC707FWZ, host id: sqNzmnOrcseaVDkklIugYH9tbbwfflYZzSHz4C5ZsvxJ0c9Y7zzjlKb0kR8KITmpd3A9Wnpomlk=
```

PR #1591, opened 2022-10-13
 fix issue when scanning without any s3_account_public_access_block

moadibfr:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | yes
| 🚀 New feature?   | yes
| ⚠ Deprecations?   | yes/no
| ❌ BC Break       | yes/no
| 🔗 Related issues | #1590 1590
| ❓ Documentation  | no 

## Description

Describing s3_account_public_access_block when you never had this resource will result in a 404 that was not handled and thus failed the scan instead of just returning no resources


wbeuil (approved):
LGTM

PR #1602, opened 2022-11-29
Update to Go 1.18

sundowndev:
| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | no
| 🚀 New feature?   | no
| ⚠ Deprecations?   | no
| ❌ BC Break       | no
| 🔗 Related issues | #...
| ❓ Documentation  | no 

## Description

Editing go.mod was missing in https://github.com/snyk/driftctl/pull/1543. Was it intentional?

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/snyk/driftctl
 
-go 1.17
+go 1.18
 
 require (
 	cloud.google.com/go/asset v0.1.0


eliecharra:
Hey @sundowndev 👋🏻 😃

Yes it was intentional, there is no direct coupling between go version and the mod version. We can use go1.18 without needing to use features from go mod 1.18. Which is the case here so yes it was intentional AFAIK.

sundowndev:
@eliecharra Thanks for the details, I'm closing this.

Issue #1344, opened 2022-02-07
ThrottlingException: Rate exceeded


Arisfx:
Hi team, do you know how we can avoid the rate exceeded error? 

Scanned states (7)      
ThrottlingException: Rate exceeded
        status code: 400, request id: 0474b16c-faee-402a-bf01-1e2a7c005714

sundowndev:
Hi @Arisfx, the rate limit error can occur when your cloud have a huge amount of resources, even when not managed by Terraform. This is part of known limitations of driftctl. Are you running in deep mode ? If yes, can you consider running in non-deep mode instead ? Note driftctl will no longer be able to show drifted in attributes. If you've identified which resource(s) is causing this, you can try to ignore this particular resource type using the [driftignore](https://docs.driftctl.com/0.19.0/usage/filtering/driftignore) or [the filter flag](https://docs.driftctl.com/0.19.0/usage/filtering/rules). If none of those solutions fit your needs, could you give further details on your use case ?

Thanks 🙏🏻

Arisfx:
Hi @sundowndev thank you for your response, seems if i run it several times, eventually it does the job, may i ask you if its possible to combine several filters? for example how can i filter for `aws_route53_zone` **and** `aws_route53_record` at the same time? Unfortunately, more than one filter commands can not be combined.
Thanks!

eliecharra:
> thank you for your response, perhaps i didnt make my self clear, i meant that we need to scan for only these 2 resources, not to exclude them :)

You can do this with the second snippet of code that @sundowndev has posted above

```ignore
# Ignore all drifts except for aws_route53_zone and aws_route53_record
*
!aws_route53_zone
!aws_route53_record
```

The first wildcard makes sure that we switch to an ignore everything mode but what is prefixed with `!`

This will save you a lot of API call and can definitively help with rate limit issues

brunzefb:
@sundowndev Just ran into the throttling issue as well with driftctl. Created a support call with AWS to increase API allowed rate.  They told me it would be too many rates to increase, and ask the authors to implement an exponential backoff when making AWS calls that hit the Throttling exception.  While this may kill the performance of the tool, maybe that does not matter so much -- especially if you are running it as a cron job once a day.

aroes:
This is an issue for me as well, I agree that a retry and backoff strategy should be implemented as @brunzefb 's AWS support suggests. Getting a full overview of a large account is almost impossible since the tool exits as soon as it runs into the error.

[...]

PR #1645, opened 2023-03-18
Added exponential backoff of calls to api gateway api

drem-darios:
Starts at 2 seconds and retries 5 times while backing off exponentially.

| Q                 | A
| ----------------- | ---
| 🐛 Bug fix?       | yes
| 🚀 New feature?   | no
| ⚠ Deprecations?   | no
| ❌ BC Break       | no
| 🔗 Related issues | #1344
| ❓ Documentation  | no

## Description
Added exponential backoff of calls to api gateway api to avoid TooManyCallsException. Calls that fail will sleep for 2 seconds before retrying. Every failure up to the max retries (default 5) takes exponentially more time than the previous sleep time.

You can see here: https://docs.aws.amazon.com/apigateway/latest/developerguide/limits.html that API Gateway allows 5 requests every 2 seconds per account for GetResources and I was hitting that limit pretty frequently. There is also a 10 request per second limit across all API Gateway management operations. To work around those limits, I added some code to the api_gateway_repository.go file that would exponentially backoff the requests in the case that we received a "TooManyRequestsException" error. I set the bar at 2 seconds since that was the limit we were hitting. Also, I had to add this logic to every function making a request to API Gateway since any of them could trigger the total operation limitation. (e.g. GetRestApisPage reaches the limit, then a call to say GetAccount will trigger the throttle).

This logic should probably be implemented in other places/repositories too...


mfractal:
when is this planned to get merged ? been hitting 'ThrottlingException: Rate exceeded' all the time recently running driftctl

markus-geiger:
I merged here https://github.com/markus-geiger/driftctl but still do get:

```
Throttling: Rate exceeded
	status code: 400, request id: a01c4539-04c7-47ec-ab6f-aa52b31f0768
```

PR #1657, opened 2023-05-08
chore: bump go-getter to resolve vuln and go mod tidy

karniwl:


test/google/asset.go
```diff
@@ -51,7 +52,7 @@ func newAssetClient(fakeServer *FakeAssetServer) (*asset.Client, error) {
 	client, err := asset.NewClient(ctx,
 		option.WithEndpoint(fakeServerAddr),
 		option.WithoutAuthentication(),
-		option.WithGRPCDialOption(grpc.WithInsecure()),
+		option.WithGRPCDialOption(grpc.WithTransportCredentials(insecure.NewCredentials())),
```
craigfurman, line 55: 😱 

But this diff doesn't make it worse, so this is non-blocking

craigfurman (approved):
🚀 

Glad we fixed those acceptance tests in order to feel better about the aws-sdk bump!

PR #1659, opened 2023-05-31
tidy go mod

chdorner-snyk:
as instructed by `go mod tidy`, ran `go mod tidy -go=1.16 && go mod tidy -go=1.17`

did a quick sanity check on those two upgrades and fairly sure they are fine


agatakrajewska (approved):

craigfurman (approved):
Thanks! Approving since this is indeed a net improvement, but ideally let's also check for no diff after go-mod-tidy in CI

PR #1664, opened 2023-06-01
[CTX-601] chore: always retry terraform-destroy

craigfurman:
Many acceptance test failures are caused by stray resources left behind from previous runs. We think failed terraform-destroy cleanup might be to blame. This commit adds unconditional retry with a long exponential backoff to try to ensure that cleanup can work within cloud provider rate limits, reducing the maintenance burden.

Remove RetryConfig, which a small number of tests were using to optionally add retry behaviour.


chdorner-snyk (approved):
🍘

PR #1666, opened 2023-06-02
chore: disable GitHub acceptance tests

craigfurman:
FYI @karniwl @agatakrajewska @chdorner-snyk 

Agata and I paired on recovering the acc test org and user, and failed. The token in circle is not an admin of that org. We have no choice but to abandon these. If we wanted, we could create new ones, but it doesn't seem worth the effort atm.


chdorner-snyk (approved):
🍘

agatakrajewska (approved):

PR #1748, opened 2026-05-24
fix: add timeout to gRPC provider Cleanup to prevent scan hang

### LINKED ISSUE #1653: driftctl hangs.
Hi!

When i run a scan againt s3, driftctl hangs. And i do not get any info from debug:

`Scan duration: 52s
Provider version used to scan: . Use --tf-provider-version to use another version.`

But when i run a scan locally it doesn't hang, and the debug shows me the following:

`can duration: 11s
Provider version used to scan: . Use --tf-provider-version to use another version.
DEBU[0013] Closing providers
DEBU[0013] Closing provider key=aws
DEBU[0013] Closing gRPC client alias=eu-north-1`

MaheshYadlapati:
Fixes #1653

## Problem
`driftctl scan` hangs after completing all resource enumeration in certain
environments (containerised, slow networks, CI). The scan completes
successfully but the process never exits.

## Root Cause
`Cleanup()` in `enumeration/remote/terraform/provider.go` calls
`client.Close()` as a bare blocking call with no timeout:

```go
func (p *TerraformProvider) Cleanup() {
    for alias, client := range p.grpcProviders {
        logrus.WithFields(...).Debug("Closing gRPC client")
        client.Close()  // blocks indefinitely
    }
}
```

The debug output from affected users confirms the scan completes and
reaches "Closing gRPC client alias=eu-north-1" — the hang is on that
call in environments where gRPC shutdown does not return cleanly.

## Fix
Wrap `client.Close()` in a goroutine with a 5-second timeout. If the
close does not complete within the timeout, a warning is logged and
execution continues rather than blocking forever.

No new imports required — `time` and `plugin` are already imported.

## Testing
Existing test suite unchanged. The fix is isolated to the shutdown path
and does not affect any resource enumeration or drift detection logic.

### DIFF
--- a/enumeration/remote/terraform/provider.go
+++ b/enumeration/remote/terraform/provider.go
@@ -219,6 +219,20 @@ func (p *TerraformProvider) Cleanup() {
 		logrus.WithFields(logrus.Fields{
 			"alias": alias,
 		}).Debug("Closing gRPC client")
-		client.Close()
+		done := make(chan struct{})
+		go func(c *plugin.GRPCProvider) {
+			c.Close()
+			close(done)
+		}(client)
+		select {
+		case <-done:
+			logrus.WithFields(logrus.Fields{
+				"alias": alias,
+			}).Debug("gRPC client closed")
+		case <-time.After(5 * time.Second):
+			logrus.WithFields(logrus.Fields{
+				"alias": alias,
+			}).Warn("gRPC client close timed out, forcing exit")
+		}
 	}
 }


MaheshYadlapati:
Hi team, raised this PR to address #1653. @MaraDascalu01, would really appreciate a review and approval when you get a chance.

MaheshYadlapati:
@MaraDascalu01 , Please review and approve the PR

